PR #14, opened 2023-08-29
update subnet-evm v0.5.4

cam-schultz:
## Why this should be merged
Pins subnet-evm to v0.5.4 and bumps avalanchego to v1.10.9

## How this was tested
CI


michaelkaplan13 (approved):

go.mod
```diff
@@ -54,6 +53,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.11.0 // indirect
 	go.opentelemetry.io/proto/otlp v0.19.0 // indirect
 	go.uber.org/atomic v1.10.0 // indirect
+	go.uber.org/mock v0.2.0 // indirect
```
minghinmatthewlam, line 56: The `uber/mock` pkg was updated because no one maintains `golang/mock` anymore right? For relayer tests I decided to leave as `golang/mock` since in `avalanchego` there are some uses of the mock pkg to create mock loggers, etc.

cam-schultz: Yes, `golang/mock` is deprecated. Note that in https://github.com/ava-labs/awm-relayer/pull/17 I updated some of our tests to use `uber/mock`.

minghinmatthewlam: looks like avalanchego also updated to use uber's gomock so lgtm

minghinmatthewlam (approved):

PR #16, opened 2023-08-29
Update SPDX License ID in Solidity files

michaelkaplan13:
## Why this should be merged

Aligns the SDPX License ID to the LICENSE file in the repo. Note that "Ecosystem" is not an official SPDX License ID on this list [here](https://spdx.org/licenses/), but according to the [documentation here](https://docs.soliditylang.org/en/v0.8.21/layout-of-source-files.html#spdx-license-identifier), "The compiler does not validate that the license is part of the list allowed by SPDX, but it does include the supplied string in the bytecode metadata.".

Thank you to @minghinmatthewlam for calling this out.

## How this was tested

CI/CD

## How is this documented

N/A


minghinmatthewlam (approved):
LGTM

PR #18, opened 2023-08-30
Fixing solhint lint workflow

### LINKED ISSUE #11: Solhint workflow not catching lint errors
**Describe the bug**
The github workflow that runs `solhint` is not catching lint errors.

**To Reproduce**
On the main branch the `Linter` check passes, even though there are meant to be lint errors. Checking raw logs `solhint` never outputs anything after its version.

**Expected behavior**
Github workflow linter check outputs `solhint` lint errors and fails the check.

**Screenshots**
If ran locally:
image

**Operating System**
Mac M1 Ventura 13.5.1

minghinmatthewlam:
## Why this should be merged
Fixes #11 , where workflows should now fail on solhint errors
Adds a solhint disable for a false positive solhint error on `abi.encode`

## How this works
`solhint **/*.sol` is apparently different from `solhint '**/*'.sol` where the latter will sol files in nested directories under `contracts/src`, whereas the former only looks at sol files in `contracts/src` directory and one directory down.

## How this was tested
ci

## How is this documented
added comment in worfklow

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -17,7 +17,7 @@ jobs:
 
             - name: Install solhint
               run: |
-                  npm install solhint@3.4.1 -g
+                  npm install solhint -g
                   solhint --version
 
             - name: Run Lint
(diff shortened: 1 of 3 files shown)


cam-schultz (approved):
LGTM, thanks for fixing this. Verified that the solhint command works on my setup.

gwen917 (approved):

PR #21, opened 2023-09-06
Fix forge create argument order

martineckardt:
## Why this should be merged

[Forge create](https://book.getfoundry.sh/reference/forge/forge-create) uses the env variable `$user_private_key`. Before this change the order was:

``` bash
forge create --private-key $user_private_key src/Mocks/ExampleERC20.sol:ExampleERC20 --rpc-url $subnet_a_url 
```

If the env variable `$user_private_key` is not set, foundry thinks `src/Mocks/ExampleERC20.sol:ExampleERC20` is the argument for the `--private-key` flag and the error message was stating that no contract was supplied. This is misleading since the real problem is, that no private key is supplied.

## How this works

Change the order of the argument

``` bash
forge create --private-key $user_private_key --rpc-url $subnet_a_url src/Mocks/ExampleERC20.sol:ExampleERC20 
```

This way of no private key is supplied, the error message states that.

## How this was tested

Ran the scripts

## How is this documented

No documentation necessary

### DIFF
--- a/scripts/local/integration-tests/basic_send_receive.sh
+++ b/scripts/local/integration-tests/basic_send_receive.sh
@@ -30,10 +30,10 @@ set -e # Stop on first error
 
 # Deploy a test ERC20 to be used in the E2E test.
 cd contracts
-erc20_deploy_result=$(forge create --private-key $user_private_key src/Mocks/ExampleERC20.sol:ExampleERC20 --rpc-url $subnet_a_url)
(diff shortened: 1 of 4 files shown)


minghinmatthewlam:
yup the foundry error messages for missing values can definitely be misleading, would the same issue happen if `$subnet_a_url` was not set? I'm wondering if changing the order so that the contract is passed first would fix that `forge create src/Mocks/ExampleERC20.sol:ExampleERC20 --private-key $user_private_key --rpc-url $subnet_a_url`

martineckardt:
Very good point. We will look into this and test this out. Based on that we will propose new changes. I marked this PR as do not merge for now.

PR #28, opened 2023-09-19
Ginkgo basic test

cam-schultz:
## Why this should be merged
Adds Ginkgo testing framework and implements a basic test that sends a Teleporter message one way from one subnet to another.

## How this works
Adds Ginkgo test suite that initializes a local network, deploys Teleporter to two subnets, and sends a Teleporter message between them. Rather than adding a full dependency on `awm-relayer`, the test fetches the aggregate signature from the Warp backend and sends the transaction to the destination itself.

Future work will consolidate these E2E tests with the `awm-relayer` E2E tests, with the aim to build a set of composable test functions for setting up a network, sending messages, and relaying them under a variety of test scenarios.

## How this was tested
Existing CI and new E2E job.

## How is this documented
Added E2E steps to README


go.mod
```diff
@@ -3,46 +3,113 @@ module github.com/ava-labs/teleporter
...
 
 require (
-	github.com/ava-labs/avalanchego v1.10.9 // indirect
+	github.com/ava-labs/avalanchego v1.10.9
 	github.com/supranational/blst v0.3.11 // indirect
 )
 
 require (
+	github.com/ava-labs/avalanche-network-runner v1.7.2-0.20230825150237-723bc7b31724
+	github.com/ava-labs/awm-relayer v0.2.1-0.20230920145222-b8571020bdaa
```
cam-schultz, line 12: We can remove this dependency on `awm-relayer` in a future PR by migrating the used parts of `package teleporter` in `awm-relayer` to `teleporter`. We use the ABI packing/unpacking methods defined there, so it would make sense to have those live in the same repo as the `TeleporterMessenger` contract from which they are derived.

contract-deployment/contractDeploymentTools.go
```diff
@@ -153,7 +25,10 @@ func main() {
 		if len(os.Args) != 3 {
 			log.Panic("Invalid argument count. Must provide JSON file containing contract bytecode.")
 		}
-		constructKeylessTransaction(os.Args[2])
+		_, _, _, err := tests.ConstructKeylessTransaction(os.Args[2], true)
```
michaelkaplan13, line 28: It feels a little odd to be using a utility from the `tests` package in the `main` package of another application. Could we move it to a `delpoyment-utils` package under `contract-deployment` still maybe?

cam-schultz: Moved to `package deployment_utils` under `contract-deployment`

contract-deployment/contractDeploymentTools.go
```diff
@@ -4,143 +4,15 @@
...
-	"encoding/json"
 	"fmt"
-	"io/fs"
 	"log"
-	"math/big"
 	"os"
 	"strconv"
 
-	"github.com/ava-labs/subnet-evm/core/types"
+	deployment_utils "github.com/ava-labs/teleporter/contract-deployment/deployment-utils"
```
minghinmatthewlam, line 12: I think we should just name the pkg `utils`, and places that import can rename if needed.

contract-deployment/contractDeploymentTools.go
```diff
@@ -166,7 +41,7 @@ func main() {
 			log.Panic("Failed to parse nonce as uint", err)
 		}
 
-		resultAddress := deriveEVMContractAddress(deployerAddress, nonce)
+		resultAddress := deployment_utils.DeriveEVMContractAddress(deployerAddress, nonce)
```
minghinmatthewlam, line 44: would prefer the utils function to return the error, then we handle the err here with panic

scripts/local/versions.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+# Set up the versions to be used
+awm_relayer_version=${AWM_RELAYER_VERSION:-'v0.2.1'}
```
minghinmatthewlam, line 6: can't we uppercase these similar to `GINKGO_VERSION`? We'd still be able to override when set.

tests/e2e_test.go
```diff
@@ -0,0 +1,565 @@
...
+				SubnetSpec: &rpcpb.SubnetSpec{
+					SubnetConfig: "",
+					Participants: subnetBNodeNames,
+				},
+			},
+		},
+	)
+	Expect(err).Should(BeNil())
+
+	// Issue transactions to activate the proposerVM fork on the receiving chain
```
minghinmatthewlam, line 142: for the comment in this case not just receiving chain

tests/e2e_test.go
```diff
@@ -0,0 +1,565 @@
...
+	log.Info("Running ginkgo after suite")
+	Expect(manager).ShouldNot(BeNil())
+	Expect(manager.TeardownNetwork()).Should(BeNil())
+	Expect(os.Remove(warpChainConfigPath)).Should(BeNil())
+})
+
+// Ginkgo describe node that acts as a container for the teleporter e2e tests. This test suite
+// will run through the following steps in order:
+// 1. Send a transaction to the Teleporter contract on Subnet A
+// 2. Aggregate signatures and send the Warp message to Subnet B
```
minghinmatthewlam, line 311: can you add that it verifies the received warp message values

tests/utils.go
```diff
@@ -0,0 +1,106 @@
...
+	// Parse the port
+	port, err := strconv.ParseUint(hostAndPort[1], 10, 32)
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to parse port: %w", err)
+	}
+
+	return hostAndPort[0], uint32(port), nil
+}
+
+func newTestTeleporterMessage(chainIDInt *big.Int, teleporterAddress common.Address, nonce uint64, data []byte) *types.Transaction {
```
minghinmatthewlam, line 78: nit: can we rename to `newTestTeleporterTransaction` and replace Message with Transaction in the default values

tests/e2e_test.go
cam-schultz: I actually think we should remove Warp message verification, since the interface we're testing end-to-end is Teleporter, not Warp. Explicit Warp message verification is best left up to subnet-evm. I've removed this now.

tests/e2e_test.go
```diff
@@ -0,0 +1,497 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/ava-labs/avalanche-network-runner/rpcpb"
+	"github.com/ava-labs/avalanchego/ids"
+	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
+	"github.com/ava-labs/awm-relayer/messages/teleporter"
+	relayerEvm "github.com/ava-labs/awm-relayer/vms/evm"
+	"github.com/ava-labs/coreth/rpc"
```
minghinmatthewlam, line 20: we should use `subnet-evm/rpc` here and remove coreth dependency

cam-schultz: Good call, replaced.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,55 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - "*"
+
+env:
+  GO_VERSION: "1.20.7"
```
michaelkaplan13, line 15: We should be able to use 1.20.8 now.

scripts/local/constants.sh
```diff
@@ -0,0 +1,19 @@
...
+TELEPORTER_PATH=$(
+    cd "$(dirname "${BASH_SOURCE[0]}")"
+    cd ../.. && pwd
+)
+
+# Set the PATHS
+GOPATH="$(go env GOPATH)"
+
+# Current branch
+current_branch=$(git symbolic-ref -q --short HEAD || git describe --tags --exact-match || true)
```
minghinmatthewlam, line 17: what is this used for? Does it need to be capitalized for use elsewhere

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Build Subnet-EVM Plugin Binary
+        run: |
+          cd subnet-evm
+          ./scripts/build.sh /tmp/e2e-test/avalanchego/plugins/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy
+
+      - name: Checkout teleporter repository
+        uses: actions/checkout@v4
+
+      - name: Install Forge and Run E2E Tests
```
michaelkaplan13, line 46: It'd probably be slightly cleaner to split this into two steps, one for installing foundry, the other to run the test script.

minghinmatthewlam (dismissed):

contract-deployment/utils/deployment_utils.go
```diff
@@ -0,0 +1,146 @@
...
+		return common.Address{}, errors.Wrap(err, "Failed to RLP encode address and nonce value.")
+	}
+	hash := crypto.Keccak256Hash(rlpEncoded)
+	return common.HexToAddress(fmt.Sprintf("0x%x", hash.Bytes()[12:])), nil
+}
+
+// Constructs a keyless transaction using Nick's method
+// Optionally writes the transaction, deployer address, and contract address to file
+// Returns the transaction bytes, deployer address, and contract address
+func ConstructKeylessTransaction(byteCodeFileName string, writeFile bool) ([]byte, common.Address, common.Address, error) {
```
michaelkaplan13, line 63: It looks like `writeFile` is not properly used/checked in the implementation.

scripts/local/versions.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+# Set up the versions to be used
+AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v0.2.1'}
+SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.5.4'}
```
michaelkaplan13, line 7: You're probably already on top of this, but we'll need to update this to `v0.5.6` with the latest commits from `main`.

tests/utils.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type SendCrossChainMessageEvent struct {
+	DestinationChainID ids.ID
+	MessageID          *big.Int
+	Message            teleporter.TeleporterMessage
+}
+
+// Teleporter contract sendCrossChainMessage input type
+type TeleporterMessageInput struct {
```
michaelkaplan13, line 39: Any reason to not name this `SendCrossChainMessageInput`?

tests/utils.go
```diff
@@ -0,0 +1,113 @@
...
+	// Parse the port
+	port, err := strconv.ParseUint(hostAndPort[1], 10, 32)
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to parse port: %w", err)
+	}
+
+	return hostAndPort[0], uint32(port), nil
+}
+
+func newTestTeleporterTransaction(chainIDInt *big.Int, teleporterAddress common.Address, nonce uint64, data []byte) *types.Transaction {
```
michaelkaplan13, line 85: Is there anything that makes this function specific to Teleporter necessarily?

tests/utils.go
```diff
@@ -0,0 +1,113 @@
...
+		To:        &teleporterAddress,
+		Gas:       defaultTeleporterTransactionGas,
+		GasFeeCap: defaultTeleporterTransactionGasFeeCap,
+		GasTipCap: defaultTeleporterTransactionGasTipCap,
+		Value:     defaultTeleporterTransactionValue,
+		Data:      data,
+	})
+}
+
+func setUpProposerVm(ctx context.Context, fundedKey *ecdsa.PrivateKey, manager *runner.NetworkManager, index int) {
```
michaelkaplan13, line 98: Nit: `setUpProposerVM`

Could you also include a comment what this is used for?

tests/utils.go
michaelkaplan13, line 85: Nit: Could we have the first parameter name just be `chainID`?

tests/e2e_test.go
```diff
@@ -0,0 +1,497 @@
...
+			// Loop until each node has advanced to >= the height of the block that emitted the warp log
+			for {
+				block, err := client.BlockByNumber(ctx, nil)
+				Expect(err).Should(BeNil())
+				if block.NumberU64() >= newHeadA.Number.Uint64() {
+					log.Info("client accepted the block containing SendWarpMessage", "client", i, "height", block.NumberU64())
+					break
+				}
+			}
+		}
```
michaelkaplan13, line 407: Would it be possible to somehow move this to a `waitForAllValidatorsToAcceptBlock` helper? I imagine it could be useful for other tests in the future.

tests/e2e_test.go
```diff
@@ -0,0 +1,497 @@
...
+		)
+
+		// Sign and send the transaction on the destination chain
+		signer := types.LatestSignerForChainID(chainBIDInt)
+		signedTxB, err := types.SignTx(destinationTx, signer, fundedKey)
+		Expect(err).Should(BeNil())
+
+		log.Info("Sending transaction to destination chain")
+		err = chainBRPCClient.SendTransaction(context.Background(), signedTxB)
+		Expect(err).Should(BeNil())
```
michaelkaplan13, line 462: Similarly, could this be moved to a `broadcastSignedMessage` helper?

tests/utils.go
cam-schultz: Just the context of being in a Teleporter test utility file, and the fact that the default values used to populate the tx fields are (loosely) tied to Teleporter.

cam-schultz: If we do so, we'll have to convert from `ids.ID` to `*big.Int` anyway. The blockchainID is returned by the ethclient as `*big.Int`, so I think we should leave it as is.

contract-deployment/utils/deployment_utils.go
cam-schultz: Whoops, fixed now.

tests/utils.go
cam-schultz: Renamed.

scripts/local/versions.sh
cam-schultz: Yup, just waiting on https://github.com/ava-labs/teleporter/pull/29

tests/e2e_test.go
cam-schultz: Added a helper

tests/e2e_test.go
cam-schultz: Added a helper

scripts/local/constants.sh
cam-schultz: Not used currently. This was taken from `constants.sh` in `awm-relayer`, which took it from `avalanchego`

.github/workflows/e2e.yml
cam-schultz: I tried that, but was running into issues persisting the `PATH` to include `foundryup` and `ginkgo`.

tests/e2e_test.go
```diff
@@ -0,0 +1,444 @@
...
+		// Loop over each client on chain A to ensure they all have time to accept the block.
+		// Note: if we did not confirm this here, the next stage could be racy since it assumes every node
+		// has accepted the block.
+		waitForAllValidatorsToAcceptBlock(ctx, chainANodeURIs, blockchainIDA, newHeadA.Number.Uint64())
+
+		// Get the aggregate signature for the Warp message
+		log.Info("Fetching aggregate signature from the source chain validators")
+		warpClient, err := warpBackend.NewWarpClient(chainANodeURIs[0], blockchainIDA.String())
+		Expect(err).Should(BeNil())
+		signedWarpMessageBytes, err := warpClient.GetAggregateSignature(ctx, unsignedWarpMessageID, params.WarpQuorumDenominator)
```
minghinmatthewlam, line 397: do you know the pros/cons of using `getAggregateSignature` vs how we get signatures and aggregate ourselves in the relayer?

cam-schultz: Here, it's the simplest way to get the aggregate signature in the end to end test.

This is outside the scope of this PR, but an off chain relayer aggregating signatures may wish to do so by requesting individual signatures from validators directly because:
- Signature aggregation wouldn't depend on a single node's availability
- At least one fewer network round trips since when compared to requesting an external node collect and aggregate signatures

In practice, the first point is probably the more relevant of the two.

tests/utils.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// Split the uri into host and port
+	hostAndPort := strings.Split(uri, ":")
+	Expect(len(hostAndPort)).Should(Equal(2))
+
+	// Parse the port
+	port, err := strconv.ParseUint(hostAndPort[1], 10, 32)
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to parse port: %w", err)
```
bernard-avalabs, line 84: Wouldn't this fail if the port is not specified? Or are we always using a non-default port in this case?

bernard-avalabs (commented):
Generally makes sense to me.

tests/utils.go
cam-schultz: This function requires the URI to include the port. If it does not, then the line `Expect(len(hostAndPort)).Should(Equal(2))` would fail.

tests/e2e_test.go
michaelkaplan13: Yep, agree here. In practice, the single node may already have a complete aggregate signature that it can return directly rather than needing to query its peers (for instance, if it caches the results), so that could cut down on the total amount of network requests.

Outside the scope either way here though 👍

.github/workflows/e2e.yml
michaelkaplan13: I see this also consistent with our current workflow for running unit tests with `forge`, so it makes sense to keep it for now.

michaelkaplan13:
Generally LGTM! Curious if you think it would be worth while to include a script to run these e2e locally. It could check that `subnet-evm` is properly checked out side-by-side, and run the necessary install/build/run commands.

michaelkaplan13:
One other small suggestion (not sure how easy it is), would be to have the local nodes started by Ginkgo tests to have a log level of INFO rather than DEBUG. Looking to see where this could be set now.

cam-schultz:
> Generally LGTM! Curious if you think it would be worth while to include a script to run these e2e locally. It could check that `subnet-evm` is properly checked out side-by-side, and run the necessary install/build/run commands.

We could add a script that takes as arguments the `subnet-evm` path and the data directory. I don't think we should impose that `subnet-evm` is in any particular place relative to `teleporter`. I'll put this together.

> One other small suggestion (not sure how easy it is), would be to have the local nodes started by Ginkgo tests to have a log level of INFO rather than DEBUG. Looking to see where this could be set now.

I tried adjusting the log level in the `ANRConfig` object, but it doesn't look like that propagates down to the Avalanchego nodes. Luckily, once the network is up and running, there's no further logs from the nodes to interfere with the actual test logic.

scripts/local/e2e_test_manual.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
minghinmatthewlam, line 2: thoughts on making this a command option in the original `e2e_test.sh` like `./scripts/local/e2e_test --local args...`?

minghinmatthewlam (commented):
LGTM, just a comment about replacing `e2e_test_manual` with a command option in `e2e_test.sh`

minghinmatthewlam (approved):
LGTM!

gwen917 (approved):
LGTM

PR #32, opened 2023-09-21
Define ReceiptQueue as library

gwen917:
## Why this should be merged
- Define ReceiptQueue as library instead of contract

## How this works
By running the gas report against entire unit tests, we find that the gas cost of `TeleporterMessenger` deployment reduces from `2393532` to `2117016`. The average gas cost of `receiveCrossChainMessage` reduces from `257916` to `89808`, and the average gas cost of `sendCrossChainMessage` increases from `100103` to `101576`. The `receiveCrossChainMessage` add receipts into outstanding receipt queue and `sendCrossChainMessage` removes receipts from the queue.
- The gas report for the ReceiptQueue contract
Screenshot 2023-09-25 at 3 54 11 PM

- The gas report for the ReceiptQueue library
Screenshot 2023-09-25 at 3 53 35 PM

## How this was tested
- Unit tests


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -771,10 +771,10 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     ) private returns (TeleporterMessageReceipt[] memory result) {
         // Get the current outstanding receipts for the given chain ID.
         // If the queue contract doesn't exist, there are not outstanding receipts to send.
-        ReceiptQueue outstandingReceiptForDestination = outstandingReceipts[
+        ReceiptQueue.TeleporterMessageReceiptQueue storage outstandingReceiptForDestination = outstandingReceipts[
```
cam-schultz, line 774: Can we specify `outstandingReceiptForDestination` as `memory`? `storage` doesn't seem appropriate since this is a stack variable that doesn't need to be preserved between function calls.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -66,32 +69,31 @@ contract ReceiptQueue {
...
-        external
+    function dequeue(
+        TeleporterMessageReceiptQueue storage queue
+    )
+        internal
         returns (TeleporterMessageReceipt memory result)
     {
-        if (msg.sender != owner) {
-            revert Unauthorized();
+        unchecked {
```
cam-schultz, line 78: What are the gas savings of using `unchecked` here vs the default checked arithmetic? Have you tried using `SafeMath`? Overflow is still theoretically possible here, even though it's practically impossible.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -37,25 +39,26 @@ contract ReceiptQueue {
...
      *
      * Requirements:
      *
      * - `msg.sender` must be the owner.
      */
-    function enqueue(TeleporterMessageReceipt calldata receipt) external {
-        if (msg.sender != owner) {
-            revert Unauthorized();
+    function enqueue(
+        TeleporterMessageReceiptQueue storage queue,
```
cam-schultz, line 50: Is `storage` specified here in order to mutate the queue passed by the caller?

cam-schultz (commented):
What are the gas savings on the first teleporter message delivery when using a library vs the existing approach of deploying a `ReceiptQueue` contract? What is the gas analysis for subsequent message delivery?

I also have a handful of other questions that I left as comments.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -11,11 +11,13 @@ import "./ITeleporterMessenger.sol";
...
  * TeleporterMessageReceipt structs. It provides FIFO properties.
  */
-contract ReceiptQueue {
-    address public immutable owner;
-    mapping(uint256 => TeleporterMessageReceipt) public queue;
-    uint256 public first = 0;
-    uint256 public last = 0;
+library ReceiptQueue {
+    struct TeleporterMessageReceiptQueue {
+        address owner;
```
michaelkaplan13, line 16: We shouldn't need to have an "owner" if using this as a library because it is not deployed to it's address with its own storage.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: It unfortunately needs to be declared as `storage` since the struct contains a mapping, which can only live in storage of a contract.

That being said, could we move `_getOutstandingReceiptsToSend` to be a function in the `ReceiptQueue` library? I think that would likely be cleaner and safe on gas.

contracts/src/Teleporter/ReceiptQueue.sol
minghinmatthewlam: is `owner` just used right now for checking whether the receipt queue has been used for a chain yet? Similar to previous `address(receiptQueue) == address(0)`?

minghinmatthewlam: Can we achieve the same by checking whether size of receipt queue is 0?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -40,7 +41,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     // Tracks the outstanding receipts to send back to a given subnet in subsequent messages sent to it.
     // Key is the subnet ID of the other subnet, and the value is a queue of pending receipts for messages
     // we have received from that subnet.
-    mapping(bytes32 => ReceiptQueue) public outstandingReceipts;
+    mapping(bytes32 => ReceiptQueue.TeleporterMessageReceiptQueue) public outstandingReceipts;
```
minghinmatthewlam, line 44: when setting this state variable, and the fields within set to their default value? How about for the struct `TeleporterMessageReceipt` inside the receipt queue's data?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -350,12 +351,11 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         // If the receipts queue contract for this chain doesn't exist yet, create it now.
-        ReceiptQueue receiptsQueue = outstandingReceipts[
+        ReceiptQueue.TeleporterMessageReceiptQueue storage receiptsQueue = outstandingReceipts[
             warpMessage.originChainID
         ];
-        if (address(receiptsQueue) == address(0)) {
-            receiptsQueue = new ReceiptQueue();
-            outstandingReceipts[warpMessage.originChainID] = receiptsQueue;
+        if (receiptsQueue.owner == address(0)) {
+            receiptsQueue.owner = msg.sender;
```
minghinmatthewlam, line 358: in this case is the first relayer that delivers a Teleporter message to this chain set as the "owner" of the receipt queue?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -771,24 +769,20 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     ) private returns (TeleporterMessageReceipt[] memory result) {
         // Get the current outstanding receipts for the given chain ID.
         // If the queue contract doesn't exist, there are not outstanding receipts to send.
-        ReceiptQueue outstandingReceiptForDestination = outstandingReceipts[
-            chainID
-        ];
-        if (address(outstandingReceiptForDestination) == address(0)) {
+        uint256 outstandingReceiptsSize = outstandingReceipts[chainID].size();
```
michaelkaplan13, line 772: I think it will be cleanest to move `_getOutstandingReceiptsToSend` to now be a function in the `ReceiptQueue` library.

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: removed the owner

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: move getOutstandingReceiptsToSend into the library.

contracts/src/Teleporter/ReceiptQueue.sol
gwen917: it has to be storage because `TeleporterMessageReceiptQueue` contains a mapping

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: move `getOutstandingReceiptsToSend` into the library.

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: yes, it will be the default value if it does not exist in the map. The fields in the struct are initialized as default value.
```
struct TeleporterMessageReceiptQueue {
        uint256 first;
        uint256 last;
        mapping(uint256 index => TeleporterMessageReceipt) data;
}
```
`first`, `last`, `data` are initialized as the default value. When we access `data[idx]`, if it does not exist, it should be default value of struct `TeleporterMessageReceipt`.

contracts/src/Teleporter/ReceiptQueue.sol
michaelkaplan13: As of Solidity 0.8.0, the compiler itself takes care of overflow and underflow checks, so `SafeMath` shouldn't be needed here.

It's still worth thinking through if we should allow it or not though (even though it would only occur once more than 2^256 messages have been received by a subnet 😅). If we don't allow overflow, I believe there is a hard limit of 2^256 messages that could ever be received by a single `TeleporterMessenger` instance, because all subsequent messages would revert here when they are tried to be delivered. If we do allow overflow, then after the 2^256th message is received, the next message will loop around and be stored at slot 0. This could be okay, if by that time slot 0 was freed up by the receipt being removed from the queue. If it had yet to be removed, we would overwrite it, and the `first` and `last` values would become equal, effectively making the queue empty. 

While it's unfeasible for that to occur, there are already other places in the contract where we have a theoretical limit of 2^256 messages that could ever be sent between two subnets (because message IDs are tracked as `uint256`), so I don't think it's worth using `unchecked` here. If something overflows, lets just revert rather than possibly having it silently continue on.

(2^256 is so large that this whole thought experiment is a bit ridiculous either way. 😅)

bernard-avalabs: Agree that we should revert on overflow. If we send 1000 messages per second, it would take: 3.6712774e+64 centuries to overflow 2^256.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -66,32 +65,53 @@ contract ReceiptQueue {
      * - `msg.sender` must be the owner.
```
bernard-avalabs, line 65: Same with this comment.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -66,32 +65,53 @@ contract ReceiptQueue {
...
         returns (TeleporterMessageReceipt memory result)
     {
-        if (msg.sender != owner) {
-            revert Unauthorized();
-        }
+        uint256 first_ = queue.first;
+        if (queue.last == first_) revert EmptyQueue();
+        result = queue.data[first_];
+        delete queue.data[first_];
+        queue.first = first_ + 1;
```
bernard-avalabs, line 78: Is it worthwhile to reset first and last to their initial value (0) if they are ever equal after a dequeue?

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -66,32 +65,53 @@ contract ReceiptQueue {
...
+     * @dev Returns the outstanding receipts for the given chain ID that should be included in the next message sent.
+     */
+    function getOutstandingReceiptsToSend(
+        TeleporterMessageReceiptQueue storage queue
+    )
+        internal
+        returns (TeleporterMessageReceipt[] memory result)
+    {
+        // Get the current outstanding receipts for the given chain ID.
+        // If the queue contract doesn't exist, there are not outstanding receipts to send.
```
bernard-avalabs, line 93: typo: there are "no" outstanding

bernard-avalabs (commented):
Generally looks good. I've left a few minor comments.

contracts/src/Teleporter/ReceiptQueue.sol
gwen917: fixed

contracts/src/Teleporter/ReceiptQueue.sol
gwen917: fixed

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -7,15 +7,19 @@ pragma solidity 0.8.18;
 
 import "./ITeleporterMessenger.sol";
 
+// The maximum number of receipts to include in a single message.
+uint256 constant MAXIMUM_RECEIPT_COUNT = 5;
+
 /**
  * @dev ReceiptQueue is a convenience contract that creates a queue-like interface of
```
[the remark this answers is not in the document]
gwen917: fixed

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -37,23 +41,18 @@ contract ReceiptQueue {
...
-    constructor() {
-        owner = msg.sender;
-    }
-
     /**
      * @dev Adds a receipt to the queue.
      *
      * Requirements:
      *
      * - `msg.sender` must be the owner.
```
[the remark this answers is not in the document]
gwen917: fixed

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -350,13 +346,10 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         // Store the receipt of this message delivery. When a subsquent message is sent back
         // to the origin of this message, we will clean up the receipt state.
         // If the receipts queue contract for this chain doesn't exist yet, create it now.
-        ReceiptQueue receiptsQueue = outstandingReceipts[
+        ReceiptQueue.TeleporterMessageReceiptQueue storage receiptsQueue = outstandingReceipts[
```
cam-schultz, line 349: What happens in this case if the mapping does not exist?

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -62,36 +56,55 @@ contract ReceiptQueue {
...
+     */
+    function getOutstandingReceiptsToSend(
+        TeleporterMessageReceiptQueue storage queue
+    )
+        internal
+        returns (TeleporterMessageReceipt[] memory result)
+    {
+        // Get the current outstanding receipts for the given chain ID.
+        // If the queue contract doesn't exist, there are no outstanding receipts to send.
+        uint256 outstandingReceiptsSize = size(queue);
```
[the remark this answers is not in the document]
michaelkaplan13: Nevermind, `queue.size()` is not able to used within the library itself.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -7,15 +7,19 @@ pragma solidity 0.8.18;
 
 import "./ITeleporterMessenger.sol";
 
+// The maximum number of receipts to include in a single message.
+uint256 constant MAXIMUM_RECEIPT_COUNT = 5;
```
minghinmatthewlam, line 11: any reason we define the constant outside library?

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: IIUC since the `outstandingReceipts` is a mapping to structs instead of `ReceiptQueue` contract instances now, all chain ids will have the default mapping exist. Will want to make sure to have this tested.

minghinmatthewlam: can we add a unit test covering this case, and making sure it defaults to all the zero values for chains that we have not received from.

contracts/src/Teleporter/ReceiptQueue.sol
gwen917: moved into library.

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: I think [this test](https://github.com/ava-labs/teleporter/blob/main/contracts/src/Teleporter/tests/ReceiveCrossChainMessageTests.t.sol#L24) covered this case.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -576,6 +569,28 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return _checkIsAllowedRelayer(delivererAddress, allowedRelayers);
     }
 
+    /**
+     * @dev Returns the number of outstanding receipts that have been received from the given chain ID.
+     */
+    function getOutstandingReceiptQueueSize(bytes32 chainID)
```
michaelkaplan13, line 575: I think we should add these to the `ITeleporterMessenger` interface as well so that other contracts that want to use them only need to import the interface and not the entire implementation.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -8,90 +8,94 @@ pragma solidity 0.8.18;
...
-    function size() external view returns (uint256) {
-        return last - first;
+    function size(TeleporterMessageReceiptQueue storage queue) internal view returns (uint256) {
+        return queue.last - queue.first;
+    }
+
+    /**
+     * @dev Returns the receipt at the given index in the queue.
+     */
+    function getReceiptAtIndex(TeleporterMessageReceiptQueue storage queue, uint256 index)
```
bernard-avalabs, line 94: Since we never return the index to a receipt, do we need a get a receipt at a specific index? Or is this mainly used for testing?

bernard-avalabs (approved):
LGTM. Just one question about the need for getReceiptAtIndex.

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -204,4 +204,17 @@ interface ITeleporterMessenger {
         bytes32 destinationChainID,
         uint256 messageID
     ) external view returns (address feeAsset, uint256 feeAmount);
+
+    /**
+     * @dev Gets the number of outstanding receipts that have been sent to the given destination chain ID.
+     */
+    function getOutstandingReceiptQueueSize(bytes32 chainID) external view returns (uint256 size);
```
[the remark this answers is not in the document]
gwen917: fixed

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -8,90 +8,94 @@ pragma solidity 0.8.18;
...
+        internal
+        view
+        returns (TeleporterMessageReceipt memory)
+    {
+        return queue.data[index];
```
michaelkaplan13, line 99: In addition to changing this to treat the index as the index in the queue, could you add a unit test for it? 🙏

michaelkaplan13 (approved):
LGTM!

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -566,6 +559,28 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return _getNextMessageID(chainID);
     }
 
+    /**
+     * @dev Returns the number of receipts that have been received from the given chain ID.
```
minghinmatthewlam, line 563: For comment we can do something like `See {ITeleporterMessenger-getReceiptQueueSize`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -566,6 +559,28 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        view
+        returns (uint256)
+    {
+        return outstandingReceipts[chainID].size();
+    }
+
+    /**
+     * @dev Returns the receipt at the given index in the queue for the given chain ID.
+     */
+    function getReceiptAtIndex(bytes32 chainID, uint256 index)
```
minghinmatthewlam, line 576: Same as above, and just add requirement note for index

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -8,90 +8,96 @@ pragma solidity 0.8.18;
 import "./ITeleporterMessenger.sol";
 
 /**
- * @dev ReceiptQueue is a convenience contract that creates a queue-like interface of
+ * @dev ReceiptQueue is a convenience library that creates a queue-like interface of
  * TeleporterMessageReceipt structs. It provides FIFO properties.
```
minghinmatthewlam, line 12: can you add a note here about the need for functions to be `internal`

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -204,4 +204,17 @@ interface ITeleporterMessenger {
...
+
+    /**
+     * @dev Gets the number of receipts that have been sent to the given destination chain ID.
+     */
+    function getReceiptQueueSize(bytes32 chainID) external view returns (uint256 size);
+
+    /**
+     * @dev Gets the receipt at the given index in the queue for the given chain ID.
+     */
+    function getReceiptAtIndex(
```
minghinmatthewlam, line 216: can you add a `@param` comment about how `index` is compared to the front of the queue, and possibly example that `index=0` returns next receipts to be sent

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: can we add a specific test case for this in terms of:
- create a `mapping` similar to `outstandingReceipts`
- grab a receipt queue for a mock chain id
- go through every value in the queue struct and check defaults to 0

bernard-avalabs (dismissed):
LGTM

contracts/src/Teleporter/ITeleporterMessenger.sol
gwen917: added `@param`

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: fixed

contracts/src/Teleporter/ReceiptQueue.sol
gwen917: added comments

contracts/src/Teleporter/tests/ReceiptsQueueTests.t.sol
```diff
@@ -9,87 +9,99 @@ import "forge-std/Test.sol";
...
+    TeleporterMessageReceipt private _receipt2 = TeleporterMessageReceipt({
+        receivedMessageID: 684384,
+        relayerRewardAddress: 0x10eB43ef5982628728E3E4bb9F78834f67Fbb40b
+    });
+    TeleporterMessageReceipt private _receipt3 = TeleporterMessageReceipt({
+        receivedMessageID: 654351,
+        relayerRewardAddress: 0xcC8E718045817AebA89592C72Ae1C9917f5D0894
+    });
 
     function testReceiptQueue() public {
```
michaelkaplan13, line 32: Could we renaming this test case `testEnqueueDequeueSuccess` to differentiate it now that there are other test cases for the `ReceiptQueue`?

michaelkaplan13 (approved):
LGTM! 

I re-ran the gas reports also and it looks like there is a significant decrease in the gas cost of `receiveCrossChainMessage` (likely due to no longer needing to deploy the queue contract the first time), and a marginal decrease in the gas cost of `sendCrossChainMessage` (~1,000 gas less for the median). Overtime, both of those could be significant.

minghinmatthewlam (commented):
LGTM, just one comment for if we can add a unit test for https://github.com/ava-labs/teleporter/pull/32#discussion_r1341820155

contracts/src/Teleporter/tests/ReceiveCrossChainMessageTests.t.sol
```diff
@@ -238,4 +240,13 @@ contract ReceiveCrossChainMessagedTest is TeleporterMessengerTest {
...
     }
+
+    function testEmptyReceiptQueue() public {
+        ReceiptQueue.TeleporterMessageReceiptQueue storage receiptQueue = outstandingReceipts[
+            DEFAULT_ORIGIN_CHAIN_ID
+        ];
+
+        assertEq(receiptQueue.first, 0);
+        assertEq(receiptQueue.last, 0);
+    }
```
gwen917, line 251: Is this the unit test you mentioned? @minghinmatthewlam

minghinmatthewlam: can we move this to `ReceiptQueueTests` and add a check that `size()` returns 0, and that `getReceiptAtIndex(0)` reverts for an empty queue.

Can we also add separate unit tests for the new external functions in `ITeleporterMessenger.getReceiptQueueSize()` and `ITeleporterMessenger.getReceiptatIndex()`? These just cover cases like:
- If the chain never received a message yet
- Chain receives a receipt, queue size should increase and the get receipt at index works
- When a message gets sent back to origin chain, queue size decreases and the same index can't be queried for `getReceiptAtIndex` since index represents position relative to`start`

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -8,90 +8,97 @@ pragma solidity 0.8.18;
...
 
-        result = queue[first_];
-
-        delete queue[first_];
-        first = first_ + 1;
+        // Calculate the result size as the minimum of the number of receipts and maximum batch size.
+        uint256 resultSize = _MAXIMUM_RECEIPT_COUNT;
+        if (outstandingReceiptsSize < _MAXIMUM_RECEIPT_COUNT) {
+            resultSize = outstandingReceiptsSize;
+        }
```
cam-schultz, line 78: I think this would save a bit of gas by not declaring an extra variable. We'd probably want to rename `outstandingReceiptsSize` to `resultSize` though.
```suggestion
        if (outstandingReceiptsSize > _MAXIMUM_RECEIPT_COUNT) {
            outstandingReceiptsSize = _MAXIMUM_RECEIPT_COUNT;
        }
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -126,9 +124,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
                 requiredGasLimit: messageInput.requiredGasLimit,
                 allowedRelayerAddresses: messageInput.allowedRelayerAddresses,
                 message: messageInput.message,
-                receipts: _getOutstandingReceiptsToSend(
-                    messageInput.destinationChainID
-                )
+                receipts: outstandingReceipts[messageInput.destinationChainID].getOutstandingReceiptsToSend()
```
cam-schultz, line 127: Is this Solidity syntactic sugar for `getOutstandingReceiptsToSend(outstandingReceipts[messageInput.destinationChainID])`? The signature of this function is
```
function getOutstandingReceiptsToSend(
        TeleporterMessageReceiptQueue storage queue
)
```
so I'm trying to understand how dot-notation is used here.

cam-schultz (commented):
One small suggestion, and a question about Solidity syntax.

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: This is because we have `using ReceiptQueue for ReceiptQueue.TeleporterMessageReceiptQueue;` at beginning of the TeleporterMessenger contract.

cam-schultz: Gotcha. From the documentation, the way it works here is that we're taking the functions defined in `ReceiptQueue.sol` and adding them as operators for the type `ReceiptQueue.TeleporterMessageReceiptQueue`. So when we call using dot-notation here, this implicitly inserts the object as the first argument to the function call in `ReceiptQueue.sol`. 

I'm generally not a fan of this kind of aliasing/implicit function call conversion/operator overloading. However, it seems to be standard Solidity practice so I'm probably just being a curmudgeon :)

contracts/src/Teleporter/ReceiptQueue.sol
gwen917: fixed

minghinmatthewlam (approved):
LGTM

PR #34, opened 2023-09-21
Ginkgo parallel

cam-schultz:
Experimental change demonstrating parallel test suite runs with Ginkgo. The advantage is quicker test runs on a common network when compared to running a large number of tests in serial. The disadvantage is the significant complexity involved in setting up the initial state in each of the parallel runners such that the runners all have access to the necessary global state, and that they don't interfere with each other. That complexity can however likely be significantly reduced, but we should assess the performance benefits before doing so.


scripts/local/e2e_test.sh
```diff
@@ -33,9 +33,7 @@ ginkgo build ./tests/
 
 # Run the tests
 echo "Running e2e tests $RUN_E2E"
-RUN_E2E=true ./tests/tests.test \
-  --ginkgo.vv \
-  --ginkgo.label-filter=${GINKGO_LABEL_FILTER:-""}
+RUN_E2E=true REPO_ROOT=$TELEPORTER_PATH ginkgo -p tests/tests.test -vv
```
cam-schultz, line 36: Parallel runs require the ginkgo cli. This feature is not supported from the binary on its own.

tests/e2e_test.go
```diff
@@ -84,7 +105,8 @@ func TestE2E(t *testing.T) {
 // registered on the P-Chain.
 // Adds two disjoint sets of 5 of the new validator nodes to validate two new subnets with a
 // a single Subnet-EVM blockchain.
-var _ = ginkgo.BeforeSuite(func() {
+var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
```
cam-schultz, line 108: `SynchronizedBeforeSuite` takes two functions as parameters. The first function is run on a single core, while all other cores wait. The first function returns `[]byte` which are passed to the second function. The second function runs in parallel on all cores. Individual suites (ginkgo `Describe` blocks) are distributed among the parallel runners.

To accomodate this framework we have to:
- Define global variables at file scope. Each parallel runner _captures its own copy_ of these global variables, so we have to set them in the second function of `SynchronizedBeforeSuite`. If they are used in the first function, they need to be set there are well.
- Give each parallel runner a unique EVM address, otherwise we'll get nonce conflicts. Tests don't necessarily need to have unique addresses, just as long as the addresses are not used simultaneously. I chose to do this by modifying the genesis file.

geoff-vball: I think instead of modifying the genesis file, we should keep one funded address/key pair funded in the genesis file, and use a global helper function `FundAddress()`, that takes in an address, and amount, and a slice of WSClients as parameters. It then uses the global funded address to transfer funds to whatever wallet the test needs. This way we can dynamically allocate the funded address for most tests, and can also statically allocate one if we want. You'd have to put a mutex on `FundAddress()`, but I don't think it would add much overhead, and we gain scaleability.

tests/e2e_test.go
```diff
@@ -294,15 +315,205 @@ var _ = ginkgo.BeforeSuite(func() {
...
+			unsignedWarpMsg := unsignedMsg
+			log.Info("Parsed unsignedWarpMsg", "unsignedWarpMessageID", unsignedWarpMessageID, "unsignedWarpMessage", unsignedWarpMsg)
+
+			// Loop over each client on chain A to ensure they all have time to accept the block.
+			// Note: if we did not confirm this here, the next stage could be racy since it assumes every node
+			// has accepted the block.
+			for i, uri := range chainANodeURIs {
+				chainAWSURI := httpToWebsocketURI(uri, blockchainIDA.String())
+				log.Info("Creating ethclient for blockchainA", "wsURI", chainAWSURI)
+				client, err := ethclient.Dial(chainAWSURI)
```
geoff-vball, line 356: We're dialing the client an arbitrary number of times here. We should either dial once and save the client, or we should close the connection after every loop.

tests/e2e_test.go
```diff
@@ -294,15 +315,205 @@ var _ = ginkgo.BeforeSuite(func() {
...
+		},
+	}
+	setupBytes, err := json.Marshal(setupStruct)
+	Expect(err).Should(BeNil())
+
+	return setupBytes
+},
+	// Re-set all global variables that are referenced in the tests and set in the above function
+	func(setupBytes []byte) {
+		var setupStruct SetupStruct
```
geoff-vball, line 457: I think with the way you have the consolidate PR set up now, we can just do the necessary duplication in the getter function. We should just have to redial the client (or not, leave it up to the test to manage it's connections) and copy the *big.Int for the chainID

minghinmatthewlam:
@cam-schultz should we close this PR, and reopen when needed? We can keep the branch open

cam-schultz:
> @cam-schultz should we close this PR, and reopen when needed? We can keep the branch open

Yeah let's go ahead and close this.

PR #38, opened 2023-09-25
Consolidate e2e

cam-schultz:
## Why this should be merged
Exports utilities for setting up a warp-enabled local network and deploying Teleporter to subnets for use with Ginkgo tests downstream of `teleporter`. For example, `awm-relayer` sets up the network in the exact same way, so exporting these functions will help reduce code duplication.

Also reorganizes the `tests/` directory. It now has the structure:
```
tests
 |-- utils/          ->   utility functions, test files
 |-- e2e-test.go     ->    e2e test entrypoint
 |-- test_*          ->    integration test case
```

## How this was tested
Existing CI

## How is this documented
Existing documentation


tests/e2e_test.go
```diff
@@ -4,71 +4,18 @@
 package tests
```
minghinmatthewlam, line 4: i like how simple this overall `e2e_test` file is, to make sure I understand correctly, is the idea for other tests to share the same network? 

How does it work between multiple files, each having their own `Describe` containers

cam-schultz: Yup exactly. `Describe` blocks will run in serial, one after the other, and use the same environment set up in `BeforeSuite`

tests/test_basic_one_way_send.go
```diff
@@ -0,0 +1,228 @@
+package tests
```
minghinmatthewlam, line 1: do we need to name the file specifically as `test_*.go` isntead of `*_test.go`?

cam-schultz: Yes. If it was `*_test.go`, then they would be included in `go test`. The `test_` prefix is not necessary, and might be redundant since they're already in package `tests`

tests/e2e_test.go
minghinmatthewlam: If you have multiple files, each with their own describe container, how does ginkgo know to run in order/ which order to run in? Or is it run randomly ordered but in serial?

tests/test_basic_one_way_send.go
minghinmatthewlam: ye can we remove the `test` prefix

tests/utils/network_setup.go
```diff
@@ -0,0 +1,271 @@
...
+		time.Sleep(5 * time.Second)
+		teleporterCode, err := ChainBRPCClient.CodeAt(ctx, TeleporterContractAddress, nil)
+		Expect(err).Should(BeNil())
+		Expect(len(teleporterCode)).Should(BeNumerically(">", 2)) // 0x is an EOA, contract returns the bytecode
+	}
+	log.Info("Finished deploying Teleporter contracts")
+}
+
+func TearDownNetwork() {
+	log.Info("Running ginkgo after suite")
```
minghinmatthewlam, line 267: nit: can we update this log to just be about tearing down network

tests/utils/network_setup.go
```diff
@@ -0,0 +1,271 @@
...
+		triggerTxB, err := types.SignTx(txB, txSignerB, FundedKey)
+		Expect(err).Should(BeNil())
+		err = ChainBRPCClient.SendTransaction(ctx, triggerTxB)
+		Expect(err).Should(BeNil())
+		time.Sleep(5 * time.Second)
+		receipt, err := ChainBRPCClient.TransactionReceipt(ctx, triggerTxB.Hash())
+		Expect(err).Should(BeNil())
+		Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+	}
+	// Deploy Teleporter on the two subnets
```
minghinmatthewlam, line 242: Can you add a log about successfully funding deployer address

tests/utils/network_setup.go
```diff
@@ -0,0 +1,271 @@
...
+	Expect(err).Should(BeNil())
+
+	ChainBIDInt, err = ChainBRPCClient.ChainID(context.Background())
+	Expect(err).Should(BeNil())
+
+	NewHeadsA = make(chan *types.Header, 10)
+
+	log.Info("Finished setting up e2e test subnet variables")
+
+	log.Info("Deploying Teleporter contract to subnets")
```
minghinmatthewlam, line 166: can you move this log down to function `DeployTeleporterContract`

minghinmatthewlam: nit: think `"Set up ginkgo before suite"` log can be removed from setup network

tests/e2e_test.go
cam-schultz: I'm not sure how ordering is determined in this case, but Ginkgo suites and specs should generally be independent of one another, so order shouldn't matter. It would be great if there was a way to shuffle `Describe` blocks, but it unfortunately seems like that shuffling is only done on `It` specs (as long as the suite isn't ordered).

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,175 @@
...
+		Message:                 []byte{1, 2, 3, 4},
+	}
+)
+
+// Ginkgo describe node that acts as a container for the teleporter e2e tests. This test suite
+// will run through the following steps in order:
+// 1. Send a transaction to the Teleporter contract on Subnet A
+// 2. Aggregate signatures and send the Warp message to Subnet B
+// 3. Verify receipt of the message on Subnet B
+var _ = ginkgo.Describe("[Teleporter one way send]", ginkgo.Ordered, func() {
```
michaelkaplan13, line 40: High level question:

Is the idea that each of the current "integration test" scripts will be translated into its own `ginkgo.Describe` container? It seems like, alternatively, each integration test script could be translated into its own `It` within a single `ginkgo.Describe` container, such that the `It`s should then be able to run in parallel.

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,175 @@
...
+
+		sendCrossChainMessageLog := receipt.Logs[0]
+		var event SendCrossChainMessageEvent
+		err = teleporter.EVMTeleporterContractABI.UnpackIntoInterface(&event, "SendCrossChainMessage", sendCrossChainMessageLog.Data)
+		Expect(err).Should(BeNil())
+		teleporterMessageID = event.Message.MessageID
+	})
+
+	ginkgo.It("Check Teleporter Message Received", ginkgo.Label("Teleporter", "TeleporterMessageReceived"), func() {
+		time.Sleep(5 * time.Second) // Give the relayer a chance to deliver the message to the destination chain
```
michaelkaplan13, line 156: I think this waiting is redundant given that we've already waited in the `It` above for a new block to be produced after sending the transaction on the destination, right?

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,175 @@
...
+var (
+	teleporterMessage = teleporter.TeleporterMessage{
+		MessageID:               big.NewInt(1),
+		SenderAddress:           FundedAddress,
+		DestinationAddress:      FundedAddress,
+		RequiredGasLimit:        big.NewInt(1),
+		AllowedRelayerAddresses: []common.Address{},
+		Receipts:                []teleporter.TeleporterMessageReceipt{},
+		Message:                 []byte{1, 2, 3, 4},
+	}
```
geoff-vball, line 32: This `teleporterMessage` is only used in one location, and only for the gas limit.

tests/utils/network_setup.go
```diff
@@ -0,0 +1,270 @@
...
+	Expect(err).Should(BeNil())
+
+	NewHeadsA = make(chan *types.Header, 10)
+
+	log.Info("Finished setting up e2e test subnet variables")
+}
+
+// DeployTeleporterContract deploys the Teleporter contract to the two subnets. The caller is responsible for generating the
+// deployment transaction information
+func DeployTeleporterContract(transactionBytes []byte, deployerAddress common.Address, contractAddress common.Address) {
```
geoff-vball, line 169: I think if we refactor this to take a slice of RPC clients as a param, and then loop over them to perform the deploy, we can simplify this function a lot. Also if we execute the fund/deploy on both chains before sleeping, we can speed up the tests a little bit by eliminating one of the sleeps.

tests/utils/network_setup.go
```diff
@@ -0,0 +1,270 @@
...
+	SubnetIDs                        []ids.ID
+	SubnetA, SubnetB                 ids.ID
+	BlockchainIDA, BlockchainIDB     ids.ID
+	ChainANodeURIs, ChainBNodeURIs   []string
+	FundedAddress                    = common.HexToAddress("0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC")
+	FundedKey                        *ecdsa.PrivateKey
+	ChainAWSClient, ChainBWSClient   ethclient.Client
+	ChainARPCClient, ChainBRPCClient ethclient.Client
+	ChainARPCURI, ChainBRPCURI       string
+	ChainAIDInt, ChainBIDInt         *big.Int
```
geoff-vball, line 41: We should be very careful with global exported values like this. We should at the very least leave a comment here detailing when they're supposed to be initialized, and that they should not be modified.

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,175 @@
...
+		subA, err := ChainAWSClient.SubscribeNewHead(ctx, NewHeadsA)
+		Expect(err).Should(BeNil())
+		defer subA.Unsubscribe()
+
+		log.Info("Sending Teleporter transaction on source chain", "destinationChainID", BlockchainIDB, "txHash", signedTx.Hash())
+		err = ChainARPCClient.SendTransaction(ctx, signedTx)
+		Expect(err).Should(BeNil())
+
+		// Sleep to ensure the new block is published to the subscriber
+		time.Sleep(5 * time.Second)
```
geoff-vball, line 84: I think we could speed the test suite up quite a bit if we created a helper function `WaitNewBlock` that takes an RPC client as a param, and calls `SubscribeNewHead` and waits for one new block to be produced. That way we can eliminate the `time.Sleep`s.

tests/basic_one_way_send.go
cam-schultz: Reading the Ginkgo docs again, the latter approach seems to be what we should aim for. For our use case, the `Describe` container should roughly translate to: "Describe the specs that should be met when messages are passed between subnets using Teleporter". Then, the `It` containers should roughly translate to each individual integration test. I agree we should refactor this code to reflect this.

tests/basic_one_way_send.go
cam-schultz: This sleep is to give the relayer a chance to construct the aggregate signatures and deliver to the destination. We could probably replace this wait with a `SubscriberNewHead` on the destination chain. Geoff made a similar suggestion, so I'll look into implementing this.

tests/utils/network_setup.go
cam-schultz: That's a good call. I think a `GetSubnetInfo` exported function would be more appropriate, that way these globals are not modifiable outside of `package utils`.

tests/basic_one_way_send.go
cam-schultz: I agree we should be waiting on new blocks rather than sleeps. However, a helper might be difficult to set up properly since we have to subscribe before sending a transaction in order to ensure that the new block will be picked up by the subscription.

tests/basic_one_way_send.go
cam-schultz: Removed.

tests/basic_one_way_send.go
cam-schultz: Removed.

tests/basic_one_way_send.go
cam-schultz: I refactored the tests to follow the pattern described above. It was relatively straightforward, and no major framework changes needed to be made.

tests/utils/network_setup.go
cam-schultz: Implemented this. Thanks for the suggestion.

tests/utils/network_setup.go
cam-schultz: Put all these vars behind exported getters. Each test case should extract these at the beginning of the test.

tests/basic_one_way_send.go
cam-schultz: `SendAndWaitForTransaction` is implemented as a wrapper around `ethclient.SendTransaction`. One case that didn't quite fit is the call to deploy the Teleporter contract raw tx bytes, but that's a one off case so I think it's fine that this helper doesn't apply.

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,156 @@
...
+	tx := NewTestTeleporterTransaction(subnetAInfo.ChainIDInt, teleporterContractAddress, nonceA, data)
+
+	txSigner := types.LatestSignerForChainID(subnetAInfo.ChainIDInt)
+	signedTx, err := types.SignTx(tx, txSigner, fundedKey)
+	Expect(err).Should(BeNil())
+
+	log.Info("Sending Teleporter transaction on source chain", "destinationChainID", subnetBInfo.BlockchainID, "txHash", signedTx.Hash())
+	newHeadA := SendAndWaitForTransaction(ctx, subnetAInfo.ChainWSClient, signedTx)
+
+	receipt, err := subnetAInfo.ChainWSClient.TransactionReceipt(ctx, signedTx.Hash())
```
minghinmatthewlam, line 71: can we move the receipt status checking into `SendAndWaitForTransaction`

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,156 @@
...
+// 1. Send a transaction to the Teleporter contract on Subnet A
+// 2. Aggregate signatures and send the Warp message to Subnet B
+// 3. Verify receipt of the message on Subnet B
+func BasicOneWaySend() {
+	var (
+		teleporterMessageID *big.Int
+	)
+
+	subnetAInfo := utils.GetSubnetATestInfo()
+	subnetBInfo := utils.GetSubnetATestInfo()
```
geoff-vball, line 33: Typo here, you're getting subnet A info for subnet B

tests/utils/utils.go
```diff
@@ -158,7 +186,7 @@ func constructAndSendTransaction(
 	numSigners, err := signedMessage.Signature.NumSigners()
 	Expect(err).Should(BeNil())
 
-	gasLimit, err := teleporter.CalculateReceiveMessageGasLimit(numSigners, teleporterMessage.RequiredGasLimit)
+	gasLimit, err := teleporter.CalculateReceiveMessageGasLimit(numSigners, requiredGasLimit)
```
minghinmatthewlam, line 189: do we plan on leavint the dependency on `awm-relayer` for calculating receiving gas limit?

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,156 @@
...
+	receipt, err := subnetAInfo.ChainWSClient.TransactionReceipt(ctx, signedTx.Hash())
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	//
+	// Relay the message to the destination
+	//
+
+	// Get the latest block from Subnet A, and retrieve the warp message from the logs
+	log.Info("Waiting for new block confirmation")
```
geoff-vball, line 80: I don't think this log applies anymore

tests/utils/utils.go
```diff
@@ -140,14 +168,14 @@ func waitForAllValidatorsToAcceptBlock(ctx context.Context, nodeURIs []string, b
 
 // Constructs and sends a transaction containing a warp message for the destination chain.
 // Returns the signed transaction.
-func constructAndSendTransaction(
+func ConstructAndSendTransaction(
```
minghinmatthewlam, line 171: can you add warp in name like `SendWarpTransaction`

tests/utils/utils.go
minghinmatthewlam: also maybe add the receipt checking in the utils function

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,156 @@
...
+		subnetBInfo.ChainIDInt,
+	)
+
+	receipt, err = subnetBInfo.ChainWSClient.TransactionReceipt(ctx, signedTxB.Hash())
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	sendCrossChainMessageLog := receipt.Logs[0]
+	var event SendCrossChainMessageEvent
+	err = teleporter.EVMTeleporterContractABI.UnpackIntoInterface(&event, "SendCrossChainMessage", sendCrossChainMessageLog.Data)
```
minghinmatthewlam, line 133: why does this unpack into sendCrossChainmessage, shouldn't it be `receiveCrossChainMessage`

cam-schultz: This is the `SendCrossChainMessage` event on the sending subnet. We use it to get the Teleporter message ID

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,156 @@
...
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	//
+	// Relay the message to the destination
+	//
+
+	// Get the latest block from Subnet A, and retrieve the warp message from the logs
+	log.Info("Waiting for new block confirmation")
+	blockHashA := newHeadA.Hash()
```
geoff-vball, line 81: It would be really helpful if we could extract the manual relaying of warp messages into a util function. I think it probably makes sense in two parts: getting the `signedWarpMessageBytes`, and then sending it on the destination chain.

tests/utils/utils.go
cam-schultz: We can consider moving this to `teleporter` as well.

tests/basic_one_way_send.go
cam-schultz: Added a helper 👍

tests/basic_one_way_send.go
cam-schultz: Whoops, fixed now. I'm surprised the test worked with this. I wouldn't expect it to work in the general case.

tests/utils/network_setup.go
```diff
@@ -0,0 +1,276 @@
...
+			SendAndWaitForTransaction(ctx, subnetInfo.ChainWSClient, triggerTx)
+			receipt, err := client.TransactionReceipt(ctx, triggerTx.Hash())
+			Expect(err).Should(BeNil())
+			Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+		}
+		log.Info("Finished funding Teleporter deployer")
+
+		// Deploy Teleporter
+		{
+			rpcClient, err := rpc.DialContext(ctx, HttpToRPCURI(subnetInfo.ChainNodeURIs[0], subnetInfo.BlockchainID.String()))
```
geoff-vball, line 251: I think we should include the RPC clients in `SubnetTestInfo`. There are a few useful rpc endpoints that we might want in tests.

cam-schultz: What are those use cases? For context, I removed the RPC clients from the tests altogether, since they are redundant with the websocket clients. Websocket clients are able to make RPC requests, but RPC clients aren't able to establish websocket connections. The `rpcClient` here is normally wrapped by `ethclient.Client`, but `ethclient.Client` doesn't have a way to send raw transaction bytes, so we have to do it this was here. Since for this test we're only doing that to issue the initial Teleporter deploy tx, there's no reason to expose the `rpcClient` to the tests themselves.

tests/basic_one_way_send.go
minghinmatthewlam: I think this is the receiving subnet and is the receipt from sending the `receiveCrossChainMessage` to destination chain.

geoff-vball: Yeah, this is the receiving subnet. `SendCrossChainMessage` and `ReceiveCrossChainMessage` happen to have the same fields, so the message unpacks without error.

tests/utils/network_setup.go
geoff-vball: `ethclient.Client` limits us to the enumerated function calls in that interface. Being able to call `rpcClient.CallContext` with `debug_traceTransaction` would have helped speed my debugging up quite a bit. I've been pausing the test and making those calls from the command line, but being able to do them programmatically would be helpful.

tests/basic_one_way_send.go
minghinmatthewlam: can we add a check to confirm with the current unpacking of `event.Destination.ChainID` to blockChainB? This should fail. 

Then we can update the unpacking to `receiveCrossCHainMessage` and check `originChainID`

cam-schultz: Yup, you're right. Fixed this now to retrieve the Teleporter message ID from the sending subnet logs as intended.

tests/utils/network_setup.go
cam-schultz: I would argue that development/debugging tools shouldn't factor in strongly to an interface. When debugging, you can always construct the `rpcClient` from the `subnetInfo`. I would assume that debugging code would be removed before merging anyway.

tests/basic_one_way_send.go
michaelkaplan13: 🙌

tests/basic_one_way_send.go
minghinmatthewlam: I liked the check for unpacking the destination chain tx. Can we unpack the destination tx receipt log against the `receiveCrossChainMessage` and check it's `OriginChainID` and `MessageID`?

tests/utils/network_setup.go
geoff-vball: Yeah that's fair. The debugging code will obviously be removed from my test, but I'd like to leave behind a utility that can be used to debug easily in the future. I can do that without the RPC client being exposed so this is good 👍

tests/utils/utils.go
```diff
@@ -55,15 +64,42 @@ type FeeInfo struct {
...
 }
 
-func httpToWebsocketURI(uri string, blockchainID string) string {
+//
+// Test utility functions
+//
+
+// Subscribes to new heads, sends a tx, and waits for the new head to appear
+// Returns the new head
+func SendAndWaitForTransaction(
```
michaelkaplan13, line 73: Nit: Could we renaming this something like `SendTransactionAndWaitForAcceptance`?

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,99 @@
...
+	Expect(err).Should(BeNil())
+
+	// Send a transaction to the Teleporter contract
+	tx := NewTestTeleporterTransaction(subnetAInfo.ChainIDInt, teleporterContractAddress, nonceA, data)
+
+	txSigner := types.LatestSignerForChainID(subnetAInfo.ChainIDInt)
+	signedTx, err := types.SignTx(tx, txSigner, fundedKey)
+	Expect(err).Should(BeNil())
+
+	log.Info("Sending Teleporter transaction on source chain", "destinationChainID", subnetBInfo.BlockchainID, "txHash", signedTx.Hash())
```
michaelkaplan13, line 64: Would it be worthwhile to move this portion to a "SendTeleporterMessageTransaction" helper that takes the `SendCrossChainMessageInput` as param?

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,99 @@
...
+	subnetBInfo := utils.GetSubnetBTestInfo()
+	teleporterContractAddress := utils.GetTeleporterContractAddress()
+	fundedAddress, fundedKey := utils.GetFundedAccountInfo()
+
+	//
+	// Send a transaction to Subnet A to issue a Warp Message from the Teleporter contract to Subnet B
+	//
+	ctx := context.Background()
+
+	nonceA, err := subnetAInfo.ChainWSClient.NonceAt(ctx, fundedAddress, nil)
```
michaelkaplan13, line 38: With multiple `It`s running tests in parallel, I think we may have a race condition around the nonce usage here. 

Similarly, we could have race conditions around waiting for a transaction to be included in a block, because there may be a new block produced after a subscription is opened but before the transaction is sent/accepted by the network, given that other goroutines are also sending other transactions at the same time.

tests/utils/utils.go
```diff
@@ -87,27 +123,26 @@ func getURIHostAndPort(uri string) (string, uint32, error) {
...
-		Value:     defaultTeleporterTransactionValue,
+		Gas:       DefaultTeleporterTransactionGas,
+		GasFeeCap: DefaultTeleporterTransactionGasFeeCap,
+		GasTipCap: DefaultTeleporterTransactionGasTipCap,
+		Value:     DefaultTeleporterTransactionValue,
 		Data:      data,
 	})
 }
 
-// Issues txs to activate the proposer VM fork on the specified subnet index in the manager
```
michaelkaplan13, line 103: Could we keep this comment (or some version of it)?

michaelkaplan13 (commented):
I think there may be a couple concurrency issues with the existing set up when we add additional test `It`s, but overall I think this set up makes a lot of sense 👍

tests/basic_one_way_send.go
cam-schultz: Oh for sure. To be clear, the transition from having tests in a `Describe` vs a `It` container buys us the following:
- More idiomatic usage of Ginkgo
- Unordered test runs (good for fuzzing)

It does NOT buy us the ability to run the tests in parallel on its own. As you mentioned, there are concurrency issues with nonces, block channels, etc. We have to be careful when setting up the global vs thread state to account for this. This [draft PR](https://github.com/ava-labs/teleporter/pull/34) goes into more detail on some of those points.

tests/basic_one_way_send.go
cam-schultz: I packaged this code into the helper `CreateTeleporterTransaction`, which returns the signed transaction. This is then passed to `SendTransactionAndWaitForAcceptance`.

tests/utils/network_setup.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func GetSubnetATestInfo() SubnetTestInfo {
+	return SubnetTestInfo{
+		SubnetID:      subnetA,
+		BlockchainID:  blockchainIDA,
+		ChainNodeURIs: chainANodeURIs,
+		ChainWSClient: chainAWSClient,
+		ChainWSURI:    chainAWSURI,
+		ChainIDInt:    chainAIDInt,
```
geoff-vball, line 73: We should copy the `*big.Int` here, otherwise everyone will have a pointer to the same object.

tests/utils/network_setup.go
```diff
@@ -0,0 +1,274 @@
...
+		ChainWSClient: chainBWSClient,
+		ChainWSURI:    chainBWSURI,
+		ChainIDInt:    chainBIDInt,
+	}
+}
+func GetTeleporterContractAddress() common.Address {
+	return teleporterContractAddress
+}
+func GetFundedAccountInfo() (common.Address, *ecdsa.PrivateKey) {
+	return fundedAddress, fundedKey
```
geoff-vball, line 90: Technically we should probably also copy the private key here. I don't think there are any methods on the private key that mutate the underlying state, but the state variable are exported so you could tamper with them. There's no easy way to do this, but we _could_ call `toECDSA(FromECDSA(fundedKey))`

tests/utils/network_setup.go
```diff
@@ -0,0 +1,274 @@
...
+	// Set the package level teleporterContractAddress
+	teleporterContractAddress = contractAddress
+
+	ctx := context.Background()
+
+	for _, subnetInfo := range subnetsInfo {
+		client := subnetInfo.ChainWSClient
+
+		chainID, err := client.ChainID(ctx)
+		Expect(err).Should(BeNil())
```
geoff-vball, line 218: We actually already have the chainID in the subnetInfo

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+	DestinationChainID ids.ID
+	MessageID          *big.Int
+	Message            teleporter.TeleporterMessage
+}
+
+type ReceiveCrossChainMessageEvent struct {
+	OriginChainID ids.ID
+	MessageID     *big.Int
+	Message       teleporter.TeleporterMessage
+}
```
geoff-vball, line 50: Note: Once we start using the generated abi, the `.go` file comes with structs for events, as well as methods `Parse<eventName>()`.

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+
+	err = wsClient.SendTransaction(ctx, tx)
+	Expect(err).Should(BeNil())
+	newHead := <-newHeads
+
+	receipt, err := wsClient.TransactionReceipt(ctx, tx.Hash())
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	return newHead, receipt
```
geoff-vball, line 91: We probably don't have to return `newHead` here, the receipt has the blockhash included.

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+
+	return hostAndPort[0], uint32(port), nil
+}
+
+func CreateTeleporterTransaction(
+	ctx context.Context,
+	source SubnetTestInfo,
+	destination SubnetTestInfo,
+	fundedAddress common.Address,
+	fundedKey *ecdsa.PrivateKey,
```
geoff-vball, line 131: nit: We should rename these to `senderAddress` and `senderKey`

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+
+	txSigner := types.LatestSignerForChainID(source.ChainIDInt)
+	signedTx, err := types.SignTx(tx, txSigner, fundedKey)
+	Expect(err).Should(BeNil())
+
+	return signedTx
+}
+
+// Issues txs to activate the proposer VM fork on the specified subnet index in the manager
+func SetUpProposerVM(ctx context.Context, fundedKey *ecdsa.PrivateKey, manager *runner.NetworkManager, index int) {
```
geoff-vball, line 173: `senderKey` here as well.

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+	err = utils.IssueTxsToActivateProposerVMFork(ctx, chainIDInt, fundedKey, client)
+	Expect(err).Should(BeNil())
+}
+
+// Blocks until all validators specified in nodeURIs have reached the specified block height
+func WaitForAllValidatorsToAcceptBlock(ctx context.Context, nodeURIs []string, blockchainID ids.ID, height uint64) {
+	for i, uri := range nodeURIs {
+		chainAWSURI := HttpToWebsocketURI(uri, blockchainID.String())
+		log.Info("Creating ethclient for blockchainA", "wsURI", chainAWSURI)
+		client, err := ethclient.Dial(chainAWSURI)
```
geoff-vball, line 195: nit: we should `defer client.Close()` after the error check here.

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+	log.Info("Sending transaction to destination chain")
+	_, receipt := SendTransactionAndWaitForAcceptance(ctx, wsClient, signedTx)
+
+	return signedTx, receipt
+}
+
+// Constructs the aggregate signature, packs the Teleporter message, and relays to the destination
+func RelayMessage(
+	ctx context.Context,
+	block *types.Header,
```
geoff-vball, line 278: Let's just take the blockhash as a parameter here, we don't need anything else from the header.

geoff-vball (commented):
Left several comments. Nothing major but all things that I think would help the test suite. Feel free to kick the can down the road for any of these. I'll rereview as soon as you get the changes in so we can get this merged asap.

tests/utils/utils.go
cam-schultz: I'm going to keep the names as is, since they reflect what the keys represent in the network setup phase. A test can always alias them for readability to reflect how they're used in the test.

tests/utils/network_setup.go
cam-schultz: Made this change, thanks for the suggestion. You're right, it's probably not a realistic concern, but no harm in writing this bit defensively.

tests/utils/network_setup.go
cam-schultz: Good call. Modified.

tests/utils/utils.go
cam-schultz: Nice, those will be great to have. It's a shame there aren't similar generated functions for smart contract method inputs and outputs.

geoff-vball (dismissed):
👍

tests/basic_one_way_send.go
michaelkaplan13: Gotcha, I forgot that the specs as is will still run serially, just not in a specific order.

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+		return "", 0, fmt.Errorf("failed to parse port: %w", err)
+	}
+
+	return hostAndPort[0], uint32(port), nil
+}
+
+func CreateTeleporterTransaction(
+	ctx context.Context,
+	source SubnetTestInfo,
+	destination SubnetTestInfo,
```
michaelkaplan13, line 130: Instead of taking the `destination SubnetTestInfo` as a parameter here, I think it would be more useful to have it be a `SendCrossChainMessageInput`. That way, this utility can be used generally to creating transactions that send arbitrary messages.

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+		log.Info("Creating ethclient for blockchainA", "wsURI", chainAWSURI)
+		client, err := ethclient.Dial(chainAWSURI)
+		Expect(err).Should(BeNil())
+		defer client.Close()
+
+		// Loop until each node has advanced to >= the height of the block that emitted the warp log
+		for {
+			block, err := client.BlockByNumber(ctx, nil)
+			Expect(err).Should(BeNil())
+			if block.NumberU64() >= height {
```
michaelkaplan13, line 204: Can we add a log for the else case here? If one of the nodes stalls for whatever reason, I think it'd be helpful to be able to debug the tests.

cam-schultz: Adding a log in this tight loop would flood the output during bootstrapping, even if all nodes are healthy. We could add a sleep in the loop to reduce that, but IMO relying on the `Expect(err).Should(BeNil())` on the repeated client calls would illuminate any issues with nodes better than a "Waiting for node..." log message.

tests/utils/utils.go
cam-schultz: Changed 👍

tests/utils/utils.go
```diff
@@ -0,0 +1,333 @@
...
+				log.Info("client accepted the block containing SendWarpMessage", "client", i, "height", block.NumberU64())
+				break
+			}
+		}
+	}
+}
+
+// Constructs and sends a transaction containing a warp message for the destination chain.
+// Returns the signed transaction.
+func ConstructAndSendWarpTransaction(
```
michaelkaplan13, line 204: I think we can likely make the functions in utils here a bit more modular and symmetric with each other.

Right now we have:
- `ConstructAndSendWarpTransaction` used for creating and sending a transaction that calls "receiveCrossChainMessage" of Teleporter. Returns the signed transaction and the tx receipt (though it looks like only the receipt is ever used currently)
- `CreateTeleporterTransaction` used for creating a transaction that calls "sendCrossChainMessage" of Teleporter.
- `SendTransactionAndWaitForAcceptance` used for sending any EVM transaction and waiting until it is included in a block.

There is duplicate logic for getting/setting the nonce to be used in transactions, and there is some gas price logic that differs between the utilities also.

I'd recommend breaking those up into:
- `CreateSendCrossChainMessageTransaction`
- `CreateReceiveCrossChainMessageTransaction`
- `SendTransactionAndWaitForAcceptance`

And having some un-exported helpers used by them:
- `setGasPriceAndNonce`
- `signTransaction`

Open to any other suggestions also; there is likely more improvement options available.

cam-schultz: Made these changes. These utilities have developed organically over the past few PRs, so appreciate you taking a step back and reevaluating their usefulness.

geoff-vball (dismissed):

tests/utils/utils.go
cam-schultz: We should still have better handling for the case in which a node is not indexing blocks correctly, but also not returning an error. I've added a context timeout to this function. When the timeout expires, the call to `client.BlockByNumber` should fail.

tests/utils/utils.go
```diff
@@ -0,0 +1,335 @@
...
+
+func signTransaction(tx *types.Transaction, key *ecdsa.PrivateKey, chainID *big.Int) *types.Transaction {
+	txSigner := types.LatestSignerForChainID(chainID)
+	signedTx, err := types.SignTx(tx, txSigner, key)
+	Expect(err).Should(BeNil())
+
+	return signedTx
+}
+
+func calculateTxParams(ctx context.Context, wsClient ethclient.Client, fundedAddress common.Address) (*big.Int, *big.Int, uint64) {
```
michaelkaplan13, line 321: Nit: Can you add a comment above this function documenting what the return values are.

tests/utils/utils.go
michaelkaplan13: Great call with the timeout 🙏

tests/utils/network_setup.go
```diff
@@ -0,0 +1,275 @@
...
+	key, err := crypto.ToECDSA(crypto.FromECDSA(fundedKey))
+	Expect(err).Should(BeNil())
+	return fundedAddress, key
+}
+
+// setupNetwork starts the default network and adds 10 new nodes as validators with BLS keys
+// registered on the P-Chain.
+// Adds two disjoint sets of 5 of the new validator nodes to validate two new subnets with a
+// a single Subnet-EVM blockchain.
+func SetupNetwork(warpGenesisFile string) {
```
michaelkaplan13, line 99: Super nit: I think this should be `SetUpNetwork`. That would also make it consistent with `SetUpProposerVM`

tests/utils/utils.go
cam-schultz: Added

tests/utils/network_setup.go
cam-schultz: If it's all the same to you, I'm going to instead change it to `SetupProposerVM`. That way we're consistent with the `subnet-evm` `SetupNetwork` function call made by this function.

geoff-vball (approved):

tests/utils/network_setup.go
michaelkaplan13: 😅 I didn't notice that `SetupNetwork` was a `subnet-evm` function. The grammar police in me wants to say that's not right in this case since we're using it as a verb (set up) and not a noun (setup), but lets keep it for the sake of consistency haha.

michaelkaplan13: Could you just capitalize "SetupNetwork" in the comment above the function now that it's exported?

michaelkaplan13 (dismissed):
🚀

tests/basic_one_way_send.go
```diff
@@ -0,0 +1,84 @@
...
+
+	log.Info("Sending Teleporter transaction on source chain", "destinationChainID", subnetBInfo.BlockchainID, "txHash", signedTx.Hash())
+	receipt := utils.SendTransactionAndWaitForAcceptance(ctx, subnetAInfo.ChainWSClient, signedTx)
+
+	sendCrossChainMessageLog := receipt.Logs[0]
+	var event utils.SendCrossChainMessageEvent
+	err := teleporter.EVMTeleporterContractABI.UnpackIntoInterface(&event, "SendCrossChainMessage", sendCrossChainMessageLog.Data)
+	Expect(err).Should(BeNil())
+
+	teleporterMessageID = event.Message.MessageID
```
minghinmatthewlam, line 57: can we add extra verification about other event fields like `destinationChainID`, and comparing the even'ts teleporter message fields compared to `sendCrossChainMessageInput`?

tests/utils/utils.go
```diff
@@ -0,0 +1,340 @@
...
+		destination.ChainIDInt,
+	)
+
+	log.Info("Sending transaction to destination chain")
+	receipt := SendTransactionAndWaitForAcceptance(ctx, destination.ChainWSClient, signedTx)
+
+	receiveCrossChainMessageLog := receipt.Logs[0]
+	var event ReceiveCrossChainMessageEvent
+	err = teleporter.EVMTeleporterContractABI.UnpackIntoInterface(&event, "ReceiveCrossChainMessage", receiveCrossChainMessageLog.Data)
+	Expect(err).Should(BeNil())
```
minghinmatthewlam, line 309: I think we should eitehr return the `receipt` or `event`, then in the `BasicOneWaySend` test make sure that the received message has Teleporter fields that matches the `teleporterMessage` that was sent

minghinmatthewlam (approved):
Looks great to me. Just added some comments about extra verification for the `SendCrossChainMessage` event fields compared to `sendCrossChainMessageInput`, and also compare `receiveCrossChainMessage` event fields to the `sendCrossChainMessage` event fields. (EDIT: dismissed review was meant to leave as a comment)

tests/utils/utils.go
cam-schultz: I modified `RelayMessage` to return the receipt, but I think for this test it's sufficient to just check against the `TeleporterMessenger` contract by calling `messageReceived`. Other tests may want to verify the fields directly though, so good call to have this function return the receipt.

geoff-vball (approved):

tests/basic_one_way_send.go
cam-schultz: Added these checks, along with utilizing the abigen Go bindings.

michaelkaplan13:
Should we wait to update & merge this branch until https://github.com/ava-labs/teleporter/pull/50/files is merged? Otherwise I think we'd need a follow PR for it to stay up-to-date.

tests/utils/utils.go
```diff
@@ -0,0 +1,351 @@
...
+		destination.ChainIDInt,
+	)
+
+	log.Info("Sending transaction to destination chain")
+	receipt := SendTransactionAndWaitForAcceptance(ctx, destination.ChainWSClient, signedTx)
+
+	bind, err := teleportermessenger.NewTeleportermessenger(teleporterContractAddress, source.ChainWSClient)
+	Expect(err).Should(BeNil())
+	// Check the transaction logs for the ReceiveCrossChainMessage event emitted by the Teleporter contract
+	for _, log := range receipt.Logs {
```
minghinmatthewlam, line 311: any specific reason we iterate through these logs and then add a `Expect(false).Should(BeTrue()`, instead of similar to when we initially send message on source chain and just check parsing `receipt.Logs[0]`?

tests/utils/utils.go
minghinmatthewlam: are you thinking a separate test case to verify we sent and received the correct teleporter message with correct fields? I was thinking this test case would include sending, check received, and check that we received correct message.

cam-schultz:
> Should we wait to update & merge this branch until https://github.com/ava-labs/teleporter/pull/50/files is merged? Otherwise I think we'd need a follow PR for it to stay up-to-date.

My vot is to merge anything that's ready and resolve merge conflicts in further PRs once they're ready. Our backlog of pending PRs is messy enough as is.

tests/utils/utils.go
cam-schultz: The `ReceiveCrossChainMessage` event isn't guaranteed to be in index 0, so we iterate over all of the logs and check that one of them is the expected event. If none match, then we assert.

tests/utils/utils.go
minghinmatthewlam: More specifically I think we can return the receipt and parse the receipt log for receive event in `BasicOneWaySend)`
```
	receipt = utils.RelayMessage(ctx, receipt.BlockHash, receipt.BlockNumber, subnetAInfo, subnetBInfo)
	destinationBind, err := teleportermessenger.NewTeleportermessenger(teleporterContractAddress, subnetBInfo.ChainWSClient)
	Expect(err).Should(BeNil())
	receiveEvent, err := destinationBind.ParseReceiveCrossChainMessage(*receipt.Logs[0])
	Expect(err).Should(BeNil())
	Expect(receiveEvent.Message.MessageID).Should(Equal(teleporterMessageID))
	Expect(receiveEvent.OriginChainID).Should(Equal(subnetAInfo.BlockchainID))
	Expect(receiveEvent.Message).Should(Equal(sendCrossChainMessageInput.Message))
```

cam-schultz: This PR is to set up the testing framework, not to implement robust test cases. The test case here is the minimum test case to demonstrate the framework works as intended for this use case. Let's plan to revisit `basic_one_way_send` (or remove it altogether) in a future PR.

geoff-vball (dismissed):

tests/utils/utils.go
cam-schultz: I added helpers to iterate over a slice of logs, and return the send/receive logs if present. Else, returns an error.

minghinmatthewlam: sorry one more question to clarify own understanding, why isn't the `ReceiveCrossChainMessage` event guaranteed to be at index 0? Shouldn't the transaction receipt from the `receiveCrossChainMessage` destination tx only have the `ReceiveCrossChainMessage` event?

cam-schultz: There are three logs in this case. One of them is the `ReceiveCrossChainMessage` event, another is the `FailedMessageExecution` event (we are sending a dummy payload). I'm not sure what the third is, probably something else related to failed message execution.

michaelkaplan13 (approved):

minghinmatthewlam (approved):

tests/utils/utils.go
minghinmatthewlam: ah ye you're right, thanks for explaining

Issue #7, opened 2023-08-28
Upgrade Strategy


minghinmatthewlam:
**Context and scope**
Currently Teleporter is a non upgradeable contract, but in this case any bugs are found, or new features would like to be added, we would need some form of upgrade strategy. 

**Discussion and alternatives**
1. Implementing a pass-through proxy that forwards Teleporter calls on to the correct implementation contract, where the individual implementations maintain their own state. - @minghinmatthewlam points out that this would make it difficult to handle the `msg.sender` value when receiving teleporter messages, since it would change with each new implementation.
2. Using the standard upgradable proxy smart contract pattern. - In this case, we would need to be careful about the contract state and variable ordering. This would mean that the proxy contract needs to be deployed to the same contract on each chain using Nick's method, and that it would be possible for different Teleporter versions to send messages to one another. This is a non-starter as Teleporter depends on knowing that the messages it receives are from the same contract deployed bytecode deployed on another chain.
3. Using a upgradable protocol registry. - Instead of using a proxy, applications will simply lookup the current address to be used from a registry. We will provide a small intuitive library that wraps the necessary calls with the proper look up.

minghinmatthewlam:
closed with https://github.com/ava-labs/teleporter/pull/43

PR #43, opened 2023-09-27
Teleporter registry upgrade

minghinmatthewlam:
## Why this should be merged
Provides upgradeability of Teleporter through a registry contract #7. Creates a warp protocol registry that can be inherited by warp protocols, and for Teleporter specifically has `TeleporterRegistry`.

## How this works
Adds a warp registry contract that gets inherited by `TeleporterRegistry`. The registry can add new protocol versions through a warp out of band message, then keeps track of the version and protocol address mappings. The constructor of the registry also allows for initialization, so that the registry initially doesn't need an additional warp out of band message if Teleporter is already deployed.

Cross chain apps will take in the registry instead of previous passing Teleporter in their constructor. The best practice for sending will be to ask the registry every time for sending with the latest version. For receiving, cross chain apps are recommended to have a `minTeleporterVersion`, and for their `ITeleporterReceiver.receiveTeleporterMessage` implementation to check that the `msg.sender` matches a `version > _minTeleporterVersion`. 
`TeleporterUpgradeable` provides these best practices as an abstract contract that cross chain apps built on Teleporter can inherit and use.

## How this was tested
Unit Tests, adding tests for Teleporter/Warp registry and `TeleporterUpgradeable`
E2E test update to pass in registry for cross chain apps while warp out of band messages are still being implemented.

## How is this documented
upgrade README


minghinmatthewlam:
marking as ready for review, will add a `README` for documentation on upgrades.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        teleporterRegistry = TeleporterRegistry(teleporterRegistryAddress);
+        _minTeleporterVersion = teleporterRegistry.getLatestVersion();
+    }
+
+    /**
+     * @dev Updates `_minTeleporterVersion` to the latest version.
+     */
+    function updateMinTeleporterversion() external {
```
michaelkaplan13, line 51:
```suggestion
    function updateMinTeleporterVersion() external {
```

michaelkaplan13: Is `updateMinTeleporterVersion` a function that we want to restrict to "admin only" permission level? 

I think the expected use case is that a dApp would update the minimum Teleporter version once a newer version has been deployed and registered on both chains, and once they are confident that they don't need to receive any messages sent by either chain before it was updated to use the new version. 

Allowing anyone to call it introduces a time window where anyone can force that messages are only received from the latest version prior to the opposite chain being updated to send using the new version.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+pragma solidity 0.8.18;
+
+import "@subnet-evm-contracts/interfaces/IWarpMessenger.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+/**
+ * @dev Implementation of an abstract `WarpProtocolRegistry` contract.
+ *
+ * This implementation is a contract that can be used as a base contract for protocols that are
+ * build on top of Warp. It allows the protocol to be upgraded through a warp out of band message.
```
michaelkaplan13, line 15: Nit: "Warp" is capitalized in once place but not the other.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * @dev See {WarpProtocolRegistry-_addToRegistry}
+     *
+     * Adds the new protocol version address to the `_addressToVersion` mapping
+     * so that the registry can be queried for the version of a given protocol address.
+     * Requirements:
+     *
+     * - `protocolAddress` must not have been previously registered.
```
michaelkaplan13, line 62: Curious what the motivation for this requirement is? IIUC, I think this explicitly prohibits a "rollback" to a previously deployed version, which could be useful in a hypothetical situation where a new version is deployed to introduce a new feature, but contains a regression.

michaelkaplan13: I think we should also explicitly document whether or not changing the address of an existing version is supported. In the current implementation, it looks like this is supported, but we may want to consider disallowing it. Doing so would introduce a similar discrepancy issue between chains where one chain thinks the latest version is X and the other thinks it is Y.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -0,0 +1,74 @@
...
+        returns (ITeleporterMessenger)
+    {
+        return ITeleporterMessenger(_getVersionToAddress(_latestVersion));
+    }
+
+    /**
+     * @dev Gets the version of the given `protocolAddress`.
+     * If `protocolAddress` is not a valid protocol address, returns 0.
+     */
+    function getAddressToVersion(
```
michaelkaplan13, line 49: Nit: Could we rename this something along the lines of `getVersionFromAddress`? The fact that it starts with "getAddress" currently but is actually used to get a version was confusing to me for a minute.

We could make possibly make "get" method names symmetric then by changing `getTeleporter` to something like `getAddressFromVersion`

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+    function getLatestVersion() external view returns (uint256) {
+        return _latestVersion;
+    }
+
+    /**
+     * @dev Gets and verifies for a warp out of band message, and adds the new protocol version
+     * addres to the registry.
+     */
+    function _addProtocolVersion(uint32 messageIndex) internal virtual {
+        // Get and verify a valid warp out of band message.
```
michaelkaplan13, line 108:
```suggestion
        // Get and validate warp out-of-band message.
```

michaelkaplan13: Nit: I think "out-of-band" makes sense to hyphen-ize in all cases.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+        address protocolAddress
+    ) internal virtual {
+        // Check that the version is the increment of the latest version.
+        if (version != _latestVersion + 1) {
+            revert InvalidProtocolVersion();
+        }
+        // Check that the protocol address is a contract address.
+        if (!Address.isContract(protocolAddress)) {
+            revert InvalidProtocolAddress();
+        }
```
michaelkaplan13, line 155: I think it could potentially be useful to be able to add a protocol address prior to the contract being deployed. for instance, if a subnet has an important security patch to a protocol contract, they may want to register the new address prior to deploying the new contract code (which may make it clear what the vulnerability of the existing contract is).

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -0,0 +1,74 @@
...
+
+import "../../WarpProtocolRegistry.sol";
+import "../ITeleporterMessenger.sol";
+
+/**
+ * @dev TeleporterRegistry contract is a {WarpProtocolRegistry} and provides an upgrade
+ * mechanism for teleporter messenger contracts.
+ */
+contract TeleporterRegistry is WarpProtocolRegistry {
+    mapping(address => uint256) internal _addressToVersion;
```
michaelkaplan13, line 16: For my own understanding, what was the motivation to only include the `_addressToVersion` map in the `TeleporterRegistry` rather than putting in the `WarpProtocolRegistery`?

contracts/src/CrossChainApplications/ERC20Bridge/ERC20Bridge.sol
```diff
@@ -456,19 +465,17 @@ contract ERC20Bridge is IERC20Bridge, ITeleporterReceiver, ReentrancyGuard {
...
         address nativeContractAddress,
         address recipient,
         uint256 amount
-    ) private nonReentrant {
-        // Only allow the Teleporter messenger to deliver messages.
-        if (msg.sender != address(teleporterMessenger)) {
-            revert Unauthorized();
-        }
-
+    ) private nonReentrant onlyAllowedTeleporter {
```
michaelkaplan13, line 478: Is the `onlyAllowedTeleporter` modifier needed here? I think it ends up being redundant given that `_mintBridgeTokens` can only be called by `receiveTeleporterMessage` which already has the same modifier (as noted in your comment). Same with `_transferBridgeTokens` below.

I think it could potentially cause unnecessary confusion around where it is needed and where it is not.

michaelkaplan13 (commented):
Incredibly excited to get this in place. I left a handful of comment, but overall very pleased with the design 👍

contracts/src/CrossChainApplications/ERC20Bridge/ERC20Bridge.sol
minghinmatthewlam: I was questioning the same thing, originally added it because thought to make it clear that those functions should only be called by teleporter, but think if it's made clear that these functions are only called by `receiveTelepoterMessage` we can remove for redundancy. Removed.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: hmm I was originally thinking that this would just prevent a subnet from incorrectly adding a protocol address that was previously registered to avoid duplicate, and since they can just set their Teleporter version to an existing version in the case of a regression. But thinking again, I'll likely remove this duplicate protocol address check, so that in the case of a regression we can add a previous version address as the new version, and all dapps will continue with old address when using `getLatestTelepoter`

minghinmatthewlam: I'm assuming you mean change the address of an existing version in the registry? This isn't allowed, when we call `_addToRegistry` the version being registered has to be `_latestVersion+1`, meaning new versions have always be incrementing by 1, and can't specify old version numbers.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: renamed functions

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: removed the check for duplicate protocol address and the error

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: Originally was thinking to keep `WarpProtocolRegistry` as abstract as possible, because wasn't sure if other warp protocols would need it. We use it to check `receiveTeleporterMessage` is above a min teleporter version, but not sure if other protocols would use the same thing.

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: hmm yes that definitely sounds useful, will bring up to the team and see people's thoughts. I agree that for this case we'd probably want to remove the contract check, this was added to just help prevent a wrong protocol address from being specified, but if a significant threshold of the subnet agrees already shouldn't really be needed.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: I agree that it will likely be admin restricted for `updateMinTeleporterVersion` for most dapps, how were you thinking of going about it? There are so many variations of creating admins/governance, and not just by ownership, I was going to leave that up to dapps. Could also explore something like `TeleporterOwnable`.

michaelkaplan13: Good point, dApps could certainly have a bunch of different types of access control logic, and we don't want to pin them down here as far as having to define who an "admin" is. 

Is it possible to have the default be the contract deployer (using OpenZeppelin's [Ownable](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol)), but also make it easy for dApps to override the default behavior? Maybe something like a virtual `onlyUpdater` modifier.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
michaelkaplan13: > I'm assuming you mean change the address of an existing version in the registry? This isn't allowed, when we call _addToRegistry the version being registered has to be _latestVersion+1, meaning new versions have always be incrementing by 1, and can't specify old version numbers.

Gotcha, sounds good. Yeah I missed that previously. Would you mind adding that requirement as a comment here, explaining it's enforced by the `WarpProtocolRegistry` implementation?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
michaelkaplan13: I agree with keeping it as abstract as possible. I just think 2 things are little odd:
1. The mapping address to version (or version to address for that matter) isn't specific to Teleporter. It could apply for any protocol really.
2. (Not directly related) `WarpProtocolRegistry` seems like it shouldn't be housed in the `teleporter` repository, if it's meant to be used in a scope broader than Teleporter. Not sure of the best place to put it or other options here, still thinking.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+    error InvalidProtocolAddress();
+    error InvalidProtocolVersion();
+    error InvalidRegistryInitialization();
+
+    /**
+     * @dev Initializes the contract by setting a `chainID` and `latestVersion`.
+     */
+    constructor(
+        uint256[] memory initialVersions,
+        address[] memory initialProtocolAddresses
```
cam-schultz, line 56: Could we combine these into an array of structs to eliminate index-mismatch user errors?

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+     * addres to the registry.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Requirements:
+     *
+     * - a valid warp out-of-band message must be provided.
+     * - the version must be the increment of the latest version.
+     * - the protocol address must be a contract address.
+     */
+    function addProtocolVersion(uint32 messageIndex) external {
```
cam-schultz, line 80: How do we differentiate an out-of-band Warp message from a generic Warp message? I may be missing something, but it seems like this could be called by anyone via Warp.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -0,0 +1,61 @@
...
+
+// SPDX-License-Identifier: Ecosystem
+
+pragma solidity 0.8.18;
+
+import "./TeleporterRegistry.sol";
+
+/**
+ * @dev TeleporterUpgradeable provides upgrade utility for applications built on top
+ * of the Teleporter protocol by integrating with the {TeleporterRegistry}.
```
cam-schultz, line 12: Can you add a comment that this enforces that the latest Teleporter version registered in the registry will be used?

cam-schultz (commented):
I have a few initial high level questions. Overall, I'm a fan of this pattern for dApps to use Teleporter.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+ * @dev Implementation of an abstract `WarpProtocolRegistry` contract.
+ *
+ * This implementation is a contract that can be used as a base contract for protocols that are
+ * build on top of warp. It allows the protocol to be upgraded through a warp out-of-band message.
+ */
+abstract contract WarpProtocolRegistry {
+    // Address that the out-of-band warp message sets as the "source" address.
+    // The address is obviously not owned by any EOA or smart contract account, so it
+    // can not possibly be the source address of any other warp message emitted by the VM.
+    address public constant VALIDATORS_SOURCE_ADDRESS = address(0);
```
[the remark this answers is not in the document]
minghinmatthewlam: yup always meant to be address(0), helps us differentiate that it was an out of band message and not a regular warp message

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+        if (!valid) {
+            revert InvalidWarpMessage();
+        }
+        if (message.sourceChainID != _chainID) {
+            revert InvalidSourceChainID();
+        }
+        if (message.originSenderAddress != VALIDATORS_SOURCE_ADDRESS) {
+            revert InvalidOriginSenderAddress();
+        }
+        if (message.destinationChainID != _chainID) {
```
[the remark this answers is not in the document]
minghinmatthewlam: yup this is also one of the requirements for interacting with the registry with an out of band message

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: discussed in standup, but it's based on our own requirements for interacting with registry. In this case, sender is `address(0)` which can't be the case for regular warp message, also checking source/destination chain IDs same.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+     * - `protocolAddress` must be a contract address.
+     */
+    function _addToRegistry(
+        uint256 version,
+        address protocolAddress
+    ) internal virtual {
+        // Check that the version is the increment of the latest version.
+        if (version != _latestVersion + 1) {
+            revert InvalidProtocolVersion();
+        }
```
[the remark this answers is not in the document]
minghinmatthewlam: no don't think so

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: added

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: What are your thoughts on removing the `WarpProtocolRegistry` abstract contract, and just having `TeleporterRegistry`? My thoughts are:
- Moving `_addressToVersion` logic to `WarpProtocolRegistry` would leave `TeleporterRegistry` with only a few helper functions separately for Teleporter.
- LIke you mentioned, `WarpProtocolRegistry` seems strange to be in Teleporter repo
- Right now we have no plans of building out other warp protocols, so no additional use of `WarpProtocolRegistry`.
- Would simplify some of the code in terms of not needing to override, call base class methods

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+abstract contract WarpProtocolRegistry {
+    // Address that the out-of-band warp message sets as the "source" address.
+    // The address is obviously not owned by any EOA or smart contract account, so it
+    // can not possibly be the source address of any other warp message emitted by the VM.
+    address public constant VALIDATORS_SOURCE_ADDRESS = address(0);
+
+    WarpMessenger public constant WARP_MESSENGER =
+        WarpMessenger(0x0200000000000000000000000000000000000005);
+
+    bytes32 internal immutable _chainID;
```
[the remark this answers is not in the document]
minghinmatthewlam: think these are fine as `WarpProtocolRegistry` is abstract and meant to be inherited

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: removed the requirement for a contract address

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,181 @@
...
+ * This implementation is a contract that can be used as a base contract for protocols that are
+ * build on top of warp. It allows the protocol to be upgraded through a warp out-of-band message.
+ */
+abstract contract WarpProtocolRegistry {
+    // Address that the out-of-band warp message sets as the "source" address.
+    // The address is obviously not owned by any EOA or smart contract account, so it
+    // can not possibly be the source address of any other warp message emitted by the VM.
+    address public constant VALIDATORS_SOURCE_ADDRESS = address(0);
+
+    WarpMessenger public constant WARP_MESSENGER =
```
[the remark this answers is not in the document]
minghinmatthewlam: > we allow this address to mutate with out-of-band messages

Can you clarify this? I assumed the precompile address would stay the same

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -0,0 +1,69 @@
...
+        external
+        view
+        returns (ITeleporterMessenger)
+    {
+        return ITeleporterMessenger(_getAddressFromVersion(_latestVersion));
+    }
+
+    /**
+     * @dev Gets the version of the given `protocolAddress`.
+     * If `protocolAddress` is not a valid protocol address, returns 0.
```
bernard-avalabs, line 45: Is 0 a valid version? Do we need to differentiate between "not valid protocol address" and version 0 of a valid address?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * @dev See {WarpProtocolRegistry-_addToRegistry}
+     *
+     * Adds the new protocol version address to the `_addressToVersion` mapping
+     * so that the registry can be queried for the version of a given protocol address.
+     * Requirements:
+     *
+     * - `version` must be the increment of the latest version.
```
bernard-avalabs, line 60: Do we need to verify this requirement? Can we determine what the "next" version should be, and if we can, perhaps we shouldn't take an arbitrary version as a parameter.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        for (uint256 i = 0; i < initialVersions.length; i++) {
+            _addToRegistry(initialVersions[i], initialProtocolAddresses[i]);
+        }
+    }
+
+    /**
+     * @dev Gets and verifies a warp out-of-band message, and adds the new protocol version
+     * addres to the registry.
```
bernard-avalabs, line 70: typp: "address"

bernard-avalabs (commented):
Generally makes sense to me. I think I have a better understanding of out-of-band messages now, but I might have a few more questions for you about them later. I've left a few comments regarding versioning.

contracts/src/WarpProtocolRegistry.sol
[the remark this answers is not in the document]
cam-schultz: >  I assumed the precompile address would stay the same

I'm wondering if that's a safe assumption, or if we should account for that potentially changing in the future.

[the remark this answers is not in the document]
minghinmatthewlam: discussed offline and we're going with it as a safe assumption since the precompile address is rooted in the VM, and would require a network upgrade to change the address, and there's no real benefit in changing this address, so we don't forsee it changing

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: The `version` is also acting as a "nonce", without it someone can take an old signed and valid warp out-of-band message then resubmit it to the registry, and it'd be valid through the warp precompile. 

@michaelkaplan13 can you double check my understanding here? Will look to make more clear in documentation.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: 0 is not a valid version as the `_latestVersion` is iniitialized as 0, and we can only add protocol with version `_latestVersion+1`. Added some more comments to help note this, do you think a constant `INVALID_VERSION = 0` would help make more obvious?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: moved `_addressToVersion` logic to `WarpProtocolRegistry`. At this point `TeleporterRegistry` only has helper functions to convert to `ITeleporterMessenger` since dapps will be using the interface for calling it's methods.

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: changed to use `ProtocolRegistryEntry` to avoid index mismatch, good point.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
michaelkaplan13: I agree with all of your thoughts 👍 I thinking unifying them into a single `TeleporterRegistry` implementation at this point makes the most sense.

The only forward-thinking consideration I'm think is if we end up implementing an anycasting protocol as its own contract (say, `TeleporterAnycaster`) or if we ultimately just use warp directly for anycasting. If we were to have a `TeleporterAnycaster` contract, then it'd be nice to be able to use the same registry implementation for it. In that case, it may make sense to have the registry return `address` rather than `ITeleporterMessenger`.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+    error InvalidDestinationAddress();
+    error InvalidProtocolVersion();
+    error InvalidProtocolAddress();
+
+    /**
+     * @dev Initializes the contract by setting a `chainID` and `latestVersion`.
+     */
+    constructor(ProtocolRegistryEntry[] memory initialEntries) {
+        _latestVersion = 0;
+        _chainID = WARP_MESSENGER.getBlockchainID();
```
geoff-vball, line 64: Can we name this variable `blockahinID`? `blockchainID` is used as the unique identifier for a subnet, where `chainID` is an ethereum concept to differentiate between ethereum and ethereum-like chains. I got pretty confused earlier between these two.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * @dev Gets the address of a protocol version.
+     * Requirements:
+     *
+     * - the version must be a valid version.
+     */
+    function getAddressFromVersion(
+        uint256 version
+    ) external view returns (address) {
+        return _getAddressFromVersion(version);
```
geoff-vball, line 95: What's the purpose of directly calling an internal function here and with the other external functions?

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * @dev Gets the latest protocol version.
+     */
+    function getLatestVersion() external view returns (uint256) {
+        return _latestVersion;
+    }
+
+    /**
+     * @dev Gets and verifies for a warp out-of-band message, and adds the new protocol version
+     * addres to the registry.
```
geoff-vball, line 117: typo in address

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        if (entry.protocolAddress == address(0)) {
+            revert InvalidProtocolAddress();
+        }
+
+        _latestVersion++;
+        _versionToAddress[entry.version] = entry.protocolAddress;
+        _addressToVersion[entry.protocolAddress] = entry.version;
+        emit AddProtocolVersion(entry.version, entry.protocolAddress);
```
geoff-vball, line 173: What happens in the scenario where a new subnet is spun up after we're on - say - version 3. Are we guaranteed delivery of all the out-of-band messages? If so, are we guaranteed they're in order?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
geoff-vball: I think we'd probably be okay with just a sparse map. Being able to include add version that we don't already have. 

I don't think we have to worry about replays of old out-of-band messages. Warp messages come with a destination address and destination chain ID, so we could only ever deliver messages meant for this contract.

minghinmatthewlam: The worry is for a replay of a warp out of band message that was sent on this same chain and registry address. i.e. A valid warp message registers address A at version 1, someone can send that same Warp message again to the registry, and it'd be valid unless it includes the version.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: Looking at OZ contract's, each of these abstract contracts have a very specific purpose. I think we should leave `TeleporterUpgradeable` as a base level helper for integrating with the registry. Then for different types of access controls we'd have individual contracts, for example we can have one for `OwnerUpgradeable`, `AdminUpgradeable`, etc.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: hmm with a separate `TeleporterAnycaster` contract I think an option would be to rename `WarpProtocolRegistry` -> `ProtocolRegistry` or `TeleporterRegistry` as base contract, then we'd have separate `TeleporterMessengerRegistry` and `TeleporterAnycasterRegistry`. 

Another option would be to first move everything to `TeleporterRegistry`, then as we develop `TeleporterAnycaster` we refactor as needed, which I don't think is a bad option either.

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: In the constructor there is the ability to initialize the registry with version. So imo there are three options:
1) All existing versions of Teleporter are deployed to the chains, then we initialize the TeleporterRegistry with the same version->address mapping as other subnets.
2) If the subnet doesn't care for the previous existing Teleporter versions, they could have a different version mapping as other subnets. So the Teleporter at address A could be version 3 for subnet A, but is version 1 for subnet B
3) Subnet deploys existing Telepoter versions but don't initialize registry with mappings, then they'd need to update the registry with warp out of band messages, and choose whether they want the same ordering.

The first option would be the option I think we should advocate for, meaning all subnets would then have all version mappings to be the same, but for certain permissioned subnets, they may decide on their own registry mapping/orderings.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
geoff-vball: Yeah, we definitely need to include the version with the warp message. Otherwise there is a risk of registries getting out of sync.

contracts/src/WarpProtocolRegistry.sol
geoff-vball: I don't think we should make a subnet deploy every single version of teleporter if they don't want to.

I **really** don't think we should let different contracts have different versions on different chains. 

We should have a canonical mapping from version <-> address. What do we get from having the versions always start from 1 and being sequential?

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+struct ProtocolRegistryEntry {
+    uint256 version;
+    address protocolAddress;
+}
+
+/**
+ * @dev Implementation of an abstract `WarpProtocolRegistry` contract.
+ *
+ * This implementation is a contract that can be used as a base contract for protocols that are
+ * build on top of warp. It allows the protocol to be upgraded through a warp out-of-band message.
```
cam-schultz, line 23:
```suggestion
 * This implementation is a contract that can be used as a base contract for protocols that are
 * build on top of Warp. It allows the protocol to be upgraded through a Warp out-of-band message.
```

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+ */
+abstract contract WarpProtocolRegistry {
+    // Address that the out-of-band warp message sets as the "source" address.
```
cam-schultz, line 26:
```suggestion
    // Address that the out-of-band Warp message sets as the "source" address.
```

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+    // The address is obviously not owned by any EOA or smart contract account, so it
+    // can not possibly be the source address of any other warp message emitted by the VM.
```
cam-schultz, line 28:
```suggestion
    // cannot possibly be the source address of any other Warp message emitted by the VM.
```

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+        // Get and validate for a warp out-of-band message.
+        (WarpMessage memory message, bool valid) = WARP_MESSENGER
+            .getVerifiedWarpMessage(messageIndex);
+        if (!valid) {
+            revert InvalidWarpMessage();
+        }
+        if (message.sourceChainID != _chainID) {
+            revert InvalidSourceChainID();
+        }
+        if (message.originSenderAddress != VALIDATORS_SOURCE_ADDRESS) {
```
cam-schultz, line 129: Can you add a comment here that this verifies the message is sent as an out-of-band Warp message?

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,195 @@
...
+        emit AddProtocolVersion(entry.version, entry.protocolAddress);
+    }
+
+    /**
+     * @dev Gets the address of a protocol version.
+     * Requirements:
+     *
+     * - `version` must be a valid version, i.e. greater than 0 and not greater than the latest version.
+     */
+    function _getAddressFromVersion(
```
cam-schultz, line 182: This seems like it can be an external view function

contracts/src/WarpProtocolRegistry.sol
cam-schultz: IIUC it's totally acceptable for a `TeleporterRegistry` on a new subnet to have the lowest version be greater than 1 - say it's 3 as Geoff said. Then all dApps that use the registry and that inherit `TeleporterUpgradeable` would have 3 as the minimum version anyway. Perhaps I'm misunderstanding something.

In any case, I don't think a canonical mapping is necessary, since ultimately it's the Teleporter address that determines if the message can be delivered. With this `TeleporterUpgradeable` pattern, the versions are used more as a backwards counter from the latest Teleporter version, rather than a forwards counter from the initial Teleporter version. For example, say a dApp on subnetA and subnetB upgrade from Teleporter 0x123 to 0xabc, but still allow Teleporter 0x123 by not calling `updateMinTeleporterVersion`. All the registry logic should still work, even if the subnets assign different integer versions to the same Teleporter contract address.

cam-schultz: If we decide to go with the approach where all new subnets start from Teleporter version 1 internally, regardless of the global Teleporter version, we could clean up `WarpProtocolRegistry.sol` by not having to include the version number in `_addToRegistry`, as well as not having to include it in the out-of-band message.

minghinmatthewlam: Right now the `TeleporterRegistry` initialization requires that the version starts with 1 AND is incremented, meaning all Teleployer versions would need to be deployed and specified in order. I'm thinking for the initialization to remove these requirements, so you can specify any mapping and doesn't have to be in incremental order. For example if there versions 1,2,3,4,5, and a subnet wanted to just deploy 2,4,5 they can. 

The next question is after the registry is initialized should we require that versions are in order. For a subnet to have the same mapping after initialization as other subnets, it would need to register every new Teleporter version to the registry. How about for subnets that don't want to deploy every version? They would end up with a mismatched registry mapping.

Another solution is to not require incremental versions altogether, so you can add whatever version->address mapping, that way a subnet that doesn't want to deploy a certain version can skip it and not register in the registry. We'd just need to handle cases like if we're adding a version < _latestVersion that we don't update _latestVersion, etc.

minghinmatthewlam: A completely different version mapping would be a set canonical mapping of version->protocolAddress. This could be done by adding a `ITeleporterMessenger.getVersion()`, and a hardcoded constant `VERSION` in `TeleporterMessenger`. So when we want to update registry we pass in the new protocol address, cast and get version `ITeleporterMessenger(protocolAddress).getVersion()`. 

There is a case lets say we have versions 1,2,3, but version 3 has a bug, and we want dapps to use version 2. Then the previous `_minTeleporterVersion` of dapps will stay at 3 while `_latestVersion` is 2. I think there are two ways around this:
1) In `onlyAllowedTeleporter` if  `_minTeleporterVersion > _latestVersion` then we only allow addres of `_latestVersion`.
2) We tell registry to add address for version 2, registry sees that version 2 already is registered, then registers this address as a new version so we have 1,2,3,4 now. The next contract we make should skip 4 and be version 5. We may want separate registry add function for these rollback cases.

minghinmatthewlam: Overall leaning towards not the canonical mapping of registry between subnets. We can remove the requirement to start from version 1 in initialization. We'd be leaving the mapping order of version->protocolAddress up to subnets and their out-of-band warp messages, but still recommend to maintain the same ordering that we push out for versions. Curious to hear thoughts.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
bernard-avalabs: This makes sense to me. I think I understand why the "version" must be included. Thank you for the explanation.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
bernard-avalabs: Extra comments should be enough. It is clear from the rest of the code that 0 is invalid, but it is not clear when just looking at this function.

contracts/src/WarpProtocolRegistry.sol
bernard-avalabs: I have some concerns about mismatched version numbers between different subnets. I understand that it ultimately doesn't matter since only the address matters, but it can lead to some confusion/misunderstanding that can result in incorrect actions in the event of an emergency upgrade. 

What are the reasons to not have the hardcoded VERSION constant that you had suggested?

bernard-avalabs (commented):
Generally looks good. Left my comment on the versioning question.

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: there is a correspnoding `external` function, but we also have this `internal` function so that derived contracts can use it, for example in `TeleporterRegistry`

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: the additional `internal` functions are to allow inherited contracts (ex. `TeleporterRegistry`) to call. If it were only `external` the contracts would have to do `this.foo()` which costs more gas.

contracts/src/WarpProtocolRegistry.sol
cam-schultz: My main concern with a canonical version across subnets was how to guarantee the mapping of version to address was consistent across all subnets. Hardcoding the version in `TeleporterMessenger` would achieve that, as well as reducing the change for confusion as @bernard-avalabs mentioned.

That's an interesting idea for the rollback case. I'm trying to think if there are any odd edge cases that that could introduce. Say we want to rollback to version `2` at address `0x123`, and we're currently on version `3`. The internal mapping would then be `2->0x123` and `4->0x123`. In the future, it we need to rollback to `0x123` again, then either version `2` or `4` would be able to trigger that. Does this present any issues if, say, one subnet rolls back to `0x123` using version `2` while another rolls back using version `4`? Similarly, are there any compatability issues if different subnets have the same address registered at different versions due to inconsistent rollbacks? I think the answer to both is no, but we should consider similar cases.

contracts/src/WarpProtocolRegistry.sol
cam-schultz: Do we need both functions? The `external` function just immediately calls this one.

minghinmatthewlam: We need it in the derived contracts like `TeleporterRegistry`, so that the derived contracts don't need to call `this.getAddressFromVersion` which costs more gas than using internal `_getAddressFromVersion()`

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: Discussed offline, we are going with an approach that removes the requirements for initialization to start from version 1, and for versions added to the registry to be incremental. When versions that are greater than latest version are added, we update the latest version. But users can also add old versions, as long as they were not previously registered, and the registry will simply add the entry without updating latest version.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,193 @@
...
+     * @dev Gets and verifies a warp out-of-band message, and adds the new protocol version
+     * address to the registry.
+     * If a version is greater than the current latest version, it will be set as the latest version.
+     * If a version is less than the current latest version, it is added to the registry, but
+     * doesn't change the latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Requirements:
+     *
+     * - a valid warp out-of-band message must be provided.
```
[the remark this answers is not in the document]
cam-schultz: Same with `sourceChainID` and `destinationChainID` matching `_blockchainID`.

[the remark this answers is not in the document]
cam-schultz: Those might be implicit in the "valid warp out-of-band message" requirement, but none of those are required for a generic out of band message to be valid.

cam-schultz (commented):
LGTM. Just some minor comment nits.

contracts/src/WarpProtocolRegistry.sol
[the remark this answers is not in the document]
minghinmatthewlam: good point, updated the comments

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -0,0 +1,62 @@
+# Upgradeability
+
+## Overview
+
+The `TeleporterMessenger` contract is non-upgradable, once a version of the contract is deployed it cannot be changed. This is with the intention of preventing any changes to the contract that could potentially introduce bugs or vulnerabilities.
+
+However, we do want to be able to provide new features and improvements to the `TeleporterMessenger` contract. To do this we have implemented a `TeleporterRegistry`.
+
+The `TeleporterRegistry` keeps track of a mapping of `TeleporterMessenger` contract versions to their addresses. This allows us to deploy new versions of the `TeleporterMessenger` contract and then update the `TeleporterRegistry` to point to different versions. The `TeleporterRegistry` can only be updated through a Warp out-of-band message that qualifies the following requirements:
```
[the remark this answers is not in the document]
minghinmatthewlam: discussed offline, we are going to stick with "version", since "nonce" is also an overloaded term. Developers might think nonces should only be incremental, but for the registries old versions can be added. Added clarification that version mapping is independent across blockchains, and up to chain to choose their mapping.

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+- `sourceChainAddress` must match `VALIDATORS_SOURCE_ADDRESS = address(0)`
```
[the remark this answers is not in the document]
cam-schultz: Bumping this in case it fell through the cracks

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -2,28 +2,29 @@
...
+- Each registry's mapping of version to contract address is independent of registries on other blockchains, and chains can decide on their own registry mapping entries.
 - `TeleporterRegistry` contract can be initialized through a list of initial registry entries, which are `TeleporterMessenger` contract versions and their addresses.
 - The registry keeps track of a mapping of `TeleporterMessenger` contract versions to their addresses, and vice versa, a mapping of `TeleporterMessenger` contract addresses to their versions.
 - Version zero is an invalid version, and is used to indicate that a `TeleporterMessenger` contract has not been registered yet.
-- Once a version number is registered in the registry, it cannot be changed, but a previous registered protocol address can be added to the registry with a new version.
+- Once a version number is registered in the registry, it cannot be changed, but a previous registered protocol address can be added to the registry with a new version. This is especially important in the case of a rollback to a previous version, the previous Teleporter contract address can be registered with a new version to the registry.
```
cam-schultz, line 23:
```suggestion
- Once a version number is registered in the registry, it cannot be changed, but a previous registered protocol address can be added to the registry with a new version. This is especially important in the case of a rollback to a previous Teleporter version, in which case the previous Teleporter contract address would need to be registered with a new version to the registry.
```

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -2,28 +2,29 @@
 
 ## Overview
```
cam-schultz, line 3: Can you add a paragraph explaining the `getLatestVersion` method? Specifically, we should make it clear that this returns the numerically largest version number registered in the contract. The language can be a bit confusing, since "latest" may be interpreted as the most recent Teleporter version registered, not necessarily the highest number version.

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -43,7 +44,7 @@ contract ERC20Bridge is
...
+The `TeleporterUpgradeable` contract saves the Teleporter registry in a state variable used by the inheriting contract, and initializes a `_minTeleporterVersion` to the latest `TeleporterMessenger` version registered in `TeleporterRegistry`. The `onlyAllowedTeleporter` modifier ensures that `msg.sender` is a `TeleporterMessenger` contract with a version greater than or equal to `_minTeleporterVersion`. This modifier is used to restrict access to functions that should only be called by a `TeleporterMessenger` contract, i.e. `ITeleporterReceiver.receiveTeleporterMessage`. This is to support the case where a dapp wants to upgrade to a new version of the `TeleporterMessenger` contract, but still wants to be able to receive messages from the old Teleporter contract. Then once the dapp completes delivery of messages from the old Teleporter contract, it can call `TeleporterUpgradeable.updateMinTeleporterVersion` to update the `_minTeleporterVersion` to the new version.
```
cam-schultz, line 47:
```suggestion
The `TeleporterUpgradeable` contract saves the Teleporter registry in a state variable used by the inheriting contract, and initializes a `_minTeleporterVersion` to the highest number `TeleporterMessenger` version registered in `TeleporterRegistry`. The `onlyAllowedTeleporter` modifier ensures that `msg.sender` is a `TeleporterMessenger` contract with a version greater than or equal to `_minTeleporterVersion`. This modifier is used to restrict access to functions that should only be called by a `TeleporterMessenger` contract, i.e. `ITeleporterReceiver.receiveTeleporterMessage`. This is to support the case where a dapp wants to upgrade to a new version of the `TeleporterMessenger` contract, but still wants to be able to receive messages from the old Teleporter contract. Then once the dapp completes delivery of messages from the old Teleporter contract, it can call `TeleporterUpgradeable.updateMinTeleporterVersion` to update the `_minTeleporterVersion` to the new version.
```

cam-schultz (commented):
A handful of additional documentation nits for clarity

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -0,0 +1,62 @@
...
+        teleporterRegistry = TeleporterRegistry(teleporterRegistryAddress);
+        _minTeleporterVersion = teleporterRegistry.getLatestVersion();
+    }
+
+    /**
+     * @dev Updates `_minTeleporterVersion` to the latest version.
+     */
+    function updateMinTeleporterVersion() external {
+        _minTeleporterVersion = teleporterRegistry.getLatestVersion();
+    }
```
geoff-vball, line 54: Do we no longer have any option for setting the min version to be anything other than the latest? 

Should this function have some sort of control on it? Right now anyone can call it. I can't imagine we would want that.

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -0,0 +1,66 @@
+# Upgradeability
+
+## Overview
+
+The `TeleporterMessenger` contract is non-upgradable, once a version of the contract is deployed it cannot be changed. This is with the intention of preventing any changes to the deployed contract that could potentially introduce bugs or vulnerabilities.
+
+However, there should still be new versions of the `TeleporterMessenger` contract that provide new features, and `TeleporterRegistry` supports the integration with these new versions.
```
michaelkaplan13, line 7:
```suggestion
However, there could still be new versions of `TeleporterMessenger` deployed in the future. `TeleporterRegistry` is designed to enable applications using a `TeleporterMessenger` instance to integrate newer versions without minimal to no required interaction.
```

I don't really like my wording, but don't think we should specify "new versions that provide new features".

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The `TeleporterRegistry` keeps track of a mapping of `TeleporterMessenger` contract versions to their addresses. This allows us to deploy new versions of the `TeleporterMessenger` contract and then update the `TeleporterRegistry` to point to different versions. The `TeleporterRegistry` can only be updated through a Warp out-of-band message that meets the following requirements:
```
michaelkaplan13, line 9:
```suggestion
The `TeleporterRegistry` maintains a mapping of `TeleporterMessenger` contract versions to their addresses. When a new `TeleporterMessenger` version is deployed, it's address can be added to the `TeleporterRegistry`. The `TeleporterRegistry` can only be updated through a Warp out-of-band message that meets the following requirements:
```

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -0,0 +1,62 @@
...
+ * @dev TeleporterUpgradeable provides upgrade utility for applications built on top
+ * of the Teleporter protocol by integrating with the {TeleporterRegistry}.
+ *
+ * This contract is intended to be inherited by other contracts that wish to use the
+ * upgrade mechanism. It provides a modifier that restricts access to only Teleporter
+ * versions that are greater than or equal to `_minTeleporterVersion`.
+ */
+abstract contract TeleporterUpgradeable {
+    TeleporterRegistry public immutable teleporterRegistry;
+    uint256 internal _minTeleporterVersion;
```
michaelkaplan13, line 20: Is there a reason why `_minTeleporterVersion` couldn't be `public`?

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: This is related to previous discussion here: https://github.com/ava-labs/teleporter/pull/43#discussion_r1344663972

In terms of managing access I think that should be separated from the `TeleporterUpgradeable` contract since there can be variations of access. Thinking to leave it up to the dapp to implement their access pattern and override the `updateMinTeleporterVersion` if needed, or we can look at providing other utility contracts that inherit `TeleporterUpgradeable` like `TeleporterOwnerUpgradeable` `TeleporterAdminUpgradeable`, etc.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -0,0 +1,62 @@
...
+
+        teleporterRegistry = TeleporterRegistry(teleporterRegistryAddress);
+        _minTeleporterVersion = teleporterRegistry.getLatestVersion();
+    }
+
+    /**
+     * @dev Updates `_minTeleporterVersion` to the latest version.
+     */
+    function updateMinTeleporterVersion() external {
+        _minTeleporterVersion = teleporterRegistry.getLatestVersion();
```
michaelkaplan13, line 53: Should we emit an event here?

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,198 @@
...
+
+pragma solidity 0.8.18;
+
+import "@subnet-evm-contracts/interfaces/IWarpMessenger.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+
+/**
+ * @dev Registry entry that represents a mapping between protocolAddress and version.
+ */
+struct ProtocolRegistryEntry {
```
michaelkaplan13, line 14: I wonder if this struct may be more confusing than it's worth in terms of ABI encode an instance of the struct rather than just a `(uint256,address)` tuple. Or is that the same encoding given that there are no variable length elements of the struct?

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
michaelkaplan13: Hm, yeah that makes sense, but in this case dApps that inherit directly from `TeleporterUpgradable` would be vulnerable to having time window where anyone can force that messages are only received from the latest version prior to the opposite chain being updated to send using the new version. I think we should prevent this case by default.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
michaelkaplan13: Sorry for the late reply to this thread!

My slight preference is to avoid the "WarpProtocolRegistry" contract name in the `teleporter` repo and add new implementations of it in the future if needed for new Teleporter protocol contracts, but I don't have a strong opinion either way.

contracts/src/Teleporter/upgrades/README.md
minghinmatthewlam: updated

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: changed to `public`

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: added `MinTeleporterVersionUpdated` event

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: In this case it looks like `abi.encode()` will be the same with these field types, tested by changing unit tests to encode without struct and still decoding with struct. 

The struct also makes initialization simpler by passing an array of structs, instead of two arrays of version and addresses separately, with additional checks like lengths are same etc.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: I am open to this, but if we are moving forward with this, would like it a fast follow separate PR.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @dev Updates `minTeleporterVersion` to the latest version.
+     * Emits a {MinTeleporterVersionUpdated} event.
+     */
+    function updateMinTeleporterVersion() external virtual {
+        uint256 prevVersion = minTeleporterVersion;
+        minTeleporterVersion = teleporterRegistry.getLatestVersion();
+        emit MinTeleporterVersionUpdated(prevVersion, minTeleporterVersion);
+    }
```
geoff-vball, line 65: Should we add a check that the latest version is greater than the current min?

minghinmatthewlam: When adding to registry, latest version is only overriden if new version is greater

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -0,0 +1,63 @@
...
+    ITeleporterReceiver,
+    ReentrancyGuard,
+    TeleporterUpgradeable
+{
+    ...
+    constructor(
+        address teleporterRegistryAddress
+    ) TeleporterUpgradeable(teleporterRegistryAddress) {
+        currentChainID = WarpMessenger(WARP_PRECOMPILE_ADDRESS)
+            .getBlockchainID();
```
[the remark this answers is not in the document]
minghinmatthewlam: this is part of the constructor for the ERC20Bridge, so will be left alone. The warp call is removed for TeleporterMessenger

contracts/src/CrossChainApplications/ExampleMessenger/ExampleCrossChainMessenger.sol
```diff
@@ -124,6 +124,23 @@ contract ExampleCrossChainMessenger is
...
 
+    /**
+     * @dev See {TeleporterUpgradeable-updateMinTeleporterVersion}
+     *
+     * Updates the minimum Teleporter version allowed for receiving on this contract
+     * to the latest version registered in the {TeleporterRegistry}.
+     *
+     * Emits a {MinTeleporterVersionUpdated} event.
+     */
+    function updateMinTeleporterVersion() external override {
```
cam-schultz, line 135: Do we want to make this `onlyOwner` as well? If not, we should explain the use case for why not.

contracts/src/CrossChainApplications/VerifiedBlockHash/BlockHashPublisher.sol
```diff
@@ -7,16 +7,17 @@ pragma solidity 0.8.18;
...
 import "../../Teleporter/upgrades/TeleporterRegistry.sol";
-import "../../Teleporter/upgrades/TeleporterUpgradeable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "./BlockHashReceiver.sol";
 
 /**
  * Contract that publishes the latest block hash of current chain to another chain.
  */
-contract BlockHashPublisher is TeleporterUpgradeable {
+contract BlockHashPublisher {
```
cam-schultz, line 16: Why did you remove `TeleporterUpgradeable` from this contract?

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -4,23 +4,26 @@
...
 
-However, there should still be new versions of the `TeleporterMessenger` contract that provide new features, and `TeleporterRegistry` supports the integration with these new versions.
+However, there could still be new versions of `TeleporterMessenger` contracts needed to be deployed in the future. `TeleporterRegistry` provides applications that use a `TeleporterMessenger` instance a minimal step process to integrate with new versions of `TeleporterMessenger`.
 
-The `TeleporterRegistry` keeps track of a mapping of `TeleporterMessenger` contract versions to their addresses. This allows us to deploy new versions of the `TeleporterMessenger` contract and then update the `TeleporterRegistry` to point to different versions. The `TeleporterRegistry` can only be updated through a Warp out-of-band message that meets the following requirements:
+The `TeleporterRegistry` maintains a mapping of `TeleporterMessenger` contract versions to their addresses. When a new `TeleporterMessenger` version is deployed, it's address can be added to the `TeleporterRegistry`. The `TeleporterRegistry` can only be updated through a Warp out-of-band message that meets the following requirements:
```
cam-schultz, line 9:
```suggestion
The `TeleporterRegistry` maintains a mapping of `TeleporterMessenger` contract versions to their addresses. When a new `TeleporterMessenger` version is deployed, its address can be added to the `TeleporterRegistry`. The `TeleporterRegistry` can only be updated through a Warp out-of-band message that meets the following requirements:
```

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -13,20 +13,25 @@ import "./TeleporterRegistry.sol";
...
+        uint256 indexed newMinTeleporterVersion
+    );
 
     /**
      * @dev Throws if called by a `msg.sender` that is not an allowed Teleporter version.
-     * Checks that `msg.sender` matches a Teleporter version greater than or equal to `_minTeleporterVersion`.
+     * Checks that `msg.sender` matches a Teleporter version greater than or equal to `minTeleporterVersion`.
      */
-    modifier onlyAllowedTeleporter() {
+    modifier onlyAllowedTeleporter() virtual {
```
cam-schultz, line 31: Is the `virtual` keyword necessary? I'm not sure if modifier overriding is allowed, but I don't think we'd want dapps to do so here.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -43,20 +48,19 @@ abstract contract TeleporterUpgradeable {
...
-     * @dev Gets the minimum Teleporter version that is allowed to call this contract.
+     * @dev This is a virtual function that should be overridden to update the `minTeleporterVersion`
+     * allowed for modifier `onlyAllowedTeleporter`, and emit {MinTeleporterVersionUpdated} event after.
+     * The derived contract should call this function after there is a new Teleporter version
+     * registered in the `TeleporterRegistry`, and the derived contract no longer needs to receive
+     * messages from the previous Teleporter versions.
+     *
+     * Note: To prevent anyone from being able to call this function, which would disallow messages
+     * from old Teleporter versions from being received, this function should be safeguarded with access
+     * controls. For example, if the derived contract has an owner/admin, only they can call this function.
```
cam-schultz, line 63: I don't think we should include so much implementation detail here. I'd say the important pieces to note are that this should update the `minTeleporterVersion` used in the modifier, and that it should use the `TeleporterRegistry` to do so. Details such as emitting events and access controls are best documented in the README and example apps.

contracts/src/CrossChainApplications/VerifiedBlockHash/BlockHashPublisher.sol
minghinmatthewlam: The `TeleporterUpgradeable` contract is mostly for the modifier `onlyAllowedTeleporter` and this contract doesn't receive Teleporter messages, only sends them, so don't think it's needed. Now that I look back at it again though, might not hurt to add back in for consistency.

minghinmatthewlam: Oh yea and also inheriting `TeleporterUpgradeable` also means the contract needs to override `updateMinTeleporterVersion` and I think that's unnecessary for contracts that only send.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: modifiers can be overriden, but you're right that we should remove it for the modifier.

contracts/src/CrossChainApplications/ExampleMessenger/ExampleCrossChainMessenger.sol
minghinmatthewlam: added to documentation noting this is an example cross chain messenger, and that anyone would be able to call this function, disallowing receiving of old Teleporter version messages.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: I think it's still good to point out that to protect this call from being called by anyone it should be implemented by access controls. Agreed to remove the part in the comment about how this function should be called after a new version is registered in TeleporterRegistry, etc, as this is written in README.

contracts/src/CrossChainApplications/ERC20Bridge/ERC20Bridge.sol
```diff
@@ -326,6 +324,23 @@ contract ERC20Bridge is IERC20Bridge, ITeleporterReceiver, ReentrancyGuard {
         }
     }
 
+    /**
+     * @dev See {TeleporterUpgradeable-updateMinTeleporterVersion}
+     *
+     * Updates the minimum Teleporter version allowed for receiving on this contract
+     * to the latest version registered in the {TeleporterRegistry}. Also restricts this function to
+     * the owner of this contract.
```
michaelkaplan13, line 332:
```suggestion
     * to the latest version registered in the {TeleporterRegistry}.
     * Restricted to only owners of the contract.
```

contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
```diff
@@ -609,6 +624,72 @@ contract ERC20BridgeTest is Test {
...
+            ),
+            abi.encode(2)
+        );
+        vm.expectCall(
+            MOCK_TELEPORTER_REGISTRY_ADDRESS,
+            abi.encodeWithSelector(
+                WarpProtocolRegistry.getLatestVersion.selector
+            )
+        );
+        erc20Bridge.updateMinTeleporterVersion();
```
michaelkaplan13, line 654: Can we `expectEmit` of the specific event here, and also check the value of the minimum version is the expected value afterwards?

contracts/src/CrossChainApplications/ExampleMessenger/ExampleCrossChainMessenger.sol
michaelkaplan13: I generally think we should avoid using any patterns in the example that we wouldn't encourage others to use for real projects. Allowing anyone to call this method is potentially dangerous, so would heir on the side of allowing no one to call it, or it being `onlyOwner`

contracts/src/CrossChainApplications/VerifiedBlockHash/BlockHashPublisher.sol
cam-schultz: Ah gotcha, that makes sense. I'm fine leaving `TeleporterUpgradeable` out.

contracts/src/CrossChainApplications/VerifiedBlockHash/BlockHashReceiver.sol
```diff
@@ -88,6 +88,23 @@ contract BlockHashReceiver is ITeleporterReceiver {
         }
     }
 
+    /**
+     * @dev See {TeleporterUpgradeable-updateMinTeleporterVersion}
+     *
+     * Updates the minimum Teleporter version allowed for receiving on this contract
+     * to the latest version registered in the {TeleporterRegistry}. Also restricts this function to
+     * the owner of this contract.
```
michaelkaplan13, line 96:
```suggestion
     * to the latest version registered in the {TeleporterRegistry}.
     * Restricted to only owners of the contract.
```

Nit on the comment wording

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,198 @@
...
+     * Requirements:
+     *
+     * - a valid Warp out-of-band message must be provided.
+     * - Message source chain ID must be the same as the blockchain ID of the registry.
+     * - Message origin sender address must be the same as the `VALIDATORS_SOURCE_ADDRESS`.
+     * - Message destination chain ID must be the same as the blockchain ID of the registry.
+     * - Message destination address must be the same as the address of the registry.
+     * - the version must not be zero.
+     * - the version must not already be registered.
+     * - protocol address must not be zero address.
```
michaelkaplan13, line 81: Nit: Some the first letters are capitalized and others are not.

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -0,0 +1,198 @@
...
+        // Check that the message is sent through a warp out of band message.
+        require(
+            message.originSenderAddress == VALIDATORS_SOURCE_ADDRESS,
+            "WarpProtocolRegistry: invalid origin sender address"
+        );
+        require(
+            message.destinationChainID == _blockchainID,
+            "WarpProtocolRegistry: invalid destination chain ID"
+        );
+        require(
```
michaelkaplan13, line 101: I'm think of a future event when we want to roll out a TeleporterMessenger upgrade. It would be really nice if the registry contract was deployed to the same contract address on most (all) chains such that the out-of-band warp message needed to be signed by the validators of each chain was the same.

We could deploy the registry using Nick's method, but to make that less error prone, similar to the `TeleporterMessenger` contract itself, we should move external calls out of the constructor to make it less likely for the deployment to fail. Thoughts on always calling `warp.getBlockchainID` here rather than storing the blockchain ID in the contract?

contracts/src/WarpProtocolRegistry.sol
michaelkaplan13: Is there a significant gas savings to this approach as opposed to making `getAddressFromVersion` and `getVersionFromAddress` both be `public`? My understanding is that `public` is essentially the same as `external + internal`, which is the behavior we want here.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
cam-schultz: Sounds good to me 👍

contracts/src/CrossChainApplications/ExampleMessenger/ExampleCrossChainMessenger.sol
minghinmatthewlam: agreed, changed `ExampleCrossChainMessenger` to be ownable as well

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: Added issue and extra considerations in new issue: https://github.com/ava-labs/teleporter/issues/76

contracts/src/WarpProtocolRegistry.sol
minghinmatthewlam: Haven't tested the gas and whether it's significant, but would guess it's nothing significant. The `internal` also allows derived contracts to override, but added a new issue, where it should also revisit the `external` + `internal` vs `public`. 

Any reason for opposing external + internal? Maybe it complicates the contract more, or not sure if it's in regards to increasing contract size?

cam-schultz (dismissed):
LGTM! Thanks for creating the followup issues. Excited to get this merged 👍

geoff-vball (approved):
Looks good!

geoff-vball (dismissed):

cam-schultz (dismissed):

contracts/src/WarpProtocolRegistry.sol
michaelkaplan13: Could it be `public virtual` to allow for inheriting contracts to override it still?

And no strong reason, but it increase the size of the contract, and increase the mental load on someone trying to understand the contract (they have to remember or reason about why the function is duplicated in the way that it is).

michaelkaplan13 (dismissed):
LGTM

bernard-avalabs (dismissed):
LGTM

contracts/src/CrossChainApplications/VerifiedBlockHash/BlockHashPublisher.sol
```diff
@@ -52,18 +59,20 @@ contract BlockHashPublisher {
...
-                    contractAddress: address(0),
-                    amount: 0
-                }),
-                requiredGasLimit: RECEIVE_BLOCK_HASH_REQUIRED_GAS_LIMIT,
-                allowedRelayerAddresses: new address[](0),
-                message: messageData
-            })
-        );
+        messageID = teleporterRegistry
+            .getLatestTeleporter()
```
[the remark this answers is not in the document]
minghinmatthewlam: no don't think so, any particular use case you're thinking of?

scripts/local/integration-tests/erc20_bridge_multihop.sh
```diff
@@ -39,19 +43,22 @@ native_erc20_deploy_result=$(forge create --private-key $user_private_key src/Mo
 native_erc20_contract_address=$(parseContractAddress "$native_erc20_deploy_result")
 echo "Test ERC20 contract deployed to $native_erc20_contract_address on Subnet A"
 
-# Deploy the ERC20 bridge contract to both chains.
-bridge_a_deploy_result=$(forge create --private-key $user_private_key --constructor-args $teleporter_contract_address \
-    --rpc-url $subnet_a_url src/CrossChainApplications/ERC20Bridge/ERC20Bridge.sol:ERC20Bridge)
+# Deploy the ERC20 bridge contract to all chains.
+bridge_a_deploy_result=$(forge create --private-key $user_private_key \
```
michaelkaplan13, line 47: Was there a need to move `--constructor-args` to be the last argument here with the last foundry version, or did you just move it to be consistent in our usage?

michaelkaplan13 (approved):
👍

scripts/local/integration-tests/erc20_bridge_multihop.sh
minghinmatthewlam: Was not due to last version, but on their page they say "The --constructor-args flag must be positioned last in the command, since it takes multiple values.", even though it seemed to work previously without being last position. https://book.getfoundry.sh/reference/forge/forge-create

PR #51, opened 2023-10-05
Block hash publisher

cam-schultz:
## Why this should be merged
Adds a basic contract that verifies a block hash from another chain, and emits it in an event, using the `subnet-evm` features described in https://github.com/ava-labs/subnet-evm/issues/734.

## How this was tested
- [x] E2E tests in https://github.com/ava-labs/awm-relayer/pull/54
- [ ] Unit tests


contracts/src/Teleporter/TeleporterBlockHashReceiver.sol
```diff
@@ -0,0 +1,50 @@
...
+        uint32 messageIndex,
+        bytes32 sourceChainID
+    ) external receiverNonReentrant {
+        // Verify and parse the block hash payload included in the transaction access list
+        // using the warp message precompile.
+        (WarpBlockHash memory warpBlockHash, bool success) = WARP_MESSENGER
+            .getVerifiedWarpBlockHash(messageIndex);
+
+        if (!success) {
+            revert InvalidWarpBlockHash();
```
geoff-vball, line 39: `InvalidWarpBlockHash` sounds like we're verifying the hash itself here. This would make more sense to me as something like `InvalidWarpMessage`

geoff-vball (commented):
One question about the name of an error. Otherwise LGTM

contracts/src/Teleporter/TeleporterBlockHashReceiver.sol
```diff
@@ -0,0 +1,50 @@
...
+ */
+contract TeleporterBlockHashReceiver is ITeleporterBlockHashReceiver, ReentrancyGuards {
+    WarpMessenger public constant WARP_MESSENGER =
+        WarpMessenger(0x0200000000000000000000000000000000000005);
+
+    error InvalidWarpBlockHash();
+    error UnexpectedSourceChainID();
+
+    /**
+     * @dev See {ITeleporterBlockHashReceiver-receiveCrossChainMessage}
```
minghinmatthewlam, line 25:
```suggestion
     * @dev See {ITeleporterBlockHashReceiver-receiveBlockHash}
```

michaelkaplan13:
Lets discuss whether or not we think these contracts belong in the `teleporter` repo. I understand the motivation (warp-related EVM contract), but it is the first time we'd be introducing contracts that don't relate to the `TeleporterMessenger` protocol. 

Part of me thinks that the interface that the `awm-relayer` will call to deliver a block hash could live in the `awm-relayer` repo itself, and we could avoid any dependencies here.

contracts/src/Teleporter/ITeleporterBlockHashReceiver.sol
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * @dev Receives a block hash from another chain.
+     *
+     * The message specified by `messageIndex` must be provided at that index in the access list storage slots of the transaction,
+     * and is verified in the precompile predicate.
+     */
+    function receiveBlockHash(
+        uint32 messageIndex,
+        bytes32 sourceChainID
```
michaelkaplan13, line 28: It feels a bit counter intuitive to me to have the `sourceChainID` as a parameter here when the verified value is contained in the warp message itself. Could we remove it all together to simplify the interface?

bernard-avalabs (commented):
Generally makes sense to me.

contracts/src/Teleporter/ITeleporterBlockHashReceiver.sol
cam-schultz: Agreed and removed

geoff-vball (dismissed):
LGMT 👍 

`network_setup.go` seems to be getting slightly unwieldy. I'll follow up this PR with a bit of a refactor that I've already mostly done in my native bridge branch. That way I can take it out of the native bridge PR to simplify it.

michaelkaplan13:
Closing this as stale, but the idea is still planned to be use as a part of [ACP-108](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/108-evm-event-importing).

PR #52, opened 2023-10-06
Remove awm-relayer dependency

gwen917:
## Why this should be merged
- Remove `awm-relayer` dependency in `teleporter` repo
- Use the ABI generated from abigen.

## How this works
- The Teleporter packing/unpacking and utility functions are moved from `awm-relayer` to `teleporter`. This PR updates the usage in `awm-relayer`: https://github.com/ava-labs/awm-relayer/pull/61
- The ABI bindings and utilities are also organized into the following directory structure:
```
teleporter/
 |--- go-utils/
 |        |--- abi-bindings/
 |        |        |--- Teleporter/
 |        |        |         |--- TeleporterMessenger/
 |        |        |         |        |---TeleporterMessenger.go
 |        |        |         |        |--- packing.go
 |        |--- utils/
 |        |        |--- deployment_utils.go
 |        |        |--- gas_utils.go
```
Bindings for future languages can be added in the future under a similar directory structure.

## How this was tested
- CI: unit tests, end to end tests


tests/utils/abis.go
```diff
@@ -0,0 +1,72 @@
...
+
+	teleportermessenger "github.com/ava-labs/teleporter/abi-bindings/Teleporter/TeleporterMessenger"
+)
+
+type teleporterABI struct {
+	teleporterMessengerABI abi.ABI
+}
+
+func NewTeleporterABI() *teleporterABI {
+	parsed, err := abi.JSON(strings.NewReader(teleportermessenger.TeleporterMessengerABI))
```
geoff-vball, line 20: We should be able to call something like `teleportermessenger.TeleporterMessengerMetaData.GetAbi()` to get the same result here.

tests/utils/utils.go
```diff
@@ -365,3 +368,32 @@ func calculateTxParams(ctx context.Context, wsClient ethclient.Client, fundedAdd
 
 	return gasFeeCap, gasTipCap, nonce
 }
+
+// CalculateReceiveMessageGasLimit calculates the estimated gas amount used by a single call
+// to receiveCrossChainMessage for the given message and validator bit vector. The result amount
+// depends on the required limit for the message execution, the number of validator signatures
+// included in the aggregate signature, the static gas cost defined by the precompile, and an
+// extra buffer amount defined here to ensure the call doesn't run out of gas.
+func calculateReceiveMessageGasLimit(numSigners int, executionRequiredGasLimit *big.Int) (uint64, error) {
```
cam-schultz, line 377: We should put this in a application (not test) utility package and export this function. I'm thinking we could have the following directory structure:
```
teleporter/
 |--- go/
 |        |--- abi-bindings/
 |        |        |--- Teleporter/
 |        |        |         |--- TeleporterMessenger/
 |        |        |         |        |---TeleporterMessenger.go
 |        |        |         |        |--- packing.go
 |        |--- teleporter-utilities/
 |        |        |--- utils.go
```

tests/utils/utils.go
```diff
@@ -32,39 +33,47 @@ import (
...
-	RequiredGasLimit        *big.Int
-	Message                 []byte
-	AllowedRelayerAddresses []common.Address
-}
-
-type FeeInfo struct {
-	ContractAddress common.Address
-	Amount          *big.Int
-}
+// type SendCrossChainMessageInput struct {
```
cam-schultz, line 64: Let's remove commented out code.

tests/utils/utils.go
```diff
@@ -32,39 +33,47 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+const (
+	receiveCrossChainMessageStaticGasCost           uint64 = 2_000_000
+	receiveCrossChainMessageGasCostPerAggregatedKey uint64 = 1_000
+	receiveMessageGasLimitBufferAmount                     = 100_000
+)
+
```
cam-schultz, line 41: We can move these to the `utils.go` file I mentioned below.

tests/utils/network_setup.go
```diff
@@ -23,7 +22,9 @@ import (
 )
 
 const (
-	fundedKeyStr = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
+	fundedKeyStr         = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
+	baseFeeFactor        = 2
+	maxPriorityFeePerGas = 2500000000 // 2.5 gwei
```
cam-schultz, line 27: We can move these to the `utils.go` file I mentioned below.

tests/utils/abis.go
cam-schultz, line 1: Let's put the packing and unpacking methods in the same package as the generate `TeleporterMessenger` bindings. We can also remove the `teleporterABI` receiver, and get the `abi.ABI` directly from the `TeleporterMessengerMetaData` (see comment below).

tests/utils/network_setup.go
geoff-vball, line 27: Could you add comments here as to why these constants are what they are?

tests/utils/utils.go
```diff
@@ -32,39 +33,47 @@ import (
...
+// 	FeeInfo                 FeeInfo
+// 	RequiredGasLimit        *big.Int
+// 	Message                 []byte
+// 	AllowedRelayerAddresses []common.Address
+// }
+
+// type FeeInfo struct {
+// 	ContractAddress common.Address
+// 	Amount          *big.Int
+// }
```
geoff-vball, line 76: I think this can be removed now

cam-schultz (commented):
Let's go with the following directory structure:
```
teleporter/
 |--- go/
 |        |--- abi-bindings/
 |        |        |--- Teleporter/
 |        |        |         |--- TeleporterMessenger/
 |        |        |         |        |---TeleporterMessenger.go
 |        |        |         |        |--- packing.go
 |        |--- teleporter-utilities/
 |        |        |--- utils.go
```

We can put the packing/unpacking functions in `packing.go` and extra utilities in `utils.go`

geoff-vball (commented):
Just a few small comments. Thanks for doing this!

go/abi-bindings/Teleporter/TeleporterMessenger/packing.go
```diff
@@ -0,0 +1,87 @@
...
+func PackMessageReceivedOutput(success bool) ([]byte, error) {
+	abi, err := TeleporterMessengerMetaData.GetAbi()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get abi")
+	}
+
+	return abi.PackOutput("messageReceived", success)
+}
+
+// CAUTION: PackEvent is documented as not supporting struct types, so this should only be used for testing puposes.
```
bernard-avalabs, line 76: typo: "purposes"

bernard-avalabs (approved):
LGTM

go/teleporter-utilities/utils.go
```diff
@@ -0,0 +1,46 @@
...
+	"errors"
+	"math/big"
+
+	"github.com/ava-labs/avalanchego/utils/math"
+)
+
+const (
+	receiveCrossChainMessageStaticGasCost           uint64 = 2_000_000
+	receiveCrossChainMessageGasCostPerAggregatedKey uint64 = 1_000
+	receiveMessageGasLimitBufferAmount              uint64 = 100_000
```
cam-schultz, line 13: Let's export these

cam-schultz (commented):
All the changes LGTM. 

Can you add a `README` under `go/abi-bindings` describing the directory contents? Specifically, note that the folders/files with the same names as the Solidity contracts are created by the `abi_go_bindings.sh` script, and also note that the `packing.go` files are written manually on a case-by-case basis, since `abigen` does not generate them out of the box.

michaelkaplan13:
Looks like there are still a few open comments from @cam-schultz on this PR.

The only additional one I have is that it's a little odd as-is because `contract-deployment/` contains golang files that are not under the `go/` directory. I'm open to either moving it under `go/`, or removing the `go/` directory all together and moving `abi-bindings/` back to the top level. Curious if anyone has opinions here? 

cc @bernard-avalabs @geoff-vball @minghinmatthewlam

minghinmatthewlam:
> > The only additional one I have is that it's a little odd as-is because `contract-deployment/` contains golang files that are not under the `go/` directory. I'm open to either moving it under `go/`, or removing the `go/` directory all together and moving `abi-bindings/` back to the top level. Curious if anyone has opinions here?
> 
> My thought was to structure the directories in such a way that would allow bindings to other languages, not just Go. However, a better layout would be putting `go` beneath `abi-bindings`, like so: `abi-bindings/go-bindings`. Other language bindings would be side-by-side with `go-bindings`. If we go this route, I definitely don't think the deployment utils should be in `go-bindings`.
> 
> It seems reasonable that we may want to support other language bindings for Teleporter in the future, especially as Teleporter development expands beyond in-house. I'd advocate for having the directory structure we go with now reflect that.

It seems strange to me to move the abi bindings under a `go` directory when we have the `go.mod` and `go.sum` in the root of the repo, and the e2e tests/contract deployments are also in go. I feel part of the repo is factoring in other languages in the future, while other parts only support go. Would consider removing the separation of `go` directories for now, and defer refactoring to when we do add other language support.

go-utils/abi-bindings/Teleporter/TeleporterMessenger/packing_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"bytes"
+	"math/big"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func testTeleporterMessage(messageID int64) TeleporterMessage {
```
michaelkaplan13, line 15: Nit: Can we name this `createTestTeleporterMessage` so that it's more clear that it's not a test itself?

michaelkaplan13:
> It seems strange to me to move the abi bindings under a go directory when we have the go.mod and go.sum in the root of the repo, and the e2e tests/contract deployments are also in go. I feel part of the repo is factoring in other languages in the future, while other parts only support go. Would consider removing the separation of go directories for now, and defer refactoring to when we do add other language support.

I tend to agree now that you call that out. Cam noted that the [go.mod for blst](https://github.com/supranational/blst) is in the root directory of their repository even when all of the golang code is in a subdirectory, but the e2e tests and contract deployment parts are definitely a little odd where they live currently. I'd be on board with either moving both of them under `go/` (removing the "-utils"), or getting rid of `go-utils/` all together for now and deferring the refactoring for other adding other languages to when that time comes.

cam-schultz:
> I tend to agree now that you call that out. Cam noted that the [go.mod for blst](https://github.com/supranational/blst) is in the root directory of their repository even when all of the golang code is in a subdirectory, but the e2e tests and contract deployment parts are definitely a little odd where they live currently. I'd be on board with either moving both of them under `go/` (removing the "-utils"), or getting rid of `go-utils/` all together for now and deferring the refactoring for other adding other languages to when that time comes.

I'd strongly prefer to decide on a directory structure that takes into account other language bindings now, rather than introduce breaking changes to downstream dependencies when we decide to add additional language bindings. How about we move the contract deployment app and the test suites+utilities under `go/`, and put the bindings under `abi-bindings/go`? That way `abi-bindings` can be easily expanded to include other languages in the future. I think this directory structure would only be confusing with a single language binding, but if `abi-bindings` had multiple subdirectories, one of which is `go/`, it would be clear what that directory represents.

michaelkaplan13:
All the changes LGTM, would you also mind adding this README that you mentioned prior?

> 
Can you add a README under go/abi-bindings describing the directory contents? Specifically, note that the folders/files with the same names as the Solidity contracts are created by the abi_go_bindings.sh script, and also note that the packing.go files are written manually on a case-by-case basis, since abigen does not generate them out of the box.

michaelkaplan13 (dismissed):
LGTM!

.github/workflows/abi_bindings_checker.yml
```diff
@@ -33,7 +33,24 @@ jobs:
...
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Checkout repositories and submodules
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Run ABI Binding Unit Tests
```
minghinmatthewlam, line 55: nit: these are not just abi binding unit tests right?

abi-bindings/README.md
```diff
@@ -0,0 +1 @@
+This directory contains ABI bindings for the Solidity contracts in the `contracts/src/CrossChainApplications` and `contracts/src/Teleporter` directories. The files with the same name as the Solidity source files are automatically generated by the `scripts/abi_bindings.sh` script. Other files in this directory (such as the packing utilities) are manually created and maintained.
```
minghinmatthewlam, line 1: nit: can we add a top level heading for `README`s

cam-schultz: Added

.github/workflows/abi_bindings_checker.yml
cam-schultz: Good catch. Added a `cd abi-bindings/go` so it will only run the ABI bindings unit tests. Running `go test ./...` will call the E2E tests. They won't run since the necessary env var isn't set, but that's still not what we're targeting here.

michaelkaplan13 (dismissed):

minghinmatthewlam (approved):
LGTM

michaelkaplan13 (approved):

PR #53, opened 2023-10-06
Add fee and relayer info to events

michaelkaplan13:
## Why this should be merged & How this works

Currently, the fee/incentive information for a given message is not included in the event emitted when the message is submitted. In order for relayers or anyone else to check that information for the message, they must make an RPC call and use `getFeeInfo` for the message. This makes the data more difficult to index properly, and also introduces unnecessary steps and latency for relayers to determine whether or not they should relay a given message. In this PR, we add the fee information for the message into the `SendCrossChainMessage` event, such that anyone listening/indexing that event will be able to view the fee information immediately.

Additionally, the current contracts don't provide an easy way to view who delivered a specific message, or what address is able to redeem the reward for it. This could be useful metrics to have regarding relayer speed and usage. These values are added to the `ReceiveCrossChainMessage` event in this PR. The `deliverer` is the `msg.sender` of the `receiveCrossChainMessage` call, and the `rewardRedeemer` is the address that the `deliverer` specifies is allowed to redeem the reward on the source chain once the receipt is returned.

## How this was tested

Unit tests

## How is this documented

N/A


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -142,17 +142,17 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         TeleporterMessage calldata message
     ) external senderNonReentrant {
         // Get the previously sent message hash.
-        bytes32 messageHash = sentMessageInfo[destinationChainID][
-            message.messageID
-        ].messageHash;
-        if (messageHash == bytes32(0)) {
+        SentMessageInfo memory existingMessageInfo = sentMessageInfo[
+            destinationChainID
+        ][message.messageID];
```
cam-schultz, line 147: Do you know if there is any significant gas difference in bringing the entire `SentMessageInfo` into `memory`? I'm not sure how well the Solidity compiler is able to optimize the existing case of accessing the `messageHash` inline.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -644,18 +647,20 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
 
         // Store the fee asset and amount to be paid to the relayer of this message upon receiving the receipt.
         // Also store the message hash so that it can be retried until we get receipt of its delivery.
+        TeleporterFeeInfo memory adjustedFeeInfo = TeleporterFeeInfo({
+            contractAddress: feeInfo.contractAddress,
+            amount: adjustedFeeAmount
+        });
```
cam-schultz, line 653: Similar question here regarding gas differences when declaring the `TeleporterFeeInfo` as `memory` vs. constructing it inline in the `SentMessageInfo` construction.

cam-schultz (dismissed):
LGTM, just a couple of questions regarding gas usage.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Good question. 

When running `forge test --gas-report` on the current version in `main`, I get:
```
| src/Teleporter/TeleporterMessenger.sol:TeleporterMessenger contract |                 |        |        |        |         |
|---------------------------------------------------------------------|-----------------|--------|--------|--------|---------|
| Deployment Cost                                                     | Deployment Size |        |        |        |         |
| 2174481                                                             | 10854           |        |        |        |         |
| Function Name                                                       | min             | avg    | median | max    | # calls |
| addFeeAmount                                                        | 3486            | 5545   | 4070   | 9507   | 7       |
| checkIsAllowedRelayer                                               | 950             | 1121   | 950    | 1643   | 5       |
| checkRelayerRewardAmount                                            | 764             | 764    | 764    | 764    | 8       |
| getFeeInfo                                                          | 851             | 1651   | 851    | 4851   | 5       |
| getMessageHash                                                      | 643             | 928    | 643    | 2643   | 7       |
| getNextMessageID                                                    | 660             | 1993   | 2660   | 2660   | 3       |
| getReceiptAtIndex                                                   | 903             | 1038   | 903    | 1616   | 21      |
| getReceiptQueueSize                                                 | 730             | 3896   | 4730   | 4730   | 24      |
| getRelayerRewardAddress                                             | 629             | 754    | 629    | 2629   | 16      |
| messageReceived                                                     | 673             | 839    | 673    | 2673   | 12      |
| receiveCrossChainMessage                                            | 609             | 87978  | 101991 | 178925 | 51      |
| redeemRelayerRewards                                                | 1946            | 2230   | 2008   | 2736   | 3       |
| retryMessageExecution                                               | 3748            | 16726  | 9318   | 57861  | 11      |
| retryReceipts                                                       | 8634            | 37784  | 40410  | 64310  | 3       |
| retrySendCrossChainMessage                                          | 5440            | 7799   | 7753   | 10206  | 3       |
| sendCrossChainMessage                                               | 34540           | 101038 | 111478 | 137275 | 32      |
```

And when running it on this branch, I get:
```
| src/Teleporter/TeleporterMessenger.sol:TeleporterMessenger contract |                 |       |        |        |         |
|---------------------------------------------------------------------|-----------------|-------|--------|--------|---------|
| Deployment Cost                                                     | Deployment Size |       |        |        |         |
| 2215325                                                             | 11058           |       |        |        |         |
| Function Name                                                       | min             | avg   | median | max    | # calls |
| addFeeAmount                                                        | 3486            | 5547  | 4070   | 9513   | 7       |
| checkIsAllowedRelayer                                               | 954             | 1125  | 954    | 1647   | 5       |
| checkRelayerRewardAmount                                            | 764             | 764   | 764    | 764    | 8       |
| getFeeInfo                                                          | 851             | 1651  | 851    | 4851   | 5       |
| getMessageHash                                                      | 643             | 928   | 643    | 2643   | 7       |
| getNextMessageID                                                    | 660             | 1993  | 2660   | 2660   | 3       |
| getReceiptAtIndex                                                   | 903             | 1038  | 903    | 1616   | 21      |
| getReceiptQueueSize                                                 | 730             | 3896  | 4730   | 4730   | 24      |
| getRelayerRewardAddress                                             | 629             | 754   | 629    | 2629   | 16      |
| messageReceived                                                     | 673             | 839   | 673    | 2673   | 12      |
| receiveCrossChainMessage                                            | 609             | 88430 | 102587 | 179522 | 51      |
| redeemRelayerRewards                                                | 1952            | 2234  | 2014   | 2736   | 3       |
| retryMessageExecution                                               | 3748            | 16664 | 9247   | 57790  | 11      |
| retryReceipts                                                       | 8634            | 38256 | 41118  | 65018  | 3       |
| retrySendCrossChainMessage                                          | 5692            | 9533  | 10824  | 12085  | 3       |
| sendCrossChainMessage                                               | 34546           | 99304 | 108115 | 137991 | 32      |
```

Not very significant either, but it seems the take aways are that:
- Declaring the `SentMessageInfo memory` variable here costs more gas than just a `bytes32` value. This makes sense to me. The resulting gas cost are still reasonable IMO.
- Declaring the `TeleporterFeeInfo` as `memory` in `_sendCrossChainMessage` rather than constructing it inline seems to be an incremental gas savings. This is surprising to me, but I'll take it 😅. Need to look up why it is the case.

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -83,7 +84,9 @@ interface ITeleporterMessenger {
     event ReceiveCrossChainMessage(
         bytes32 indexed originChainID,
         uint256 indexed messageID,
-        TeleporterMessage message
+        TeleporterMessage message,
+        address deliverer,
+        address rewardRedeemer
```
minghinmatthewlam, line 89: would we want to index `rewardRedeemer`?

minghinmatthewlam: or could be `deliverer`, I'm thinking indexing could help relayers check which messages they successfully deliver.

bernard-avalabs (dismissed):
LGTM

contracts/src/Teleporter/ITeleporterMessenger.sol
michaelkaplan13: Yeah that's a great call. I think indexing `deliverer` makes sense so you can query for messages delivered by a specific address. 

To that end, I also added a `RelayerRewardsRedeemed` event (with the redeemer indexed) for when `redeemRelayerRewards` is called successfully.

minghinmatthewlam (dismissed):
LGTM, left question

contracts/src/Teleporter/tests/RedeemRelayerRewardsTests.t.sol
```diff
@@ -84,6 +84,12 @@ contract RedeemRelayerRewardsTest is TeleporterMessengerTest {
                 (feeRewardInfo.relayerRewardAddress, feeRewardInfo.feeAmount)
             )
         );
+        vm.expectEmit(true, true, true, true, address(teleporterMessenger));
```
minghinmatthewlam, line 87: do we need to check for all topics if the event is only indexing topic 1

minghinmatthewlam (approved):
LGTM

contracts/src/Teleporter/tests/RedeemRelayerRewardsTests.t.sol
michaelkaplan13: We don't necessarily _need_ to, but I think it is safest to always check for all topics to match by default. The only case where this isn't possible is if the topic value for a specific log emission is not known beforehand, but otherwise I would rather any changed behavior be more likely to be caught by the units tests.

minghinmatthewlam: agreed, we should check all topics for the `expectEmit` in case of future changes.

cam-schultz (commented):
One small suggestion, otherwise LGTM

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -84,9 +84,18 @@ interface ITeleporterMessenger {
...
+        address rewardRedeemer,
+        TeleporterMessage message
+    );
+
+    /**
+     * @dev Emitted when an account redeems accumulated relayer rewards.
+     */
+    event RelayerRewardsRedeemed(
+        address indexed redeemer,
+        address asset,
```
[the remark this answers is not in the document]
michaelkaplan13: Yeah I agree. No reason not to especially given the expected relative infrequency of relayer redeeming rewards, so the marginally increased gas cost is insignificant.

I added this now.

cam-schultz (approved):
LGTM 👍

minghinmatthewlam (approved):
LGTM

PR #64, opened 2023-10-19
Separate blockchain ID initialization from contract constructor

### LINKED ISSUE #58: Do not call Warp precompile in TeleporterMessenger constructor
**Context and scope**
`TeleporterMessenger` must be deployed using [Nick's method](https://yamenmerhi.medium.com/nicks-method-ethereum-keyless-execution-168a6659479c), which requires that the deployment transaction be sent from an account with a zero nonce. If the deployment transaction reverts, then the account nonce will increment and `TeleporterMessenger` cannot be deployed at the intended address on that chain.

`TeleporterMessenger` calls the Warp precompile in its constructor to fetch the blockchain ID. If the Warp precompile has not been enabled on that chain, then the `TeleporterMessenger` constructor will revert, and the deployment will fail, precluding `TeleporterMessenger` from being deployed at the intended address on that chain. If that happens, then all chains would have to redeploy `TeleporterMessenger` at a new address in order to interact with the chain on which deployment failed.

We should avoid calling the Warp precompile in the `TeleporterMessenger` constructor, and instead fetch the blockchain ID in a separate step that cannot impact deployment.

michaelkaplan13:
## Why this should be merged
Fixes #58. 

> TeleporterMessenger calls the Warp precompile in its constructor to fetch the blockchain ID. If the Warp precompile has not been enabled on that chain, then the TeleporterMessenger constructor will revert, and the deployment will fail, precluding TeleporterMessenger from being deployed at the intended address on that chain.  

## How this works
Instead of calling the warp precompile in the contract constructor, the call is moved to an `initializeBlockchainID` method that must be called prior to receiving messages. Anyone can called `initializeBlockchainID`.

## How this was tested
Unit tests, E2E tests, integration tests

## How is this documented

README for `TeleporterMessenger` contract.


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -285,6 +280,11 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             revert InvalidOriginSenderAddress();
         }
 
+        // If the blockchain ID has yet to be initialized, try to do so now.
+        if (blockchainID == bytes32(0)) {
```
cam-schultz, line 284: Rather than explicitly checking at the callsite, we could modify `initializeBlockchainID` to be a no-op if the `blockchainID` has already been set. I imagine that would save some gas in `receiveCrossChainMessage`, at the cost of making it less obvious that a standalone call to `initializeBlockchainID` had no effect by removing the revert.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -596,6 +596,16 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return _checkIsAllowedRelayer(delivererAddress, allowedRelayers);
     }
 
+    /**
+     * @dev Sets the value of `blockchainID` to the value determined by the warp messenger precompile.
+     */
+    function initializeBlockchainID() public {
```
minghinmatthewlam, line 602: can we name this `initialize()`? Follows what is more commonly used, and `initializeBlockchainID` sounds to me we're just initializing for one field, but doesn't imply intiailizing for the contract

michaelkaplan13: I think `initializeBlockchainID` is more descriptive in this case in this case we are just initializing one field and nothing else about the contract. For instance, you can send messages prior to the blockchain ID being initialized, and message IDs and everything else will be set/incremented.

docker/run_setup.sh
```diff
@@ -110,22 +110,34 @@ if [ ! -e $dir_prefix/NETWORK_RUNNING ]; then
...
+        echo "Error deploying Teleporter Messenger on C-chain."
+        exit 1
+    fi
+    echo "Deployed TeleporterMessenger to C-chain."
+
+    # Initialize the blockchain ID of the Teleporter contract instances.
+    # TODO: Initialize the C-chain Teleporter contract instance once the warp precompile is added to the C-chain.
+    cast send --private-key $user_private_key $teleporter_contract_address "initializeBlockchainID()" --rpc-url $subnet_a_url
+    cast send --private-key $user_private_key $teleporter_contract_address "initializeBlockchainID()" --rpc-url $subnet_b_url
+    cast send --private-key $user_private_key $teleporter_contract_address "initializeBlockchainID()" --rpc-url $subnet_c_url
```
cam-schultz, line 140: With the `receiveCrossChainMessage` lazy initialization, do we want to remove these?

tests/utils/network_setup.go
```diff
@@ -263,8 +242,16 @@ func DeployTeleporterContracts(transactionBytes []byte, deployerAddress common.A
...
-		log.Info("Finished deploying Teleporter contracts")
+		log.Info("Finished deploying Teleporter contract", "blockchainID", subnetInfo.BlockchainID.Hex())
+
+		// Initialize the blockchain ID
+		{
+			initializeBlockchainIDTx := createInitializeBlockchainIDTransaction(ctx, subnetInfo, teleporterContractAddress, fundedAddress, fundedKey)
+			SendTransactionAndWaitForAcceptance(ctx, subnetInfo.ChainWSClient, initializeBlockchainIDTx)
+		}
+		log.Info("Finished initializing blockchain ID of Teleporter instance", "blockchainID", subnetInfo.BlockchainID.Hex())
 	}
```
cam-schultz, line 253: Same question here w.r.t. lazy initialization

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -596,6 +596,16 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        if (blockchainID != bytes32(0)) {
+            revert AlreadyInitialized();
+        }
+        blockchainID = WARP_MESSENGER.getBlockchainID();
```
cam-schultz, line 606: Since this is no longer set on deployment, we might want to emit an event on successful initialization.

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: not the biggest fan of having a no-op call for `initializeBlockchainID`. We could consider have an extra state `isInitialized` that gets set, and check here if it's initialized, and call `initializeBLockchainID` if not. But this approach would cost an extra state variable to store.

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: hmm true with the `receiveCrossChainMessage` lazy initialization, we don't really even need a separate transaction to `initializeBlockChainID`.

minghinmatthewlam, line 602: With the initialization check in `receiveCrossChainMessage` do we even need `initializeBlockchainID` to be made public? The one use case is if a chain wants to intiialize it before first receive.

minghinmatthewlam: Then again, we might not to attribute the additional gas costs of `WARP_MESSENGER.getBlockchainID()` to the first relayer.

michaelkaplan13: That's an interesting point. We could even remove it all and just call `WARP_MESSENGER.getBlockchainID()` if necessary in `receiveCrossChainMessage`, effectively only having lazy initialization. I'm fine with this approach because it minimizes the code changes.

Curious what you and @cam-schultz  think

michaelkaplan13: The `getBlockChainID` gas cost is very very very small also: https://github.com/ava-labs/subnet-evm/blob/master/x/warp/contract.go#L25C2-L25C24

cam-schultz: Nobody's going to care about 2 gas, but somebody is certainly going to forget to get confused about `initializeBlockchainID`. I'm in favor of hiding it behind the interface.

minghinmatthewlam: yup I was only worried about gas incurred by first relayer. Since it's so low, I'm good with just initialization in `receiveCrossChainMessage`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -76,10 +76,12 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     uint256 public constant REQUIRED_ORIGIN_CHAIN_ID_START_INDEX = 4;
     uint256 public constant MINIMUM_REQUIRED_CALL_DATA_LENGTH = 68;
 
-    // The blockchain ID of the chain the contract is deployed on. Determined by warp messenger precompile.
-    bytes32 public immutable blockchainID;
+    // The blockchain ID of the chain the contract is deployed on.
+    // Set at most once by calling initializeBlockchainID() or receiveCrossChainMessage().
+    bytes32 public blockchainID = bytes32(0);
```
minghinmatthewlam, line 81: do you need to explicity declare that it's `= bytes32(0)`, won't it be set for the default value?

If needed explicity, how come you decide to set in declaration here vs in constructor? Even if it were to be set in constructor and transaction runs out of gas, the nonce I don't think gets used up and can be deployed again. 
Was looking at https://ethereum.stackexchange.com/questions/114287/initialize-state-variables-in-the-contract-constructor-or-directly-in-the-declar, and seems result would be generally same, except more clear when used with derived contracts.

cam-schultz (approved):
LGTM

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -285,6 +280,11 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             revert InvalidOriginSenderAddress();
         }
 
+        // If the blockchain ID has yet to be initialized, do so now.
+        if (blockchainID == bytes32(0)) {
+            blockchainID = WARP_MESSENGER.getBlockchainID();
```
bernard-avalabs, line 285: It seems the original design of calling initializeBlockchainID has been replaced with call getBlockchainID. Does calling getBlockchainID require that the warp precompile be enabled (the original motivation for this PR)? If not, is there any reason for this to be called lazily instead of in the constructor?

cam-schultz: `getBlockchainID` does require the warp precompile be enabled, otherwise the transaction will revert. The only issue with calling that method in the constructor is that if the transaction to deploy the `TeleporterMessenger` contract reverts, then the deployer account nonce will still be incremented, invalidating the deployment tx output by Nick's method.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Yeah you're correct that it will default to the zero value without being explicitly set, so I removed that now.

One clarification though, if a transaction runs out of gas it still is included in a block and will use the nonce of the account that sent it. Transactions that don't have enough of the native token to cover the maximum gas it could possibly use (ie `gasFeeCap * gasLimit`), aren't valid to be included in blocks, but whether or not the `gasLimit` is too low for the code that it calls is not known until the transaction itself is executed.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Updated to go with the lazy initialization pattern, which removed `initializeBlockchainID` all together. 👍

geoff-vball (approved):
Just saw one typo, otherwise LGTM 👍

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: ah yup my previous statement about nonce not being used up for gas running out was wrong

tests/utils/utils.go
```diff
@@ -332,3 +332,26 @@ func calculateTxParams(ctx context.Context, wsClient ethclient.Client, fundedAdd
...
+	recipient common.Address,
+	amount *big.Int,
+) *types.Transaction {
+	gasFeeCap, gasTipCap, nonce := calculateTxParams(ctx, network.ChainWSClient, fundedAddress)
+
+	tx := types.NewTx(&types.DynamicFeeTx{
+		ChainID:   network.ChainIDInt,
+		Nonce:     nonce,
+		To:        &recipient,
+		Gas:       DefaultTeleporterTransactionGas,
```
minghinmatthewlam, line 350: do we want this utils function to always pass in gas for the Teleporter transaction?

michaelkaplan13: Updated to be 21,000 since it's always the gas used by a native transfer

geoff-vball (approved):

minghinmatthewlam (approved):

PR #70, opened 2023-10-26
Slither action and avalanche-cli version

minghinmatthewlam:
## Why this should be merged
Fixes the slither action CI that is failing. New `crytic-compile` version is not handling the `target` set for `slither-action` correctly, and the slither action repo seems to be less managed/updated, so moving to slither command instead.

Also updates `avalanche-cli` version

## How this works
Install `slither-analyzer` which has the slither command through pip, and install foundry through `foundry-tools`. Will run slither on all src files and upload a sarif file that gets uploaded to produce alerts in the PRs.

## How this was tested
CI

## How is this documented


geoff-vball (approved):

contracts/foundry.toml
```diff
@@ -2,5 +2,7 @@
 src = 'src'
 out = 'out'
 libs = ['lib']
+solc_version = '0.8.18'
+test = 'tests'
```
cam-schultz, line 6: Are these two fields necessary for slither to work?

.github/workflows/slither.yml
```diff
@@ -22,16 +22,27 @@ jobs:
...
+      - name: Run Slither
+        run: |
+          cd contracts
+          slither ./ --fail-none --sarif ./results.sarif
 
       - name: Upload SARIF file
         uses: github/codeql-action/upload-sarif@v2
         with:
-          sarif_file: ${{ steps.slither.outputs.sarif }}
+          sarif_file: contracts/results.sarif
```
cam-schultz, line 48: I'm assuming that the above step creates a file `contracts/results.sarif`. We should add this to the `.gitignore` if it's not there already.

cam-schultz (approved):
LGTM. Just two small comments

contracts/foundry.toml
minghinmatthewlam: the first one was recommended to be set, but should be fine without. The second one is needed so slither doesn't run on test files and bombard us with alerts in the PRs.

.github/workflows/slither.yml
minghinmatthewlam: the `results.sarif` file should be within the github action runner

.github/workflows/slither.yml
```diff
@@ -22,16 +22,27 @@ jobs:
...
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
         with:
-          target: "contracts/"
-          sarif: results.sarif
-          slither-config: "contracts/slither.config.json"
+          version: nightly
+          cache: false
+
+      - name: Run Slither
```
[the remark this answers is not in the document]
minghinmatthewlam: slither uses crytic-compile, which uses foundry/forge for compiling.

.github/workflows/slither.yml
```diff
@@ -22,16 +22,27 @@ jobs:
...
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install Slither
+        run: |
+          pip install slither-analyzer
+
+      - name: Install Foundry
```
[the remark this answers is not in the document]
minghinmatthewlam: yup @geoff-vball also has a [PR](https://github.com/ava-labs/teleporter/pull/80) pinning the foundryup commit, so want to sync on this

.github/workflows/slither.yml
geoff-vball: Technically we should be able to run slither locally if we wanted to. In which case we would want it ignored.

.github/workflows/slither.yml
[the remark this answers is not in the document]
geoff-vball: Yeah I think this is a much simpler way of install foundry. It's supported by the foundry team. The only place we can't use this is inside the docker container for the integ tests, but that should be going away.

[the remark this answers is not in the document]
minghinmatthewlam: yup think it's simpler to install, changed the other workflows to match as well.

geoff-vball (approved):
👍

michaelkaplan13 (approved):
LGTM. Thanks for working through this dependency issue.

PR #84, opened 2023-10-27
Exit on error in dockerfile

geoff-vball:
## Why this should be merged
This will cause integration tests to exit on the first error in the dockerfile, making it easier to debug issues.

## How this works
Set -e in dockerfile
`test.sh` will check to see if the PID for the docker process still exists while waiting for container to start up.

## How is this documented


geoff-vball:
Closed in favor of combining with https://github.com/ava-labs/teleporter/pull/89

PR #85, opened 2023-10-27
Fix Integration Test and Pin Foundry

michaelkaplan13:
## Why this should be merged

Updates the integration tests to work with the latest breaking Foundry changes, and manually pins the foundry installation to a specific commit.

## How this works

Foundry is installed by first fetching the script at https://raw.githubusercontent.com/foundry-rs/foundry/70d00222c2ef74fc484b8d98c5705d131ab31871/foundryup/install, which previously was not pinned to a commit. That script itself fetches another file from the main branch of their repo, which is worked around by using `sed` to again pin that URL to the same commit. Finally, `foundryup -C` is used to install from a specific commit. 

## How this was tested

CI

## How is this documented

N/A


geoff-vball (approved):

gwen917 (approved):

docker/Dockerfile
```diff
@@ -34,13 +34,15 @@ RUN apt-get update && apt-get install -y locales && rm -rf /var/lib/apt/lists/*
...
 SHELL ["/bin/bash", "-c"]
-RUN curl -L https://foundry.paradigm.xyz | bash && \
+RUN curl -L https://raw.githubusercontent.com/foundry-rs/foundry/70d00222c2ef74fc484b8d98c5705d131ab31871/foundryup/install > /tmp/foundry-install-script && \
+    sed -i 's/\/foundry-rs\/foundry\/master\/foundryup/\/foundry-rs\/foundry\/70d00222c2ef74fc484b8d98c5705d131ab31871\/foundryup/g' /tmp/foundry-install-script && \
+    cat /tmp/foundry-install-script | bash && \
     echo "export PATH=\"$PATH:/$HOME/.foundry/bin\"">> ~/.bashrc && \
     source ~/.bashrc && \
     export PATH=$PATH:$HOME/.foundry/bin:$HOME/.cargo/bin && \
-    foundryup
+    foundryup --version nightly-70d00222c2ef74fc484b8d98c5705d131ab31871
```
cam-schultz, line 45: I believe the foundry nightly builds are regularly pruned, but some stick around longer tem. Do we know if this build is one of the long term ones?

cam-schultz (approved):
One question about the specific nightly build.

Also, we should consider maintaining our own foundry build images from stable commits. We could persist them in dockerhub so that foundry can't surprise us with a broken build or a purged legacy release

docker/Dockerfile
geoff-vball: Good point. We might want to use the oct 2 release which is `5be158b`

michaelkaplan13: That's a great call out. If foundry isn't going have official release versions, I'd be a fan of keeping our own "stable" build hosted some where to ensure availability. 

Created an issue to track it here: https://github.com/ava-labs/teleporter/issues/88

The release used in this branch was just published today, so I think it will be okay to use for at least the short term.

geoff-vball: This should be fine if we merge in https://github.com/ava-labs/teleporter/pull/84, which will make it easy to spot the reason for failure if this version gets pruned.

geoff-vball: Just note that currently the oldest nightly build isn't a monthly release and hasn't been pruned yet is 4 days old.

geoff-vball: Jokes on us, this version has already been pruned 🙃

PR #86, opened 2023-10-27
Rename retryReceipts to sendReceipts

minghinmatthewlam:
## Why this should be merged
Originally the contract had function `retryReceipt` that then changed to `retryReceipts`, but this function is sending a new noop Teleporter message, so `sendReceipts` more appropriately describes the functionality.

## How is this documented
updated existing documentation for retryReceipts


michaelkaplan13:
👍 definitely agree on the concept, now I'm just thinking if `sendReceipts` make it seem as though you need to call it for receipts to be sent...

What do you think of `sendCustomReceipts`? Open to other ideas as well

minghinmatthewlam:
> 👍 definitely agree on the concept, now I'm just thinking if `sendReceipts` make it seem as though you need to call it for receipts to be sent...
> 
> What do you think of `sendCustomReceipts`? Open to other ideas as well

good point, definitely should stay away from just `sendReceipts` then. I'm good with `sendCustomReceipts`, will on it a bit, but if nothing else pops up will change accordingly

cam-schultz (dismissed):

geoff-vball (dismissed):

bernard-avalabs (commented):
Maybe sendSpecifiedReceipts instead of sendCustomReceipts? Or perhaps sendSelectedReceipts?

It is hard to guess what sendCustomReceipts is used for without reading the comments.

michaelkaplan13:
> Maybe sendSpecifiedReceipts instead of sendCustomReceipts? Or perhaps sendSelectedReceipts?
> 
> It is hard to guess what sendCustomReceipts is used for without reading the comments.

I agree `sendSpecifiedReceipts` is better than `sendCustomReceipts`. Thanks for the suggestion

michaelkaplan13 (approved):
LGTM!

cam-schultz (approved):

PR #87, opened 2023-10-27
Cleanup docs

cam-schultz:
## Why this should be merged
Brings the Teleporter READMEs up-to-date with the latest implementation.
Also moves a handful of scripts from `scripts/local/` to `scripts/`


geoff-vball (dismissed):

README.md
```diff
@@ -57,28 +57,36 @@ docker run hello-world # This should work without sudo now
...
-- `contract-deployment/` includes Golang scripts to derive the expected EVM contract address deployed from a given EOA at a specific nonce, and also construct a transaction to deploy provided byte code to the same address on any EVM chain using [Nick's method](https://yamenmerhi.medium.com/nicks-method-ethereum-keyless-execution-168a6659479c#).
-- `contracts/` is a [foundry](https://github.com/foundry-rs/foundry) project that includes the implementation of the `TeleporterMessenger` contract and other dApps that serve as examples for how to write contracts that interact with Teleporter.
-- `scripts/` includes scripts to run Teleporter and integration tests.
-  - Note that all the scripts should be run from the root of the repository.
-  - `scripts/local/` includes scripts to run Teleporter in Docker containers and tests in the `scripts/local/integration-tests` locally.
+- `scripts/` includes bash scripts for interacting with Teleporter in various environments, as well as utility scripts. All the scripts should be run from the root of the repository.
```
minghinmatthewlam, line 72: I think are scripts should mostly be able to handle not being run from the root

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -110,10 +104,10 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     /**
      * @dev See {ITeleporterMessenger-retrySendCrossChainMessage}
      *
-     * Emits {SendCrossChainMessage} event.
+     * Emits a {SendCrossChainMessage} event.
      * Requirements:
      *
-     * - `message` must have been previously sent to the given destination chain ID.
+     * - `message` must have been previously sent to the given `destinationChainID`.
```
geoff-vball, line 110: I think we should rename this to `destinationBlockchainID`, as per our other discussions of confusing this with the ethereum `chainID`.

contracts/src/CrossChainApplications/ERC20Bridge/README.md
```diff
@@ -13,7 +13,7 @@ The generic ERC20 bridge is implemented using two primary contracts.
     - Bridge contract that uses Teleporter to facilitate the bridging operations of adding new supported tokens and moving tokens between chains.
     - The bridge contract tracks the balances of each token sent to other bridge instances, and only allows those bridge instances to redeem up to the amount of tokens that has been previously sent to them.
     - Primary functions include:
-        - `submitCreateBridgeToken`: Called on the origin subnet to add support for an ERC20 native to the chain to a bridge instance on a different chain. Submits a Teleporter message that invokes `createBridgeToken` on the destination chain.
+        - `submitCreateBridgeToken`: Called on the origin chain to add support for an ERC20 native to the chain to a bridge instance on a different chain. Submits a Teleporter message that invokes `createBridgeToken` on the destination chain.
```
minghinmatthewlam, line 16:
```suggestion
        - `submitCreateBridgeToken`: Called on the origin chain to add support for a native ERC20 token on a different chain's bridge instance. Submits a Teleporter message that invokes `createBridgeToken` on the destination chain.
```

contracts/src/CrossChainApplications/README.md
```diff
@@ -39,17 +34,17 @@ pragma solidity 0.8.18;
...
+contract MyExampleCrossChainMessenger {}
 ```
 
 ### Step 2: Integrating Teleporter Messenger
 
-Now that we have our initial empty `ExampleCrossChainMessenger`, it's time to add our `ITeleporterMessenger` that will provide the functionality to deliver cross chain messages.
+Now that the initial empty `MyExampleCrossChainMessenger` is defined, it's time to integrate the `ITeleporterMessenger` that will provide the functionality to deliver cross chain messages.
 
-Create a state variable of `ITeleporterMessenger` type called `teleporterMessenger`. Then we'll create a constructor for our contract that takes in an address where the Teleporter messenger would be deployed on this chain, and set our state variable with it.
+Create a state variable of `ITeleporterMessenger` type called `teleporterMessenger`. Then create a constructor for our contract that takes in an address where the Teleporter messenger would be deployed on this chain, and set our state variable with it.
```
minghinmatthewlam, line 47: above we have diferent capitalizatoin with "Teleporter Messenger protocol", how do we want to standardize this?

contracts/src/CrossChainApplications/README.md
```diff
@@ -90,11 +85,7 @@ function receiveTeleporterMessage(
 ) external {
 ```
 
-Now that the function declarations are set up, we can move on to implementation.
-
-> To incentivize relayers to deliver our cross chain messages from source to destination, we will need to support `ERC20` fees. Relayer fees aren't needed when running your own relayers to deliver your own messages and pay costs out of pocket.
```
minghinmatthewlam, line 95: I liked a note to explain the use of fees before the step of importing the IERC20 contract, and calling out that fees aren't needed if relaying own message.

contracts/src/CrossChainApplications/README.md
```diff
@@ -136,7 +125,7 @@ return
     );
 ```
 
-With the sending complete, we move on to implementing the receiver. The receiver in our example will just receive the arbitrary string data, and check that the message is sent through Teleporter.
+With the sending side complete, the next step is to implement `receiveTeleporterMessage`. The receiver in this example will just receive the arbitrary string data, and check that the message is sent through Teleporter.
```
minghinmatthewlam, line 128:
```suggestion
With the sending side complete, the next step is to implement `ITeleporterReceiver.receiveTeleporterMessage`. The receiver in this example will just receive the arbitrary string data, and check that the message is sent through Teleporter.
```

contracts/src/CrossChainApplications/README.md
```diff
@@ -205,12 +194,12 @@ There we have it, a simple cross chain messenger built on top of Teleporter! Ful
...
 ### Step 5: Testing
 
-For testing we have a `test.sh` script that sets up a local avalanche network with three subnets deployed with Teleporter, and a relayer to deliver Teleporter messages. To add an integration test simply add a new test script under `integration-tests`. In this case we already have our `example_messenger.sh`, which performs the following steps:
+For testing, `scripts/local/test.sh` sets up a local Avalanche network with three subnets deployed with Teleporter, and a relayer to deliver Teleporter messages. To add an integration test simply add a new test script under `integration-tests`. An integration test for `ExampleCrossChainMessenger` is already included (`scripts/local/integration_tests/example_messenger.sh`), which performs the following steps:
 
-1. Deploys our [ExampleERC20](../Mocks/ExampleERC20.sol) token to subnet A.
-2. Deploy the above `ExampleCrossChainMessenger` to both subnets A and B.
-3. Approve the cross chain messenger on subnet A to spend erc20 from the default address.
-4. Send `"hello world"` from subnet A to subnet B's cross chain messenger to receive.
+1. Deploy the [ExampleERC20](../Mocks/ExampleERC20.sol) token to subnet A.
```
minghinmatthewlam, line 199:
```suggestion
1. Deploys the [ExampleERC20](../Mocks/ExampleERC20.sol) token to subnet A.
```

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -35,11 +35,11 @@ struct TeleporterFeeInfo {
 }
 
 /**
- * @dev Interface that describes functionalities for a cross chain messenger.
+ * @dev Interface that describes functionalities for a cross-chain messenger implementing the Teleporter protcol.
```
minghinmatthewlam, line 38: we have some references to "Teleporter Messenger protocol", should unify these

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -59,7 +59,7 @@ interface ITeleporterMessenger {
     );
 
     /**
-     * @dev Emitted when Teleporter message is being delivered on destination chain and address,
+     * @dev Emitted when a Teleporter message is being delivered on the destination chain to an address,
      * but message execution fails. Failed messages can then be retried.
```
minghinmatthewlam, line 63:
```suggestion
     * but message execution fails. Failed messages can then be retried with `retryMessageExecution`.
```

contracts/src/Teleporter/README.md
```diff
@@ -34,13 +34,25 @@ Teleporter provides a handful of useful properties to cross-chain applications t
...
+
+ ### Message Receipts and Fee Redemption
+In order to confirm delivery of a Teleporter message from one chain to another, a receipt is included in the next Teleporter message sent in the opposite direction. This receipt contains the message ID of the original message, as well as the address of the relayer that delivered it. That address is then able to redeem the reward on the original chain by calling `redeemRelayerRewards`. The following example illustrates this flow:
```
minghinmatthewlam, line 40:
```suggestion
In order to confirm delivery of a Teleporter message from source chain to destination chain, a receipt is included in the next Teleporter message sent from destination chain to source chain. This receipt contains the message ID of the original message, as well as the address of the relayer that delivered it. That address is then able to redeem the reward on the original chain by calling `redeemRelayerRewards`. The following example illustrates this flow:
```

contracts/src/Teleporter/README.md
```diff
@@ -34,13 +34,25 @@ Teleporter provides a handful of useful properties to cross-chain applications t
...
+- A Teleporter message is sent from Chain A to Chain B, with a relayer incentive of `10` `USDC`. This message is assigned the ID `1` by the Teleporter contract on Chain A.
+  - On Chain A, this is done by calling `sendCrossChainMessage`, and providing the `USDC` contract address and amount in the function call.
+- A relayer delivers the message on Chain B by calling `receiveCrossChainMessage` and providing its address, `0x123...`
```
minghinmatthewlam, line 43: not sure if we want capitalization for "Chain" specifically here

contracts/src/Teleporter/README.md
```diff
@@ -34,13 +34,25 @@ Teleporter provides a handful of useful properties to cross-chain applications t
...
+- The Teleporter contract on Chain B stores the relayer address in a receipt for the message ID.
+- Some time later, a separate Teleporter message is sent from Chain B to Chain A. The Teleporter contract on Chain B includes the receipt for the original message in this new message.
+- When this new message is delivered on Chain A, the Teleporter contract on Chain A reads the receipt and marks the original message (message ID `1`) as delivered by address `0x123...`.
+- Address `0x123...` may now call `redeemRelayerRewards` on Chain A, which transfers the `10` `USDC` to its address. If it tries to do this before the receipt is received on Chain A, the call will fail.
+
+It is possible for receipts to get "stuck" on the destination chain in the event that Teleporter traffic between two chains is skewed in one direction. In such a scenario, incoming messages on one chain may cause the rate at which receipts are generated to outpace the rate at which they are sent back to the other chain. To mitigate this, the method `retryReceipts` can be called to immediately send the receipts associated with the given message IDs back to the original chain.
```
minghinmatthewlam, line 49:
```suggestion
It is possible for receipts to get "stuck" on the destination chain in the event that Teleporter traffic between two chains is skewed in one direction. In such a scenario, incoming messages on one chain may cause the rate at which receipts are generated to outpace the rate at which they are sent back to the other chain. To mitigate this, the method `sendSpecifiedReceipts` can be called to immediately send the receipts associated with the given message IDs back to the original chain.
```
can we preemptively change this

contracts/src/Teleporter/README.md
```diff
@@ -34,13 +34,25 @@ Teleporter provides a handful of useful properties to cross-chain applications t
...
+Teleporter messages are delivered by calling the `receiveTeleporterMessage` function defined by the `ITeleporterReceiver` interface. Contracts must implement this interface in order to be able to receive messages. The first two paramaters of `receiveTeleporterMessage` identify the original sender of the given message on the origin chain and are set by the `TeleporterMessenger`. The third parameter to `receiveTeleporterMessage`, is the raw message payload. Applications using Teleporter are responsible for defining the exact format of this payload in a way that can be decoded on the receiving end. For example, applications may encode an action enum value along with the target method parameters on the sending side, then decode this data and route to the target method within `receiveTeleporterMessage`. See `ERC20Bridge.sol` for an example of this approach.
```
minghinmatthewlam, line 52:
```suggestion
Teleporter messages are delivered by calling the `receiveTeleporterMessage` function defined by the `ITeleporterReceiver` interface. Contracts must implement this interface in order to be able to receive messages. The first two parameters of `receiveTeleporterMessage` identify the original sender of the given message on the origin chain and are set by the `TeleporterMessenger`. The third parameter to `receiveTeleporterMessage`, is the raw message payload. Applications using Teleporter are responsible for defining the exact format of this payload in a way that can be decoded on the receiving end. For example, applications may encode an action enum value along with the target method parameters on the sending side, then decode this data and route to the target method within `receiveTeleporterMessage`. See `ERC20Bridge.sol` for an example of this approach.
```

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: will echo this as DevRel brought it up too. They also suggsted possible `avalancheChainID`

cam-schultz: I definitely agree that we should rename these variables to `destinationBlockchainID`. Slight preference for that over `destinationAvalancheChainID` for verbosity.

cam-schultz: However, I think this is best suited for a standalone set of PRs, since it will require coordination on the awm-relayer side. Ticket: https://github.com/ava-labs/teleporter/issues/95

contracts/src/CrossChainApplications/ERC20Bridge/README.md
cam-schultz: "Native" hear can be a bit confusing, since it's possible for the ERC20 on a source chain to not be "native" to that chain if it was bridged to that chain by a bridging protocol. For example, USDC.e on Avalanche is not "native" to the C-Chain, even though it's very likely that we'd bridge USDC.e from the C-Chain to a subnet via a bridge like this.

cam-schultz: Will remove native altogether, and reword the other part.

contracts/src/CrossChainApplications/README.md
cam-schultz: My opinion is that we should always capitalize Teleporter and Teleporter Messenger, since they refer to distinct entities (the Teleporter protocol and the TeleporterMessenger contract)

contracts/src/Teleporter/ITeleporterMessenger.sol
cam-schultz: Here I think just calling it Teleporter is more appropriate. "Teleporter Messenger" is an implementation of the Teleporter protocol. This interface is specifically tied to "Teleporter Messenger", but I think we should consider paring it back to conform only the Teleporter protocol. All the additional methods included in `TeleporterMessenger.sol` would be outside of the protocol.

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: sgtm thanks for making the issues

contracts/src/Teleporter/ITeleporterMessenger.sol
minghinmatthewlam: do you think it makes sense that when we refer to "protocol" it refers to Teleporter protocol only? Like you said Teleporter Messenger is an implementation of the protocol, so in the READMEs we can remove all references to Teleproter Messenger as a protocol.

cam-schultz: Good call. Replaced "Teleporter Messenger protocol" with "Teleporter protocol" in a few places. Lmk if you spot any others.

contracts/docs/.gitignore
```diff
@@ -0,0 +1 @@
+book/
```
minghinmatthewlam, line 1: what is in the generated book, and why do we want to ignore it?

minghinmatthewlam (commented):
LGTM, thanks for cleaning this up. Left a question on why we ignore `book/`, and also is there a way to have nested `contracts/docs/src/src` and move the child src into parent directory?

README.md
```diff
@@ -121,13 +129,29 @@ source vars.sh
...
+    -h, --help                        Print this help message
+  ```
+  - Note that if `-l, --local` is not set, then the latest published `awm-relayer` image will be pulled from Dockerhub.
+- This script performs the same setup steps as `scripts/local/run.sh` (described above), and then runs the tests in `./scripts/integration-tests/` automatically
+
+### Additional notes
+
+- Both the `./scripts/local/run.sh` and `./scripts/local/test.sh` scripts run local five node networks, with each of the nodes validating the primary network and three subnets (Subnet A, Subnet B, and Subnet C).
+- Logs from the subnets on one of the five nodes are printed to stdout when run using either script.
+- These logs can also be found at `~/.avalanche-cli/runs/network-runner-root-data_<DATE>_<TIMESTAMP>/node{1,5]/logs/<SUBNET_ID>.log`, or at `/var/lib/docker/overlay2/<CONTAINER_ID>/merged/root/.avalanche-cli/....` on your local machine. You will need to be the root user to access the logs under `/var/lib`.
```
feuGeneA, line 150:
```suggestion
- These logs can also be found at `~/.avalanche-cli/runs/network-runner-root-data_<DATE>_<TIMESTAMP>/node{1,5]/logs/<SUBNET_ID>.log` in the `local_network_run` container, or at `/var/lib/docker/overlay2/<CONTAINER_ID>/merged/root/.avalanche-cli/....` on your local machine. You will need to be the root user to access the logs under `/var/lib`.
```

README.md
```diff
@@ -160,7 +184,7 @@ source .env.testnet
 source .env
 ```
 
-### E2E tests
+## E2E tests
 
 E2E tests are ran as part of CI, but can also be ran locally with the `--local` flag. To run the E2E tests locally, you'll need to install Gingko following the intructions [here](https://onsi.github.io/ginkgo/#installing-ginkgo)
```
feuGeneA, line 189:
```suggestion
E2E tests are run as part of CI, but can also be run locally with the `--local` flag. To run the E2E tests locally, you'll need to install Gingko following the instructions [here](https://onsi.github.io/ginkgo/#installing-ginkgo)
```

feuGeneA (changes requested):
A partial review here. I still have some more files that I want to review, but I wanted to get these comments over to you ASAP.

I think you need to `git rm -rf contracts/docs`. I saw where you added `contracts/docs` to `.gitignore`, but you didn't actually remove those files from the repository.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -158,12 +152,12 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     /**
      * @dev See {ITeleporterMessenger-addFeeAmount}
      *
-     * Emits {AddFeeAmount} event.
+     * Emits a {AddFeeAmount} event.
```
feuGeneA, line 155:
```suggestion
     * Emits an {AddFeeAmount} event.
```

feuGeneA (changes requested):
Finished reviewing all the files :)

contracts/docs/.gitignore
cam-schultz: This gitignore is generated by the forge docs command. I had initially thought to check in the generated docs, but they include the git commit in the markdown files themselves, meaning that our approach of building the docs in CI and checking for git diffs would not work, since every commit to the repo would cause  a diff. Instead, I elected to add README instructions for generating the docs.

README.md
```diff
@@ -57,28 +57,36 @@ docker run hello-world # This should work without sudo now
...
 
-## Build + Run + Test
+## Run the Docker integration tests
 
 - Get all submodules: `git submodule update --init --recursive`
 - Install Docker as described in the setup section of the README in the root of this repository.
-- If we are using a local version of the `awm-relayer` image, build it using `./scripts/build_local_image.sh` from the root of `awm-relayer` repository.
+- If we are using a local version of the `awm-relayer` image, build it using `./scripts/build_local_image.sh` from the root of the `awm-relayer` repository.
+
+### Start up the local testnet
```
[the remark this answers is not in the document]
cam-schultz: I went ahead and implemented this behavior change 🙂 . I don't think we need to explicitly mention it here now.

README.md
```diff
@@ -93,7 +101,7 @@ cast send --private-key 0x56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5c
 cast balance --rpc-url http://127.0.0.1:9652/ext/bc/C/rpc 0x333d17d3b42bf7930dbc6e852ca7bcf560a69003
 ```
 
-- After calling `./scripts/local/run.sh`, you can directly send messages between the deployed subnets (in comparison, calling `./scripts/local/test.sh` runs the same setup steps, then runs the tests in `./scripts/integration-tests/` automatically). As an example, `./scripts/integration-tests/basic_send_receive.sh` can be run manually like so:
+- After calling `./scripts/local/run.sh`, you can directly send messages between the deployed subnets . As an example, `./scripts/integration-tests/basic_send_receive.sh` can be run manually like so:
```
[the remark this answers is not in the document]
cam-schultz: Thanks for digging into this. I'd actually prefer to keep the commands as is, that shell in the container persists after running `basic_send_receive.sh`, in case the user wants to run other tests or inspect the state of the chains manually. Providing a command via `-c` won't give the user an interactive shell, if I'm not mistaken.

minghinmatthewlam (dismissed):
LGTM thanks for cleaning up

feuGeneA (changes requested):
Requested one more small change here: https://github.com/ava-labs/teleporter/pull/87#discussion_r1384857450

README.md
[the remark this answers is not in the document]
cam-schultz: Good catch! Corrected now.

geoff-vball (approved):
👍

feuGeneA (approved):

PR #89, opened 2023-10-29
Pin foundry to most recent long-term nightly release

geoff-vball:
## Why this should be merged
Foundry is currently pinned to a nightly release that has already been pruned. This pins to the most recently monthly release that does not get pruned (at least for a while, the oldest such release is from Jan 2023).

Also sets the test to exit if there are any errors in the dockerfile. Right now if foundry fails to install, the test will fail to set everything up properly and just spin until it times out.

Also undoes a quoting change done by @michaelkaplan13 in the latest fix. We might need to change this back once we go with a newer version.

## How is this documented


docker/Dockerfile
```diff
@@ -3,6 +3,8 @@
 
 FROM ubuntu:20.04
 
+RUN set -e
```
feuGeneA, line 6: I hate to be sprinkling this link around everywhere :) but just in case my hunch is right, I need to leave it here too haha: https://github.com/ava-labs/bitcoin-sgx-bridge/pull/263#discussion_r1376351645

geoff-vball: See my comment on your other comment :)

feuGeneA, line 6: I think this isn't having any effect. I believe that this only affects the shell instance invoked by this particular `RUN` command, and doesn't affect any subsequent `RUN` commands.  I believe the repro below shows this, in that if the change in question worked, the repro below would have failed on line 4 instead of line 5.

```
gene@gene-dev:~/dev/tmp$ cat -n Dockerfile
     1  FROM bash:4.4
     2  SHELL ["/usr/local/bin/bash", "-c"]
     3  RUN set -e
     4  RUN ls jjj; true
     5  RUN set -e; ls jjj; true
gene@gene-dev:~/dev/tmp$ docker build .
[+] Building 1.3s (7/7) FINISHED                                                                           docker:default
 => [internal] load .dockerignore                                                                                    0.2s
 => => transferring context: 2B                                                                                      0.0s
 => [internal] load build definition from Dockerfile                                                                 0.1s
 => => transferring dockerfile: 140B                                                                                 0.0s
 => [internal] load metadata for docker.io/library/bash:4.4                                                          0.0s
 => [1/4] FROM docker.io/library/bash:4.4                                                                            0.0s
 => CACHED [2/4] RUN set -e                                                                                          0.0s
 => CACHED [3/4] RUN ls jjj; true                                                                                    0.0s
 => ERROR [4/4] RUN set -e; ls jjj; true                                                                             0.9s
------                                                                                                                    
 > [4/4] RUN set -e; ls jjj; true:
0.362 ls: jjj: No such file or directory
------
Dockerfile:5
--------------------
   3 |     RUN set -e
   4 |     RUN ls jjj; true
   5 | >>> RUN set -e; ls jjj; true
   6 |     
--------------------
ERROR: failed to solve: process "/usr/local/bin/bash -c set -e; ls jjj; true" did not complete successfully: exit code: 1
```

PR #91, opened 2023-10-30
Integrate subnetevm 0.5.8

gwen917:
## Why this should be merged
- Integrate with subnet-evm v0.5.8
   - DestinationChainID and DestinationAddress are removed from WarpMessage. Encode these two fields into TeleporterMessage.
- UpgradeAvalancheGo to v1.10.14

## How this was tested
- Unit tests


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -268,20 +264,23 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-        );
 
-        // Parse the payload of the message.
-        TeleporterMessage memory teleporterMessage = abi.decode(
-            warpMessage.payload,
-            (TeleporterMessage)
+        // Require the source teleporter address matches this contract.
+        // It is previously encoded as destinationAddress in WarpMessage before subnet-evm@v0.5.7
+        require(
+            teleporterMessage.sourceTeleporterAddress == address(this),
```
minghinmatthewlam, line 282: do we need the `sourceTeleoprterAddress`? In the `WarpMessage` origin sender address is included, and we already that it's from the Teleporter address.

contracts/src/Teleporter/tests/TeleporterMessengerTest.t.sol
```diff
@@ -276,6 +276,8 @@ contract TeleporterMessengerTest is Test {
             TeleporterMessage({
                 messageID: messageID,
                 senderAddress: address(this),
+                destinationChainID: MOCK_BLOCK_CHAIN_ID,
```
minghinmatthewlam, line 279: in my branch I should have the `MOCK_BLOCK_CHAIN_ID` constant removed. I think the tests shoudl use `DEFAULT_ORIGIN_CHAIN_ID` and `DEFAULT_DESTINATION_CHAIN_ID`, without the extra `MOCK_BLOCK_CHAIN_ID`

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: I think this is mirroring the check against the now removed `warpMessage.destinationAddress`.

We might not need that check at all though... We already verify that the sender is the same Teleporter contract address, so the check against the intended destination seems redundant. I agree with Matt that we can remove the field `sourceTeleporterAddress` from `TeleporterMessage`

gwen917: removed

.github/workflows/e2e.yml
```diff
@@ -12,7 +12,7 @@ on:
       - "*"
 
 env:
-  GO_VERSION: "1.20.8"
+  GO_VERSION: "1.20.10"
```
michaelkaplan13, line 15: Good call to bump this version 🙏 

Could you also set it in `abi_bindings_checker.yml`?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -268,22 +264,18 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-            "TeleporterMessenger: invalid destination address"
-        );
-
         // Parse the payload of the message.
         TeleporterMessage memory teleporterMessage = abi.decode(
             warpMessage.payload,
             (TeleporterMessage)
         );
 
+        // Require that the message was intended for this blockchain and teleporter contract.
```
michaelkaplan13, line 273: It looks like we've removed a check that the `destinationAddress` is the same as `address(this)`. IIUC, this check is not needed because we already check that the message was sent by the same address as `address(this)`. If that's correct, could you just update this comment to remove the "add teleporter contract" part?

michaelkaplan13 (dismissed):
Mainly LGTM. Just have one clarifying question. 👍

contracts/src/Teleporter/TeleporterMessenger.sol
gwen917: Yes. Previously, we check that `warpMessage.destinationAddress` is same as the teleporter contract address(`address(this)`). The destinationAddress is removed from warpMessage, and we agree that we do not need to check in this [discussion](https://github.com/ava-labs/teleporter/pull/91#discussion_r1376951683). I will update the comments.

.github/workflows/e2e.yml
gwen917: Updated the version

cam-schultz (commented):
One small request and a clarifying question, otherwise LGTM

contracts/src/Teleporter/tests/TeleporterMessengerTest.t.sol
```diff
@@ -276,6 +275,7 @@ contract TeleporterMessengerTest is Test {
             TeleporterMessage({
                 messageID: messageID,
                 senderAddress: address(this),
+                destinationChainID: DEFAULT_ORIGIN_CHAIN_ID,
```
minghinmatthewlam, line 278: this should be default destination chain id and the other unit tests that check for it

contracts/src/Teleporter/tests/HandleInitialMessageExecutionTests.t.sol
```diff
@@ -105,6 +105,7 @@ contract HandleInitialMessageExecutionTest is TeleporterMessengerTest {
         TeleporterMessage memory messageToReceive = TeleporterMessage({
             messageID: 42,
             senderAddress: address(this),
+            destinationChainID: DEFAULT_ORIGIN_CHAIN_ID,
```
minghinmatthewlam, line 108: same for these can they be changed to default destination chain id

contracts/src/Teleporter/tests/GetNextMessageIdTests.t.sol
```diff
@@ -30,8 +30,8 @@ contract GetNextMessageIDTest is TeleporterMessengerTest {
 
         vm.mockCall(
             WARP_PRECOMPILE_ADDRESS,
-            abi.encode(WarpMessenger.sendWarpMessage.selector),
-            new bytes(0)
+            abi.encode(IWarpMessenger.sendWarpMessage.selector),
+            abi.encode(bytes32(0))
```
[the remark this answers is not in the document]
gwen917: Yes, it is necessary. It requires bytes32.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -268,22 +264,18 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             blockchainID = WARP_MESSENGER.getBlockchainID();
         }
 
-        // Require that the message was intended for this blockchain and teleporter contract.
```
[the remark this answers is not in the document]
gwen917: fixed

contracts/src/Teleporter/tests/HandleInitialMessageExecutionTests.t.sol
gwen917: fixed

contracts/src/Teleporter/tests/TeleporterMessengerTest.t.sol
gwen917: fixed

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -13,6 +13,7 @@ import "@openzeppelin/contracts/utils/Address.sol";
  */
 struct ProtocolRegistryEntry {
     uint256 version;
+    address destinationAddress;
```
michaelkaplan13, line 16: It seems a little odd to add `destinationAddress` into the `ProtocolRegistryEntry` struct, since the entry itself is still just a `version->protocolAddress` map. 

It could be a bit easier to understand if had a `UpdateRegistryMessage` struct that had a `ProtocolRegistryEntry` member, as well as the destination address of the message. This way the constructor wouldn't need a "destinationAddress" value either, since it's not used.

cc @minghinmatthewlam in case you have thoughts here.

minghinmatthewlam: yup I was thinking the same so sgtm. Only nit is I prefer more similar naming, we have `addProtocolVersion` function and `AddProtocolVersion` event, so would consider something similar like `AddProtocolVersionInput`.

minghinmatthewlam: actually now that I think about it we don't need the second struct, we can decode two arguments one of them being `destinationAddress` instead 

```
        (ProtocolRegistryEntry memory entry, address destinationAddress) = abi.decode(
            message.payload,
            (ProtocolRegistryEntry, address)
        );
```

minghinmatthewlam (commented):
LGTM, just the one comment about not needing extra struct for upgrade registry, we can directly decode to `(ProtocolRegistryEntry, address)`

contracts/src/WarpProtocolRegistry.sol
gwen917: fixed

cam-schultz (dismissed):
LGTM! Just looks like there are some merge conflicts that need to be resolved before merging.

minghinmatthewlam (dismissed):
LGTM!

geoff-vball:
@gwen917 I think one of the integration tests is legitimately failing with error ` Destination chain on subnet B did not receive message before timeout.`

gwen917:
> @gwen917 I think one of the integration tests is legitimately failing with error ` Destination chain on subnet B did not receive message before timeout.`

yes, this relayer is not updated yet. In CI, the relayer version is still based on subnet-evm 0.5.6. Once this [PR](https://github.com/ava-labs/awm-relayer/pull/76) merged, it should pass. I tested locally.

geoff-vball (approved):

PR #92, opened 2023-10-30
Implement TeleporterOwnerUpgradeable

### LINKED ISSUE #75: Implement TeleporterOwnerUpgradeable
**Context and scope**
`TeleporterUpgradeable` is a utility contract dapps can inherit to help integrate with the `TeleporterRegistry`, but dapps have to implement `TeleporterUpgradeable.updateMinTeleporterVersion`, we want to provide a contract that helps implement the function for `Ownable` contracts.

Create an abstract contract `TeleporterOwnerUpgradeable` that inherits `TeleporterUpgradeable`, and implements the `updateMinTeleporterVersion`. Update the cross chain apps to use this contract and update README and examples.

**Open questions**
Consider other access control patterns

minghinmatthewlam:
## Why this should be merged
Fixes #75 

## How this works
New abstract contract `TeleporterOwnerUpgradeable` that inherits `TeleporterUpgradeable` and implements `updateMinTeleporterVersion` so that only the owner of the contract can call the function. Replaces the previous cross chain application inheritances to use new contract.

## How this was tested
- new unit tests
- CI

## How is this documented
update README


contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
```diff
@@ -637,11 +637,18 @@ contract ERC20BridgeTest is Test {
     }
 
     function testUpdateMinTeleporterVersion() public {
```
michaelkaplan13, line 639: Curious if you think it'd make sense to remove this test case from `ERC20BridgeTests` given that the implementation is tested in `TeleporterOwnerUpgradeableTests`. I don't think we commonly tests inherited functions in child contract test suites.

contracts/src/Teleporter/upgrades/tests/TeleporterOwnerUpgradeableTests.t.sol
```diff
@@ -0,0 +1,101 @@
...
+        );
+        app.updateMinTeleporterVersion();
+
+        // Check that call with old owner reverts
+        vm.prank(address(this));
+        vm.expectRevert("Ownable: caller is not the owner");
+        app.updateMinTeleporterVersion();
+    }
+
+    function testRenounceOwnership() public {
```
michaelkaplan13, line 84: Should we have a test case where `renounceOwnership` is called successfully, and then assert that the previous owner is unable to update the min teleporter version afterwards?

minghinmatthewlam: yup the last few lines check this, but lmk if this should be made more clear or separated.

contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
minghinmatthewlam: hmm ye I'd usually agree, but I at least wanted one cross chain app that inherits `TeleporterOwnerUpgradeable` to have a unit test for it, and the other cross chain apps don't have their own unit tests. Also just thought extra unit tests don't hurt.

cam-schultz: A workaround to this issue would be to implement `TeleporterOwnerUpgradable` in a separate test contract, then move the unit test for `testUpdateMinTeleporterVersion` there.

cam-schultz: Oh, I see such a contract already exists. I agree with Michael that this unit test would make more sense in `TeleporterUpgradeableTests.t.sol`.

I don't think we sacrifice test coverage in doing so, since `ERC20Bridge` doesn't override this method. If it did, we'd definitely want to test it separately.

cam-schultz (commented):
LGTM, however I think we should test `updateMinTeleporterVersion` in `TeleporterOwnerUpgradeableTests` rather than `ERC20BridgeTests`

contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
minghinmatthewlam: that's true, since ERC20Bridge doesn't add additional override/functionality and just inherits, having the same test in `TeleporterOwnerUpgradeableTests` is already covering same test coverage. Removed.

minghinmatthewlam: Thoughts on the same topic but for `Ownable.transferOwnerhip` and `Ownable.renounceOwnership`? I added extra unit tests for those in `TeleporterOwnerUpgradeableTests` though we just inherit the functions in `TeleporterOwnerUpgradeable`. Thinking behind it being these are external contracts, and we can have our own unit tests covering the integration with `updateMinTeleporterVersion`.

contracts/src/Teleporter/upgrades/tests/TeleporterOwnerUpgradeableTests.t.sol
```diff
@@ -0,0 +1,101 @@
...
+        vm.prank(teleporterAddress);
+        vm.expectEmit(true, true, true, true, address(app));
+        emit MinTeleporterVersionUpdated(
+            minTeleporterVersion,
+            minTeleporterVersion + 1
+        );
+        app.updateMinTeleporterVersion();
+
+        // Check that call with old owner reverts
+        vm.prank(address(this));
```
michaelkaplan13, line 79: Super nit: I don't think we need this `prank` here, the call will come from `address(this)` by default.

contracts/src/Teleporter/upgrades/tests/TeleporterOwnerUpgradeableTests.t.sol
michaelkaplan13: Ah my bad, I was just reading it wrong before. 👍 

One minor nit, can we have a `MOCK_OWNER_ADDRESS` value that we use for pranking & transferring ownership to in this test suite rather than using the `teleporterAddress`? I think it would make it slightly more clear and easier to reason about.

contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
michaelkaplan13: 👍 

Yep that makes sense to me. The unit tests you're talking about that call `transferOwnership` and `renounceOwnership` are actually testing the behavior of `updateMinTeleporterVersion`, given its use of those other methods. Along that line, I would recommend keeping them in place as is, but possibly renaming them to "testCannotUpdateMinVersionAfterOwnershipTransfer", etc (or something along those lines).

bernard-avalabs (dismissed):
LGTM

cam-schultz (dismissed):
LGTM

michaelkaplan13 (dismissed):

cam-schultz (dismissed):

geoff-vball (approved):
👍

cam-schultz (approved):

PR #93, opened 2023-10-31
General Ginkgo E2E test framework + Teleporter test cases

cam-schultz:
## Why this should be merged
The E2E tests currently spin up a local network and run the tests in a Ginkgo container environment. This PR adds the ability to run the same E2E tests against any network, such as a set of subnets deployed on Fuji.

This PR also serves as a dev branch for the test cases implemented in the PRs listed in https://github.com/ava-labs/teleporter/issues/94. Once all of those test cases are implemented and merged here, this PR will be eligible to merge to main.

## How this works
Adds a `Network` interface that may be implemented to point to any network. Adds a layer of indirection in the Ginkgo E2E entrypoint to inject the local network implementation. A Fuji implementation may similarly be provided.

## How this was tested
E2E tests, CI. Successfully ran E2E tests against live Fuji subnets.

## How is this documented
Added a README section describing how to implement and use `Network` to generalize the E2E tests.


tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,146 @@
...
+	return userAddress, key
+}
+
+func (g *FujiNetwork) RelayMessage(ctx context.Context,
+	sourceBlockHash common.Hash,
+	sourceBlockNumber *big.Int,
+	source utils.SubnetTestInfo,
+	destination utils.SubnetTestInfo) *types.Receipt {
+
+	// Rely on a separately deployed relayer to relay the message
```
michaelkaplan13, line 143: Should we implement that a check that the message has been successfully relayed within the given ctx timeout here, or is that best to live elsewhere?

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,146 @@
...
+	conduitBlockchainID, err = ids.FromString("9asUA3QckLh7vGnFQiiUJGPTx8KE4nFtP8c1wTWJuP8XiWW75")
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Implements Network, pointing to subnets deployed on Fuji
+type FujiNetwork struct{}
+
+func (g *FujiNetwork) GetSubnetsInfo() []utils.SubnetTestInfo {
```
michaelkaplan13, line 69: Nit: why `g` here? Could we make it `fj` for `FujiNetwork`?

tests/utils/utils.go
```diff
@@ -49,19 +49,20 @@ var (
...
+	rpcClient ethclient.Client,
 	tx *types.Transaction) *types.Receipt {
 
 	newHeads := make(chan *types.Header, 1)
 	subA, err := wsClient.SubscribeNewHead(ctx, newHeads)
 	Expect(err).Should(BeNil())
 	defer subA.Unsubscribe()
 
-	err = wsClient.SendTransaction(ctx, tx)
+	err = rpcClient.SendTransaction(ctx, tx)
```
michaelkaplan13, line 60: For my own understanding, what's the need to use the `rpcClient` here now when were previously able to use the `wsClient`?

tests/utils/network_setup.go
```diff
@@ -213,13 +219,11 @@ func DeployTeleporterContracts(transactionBytes []byte, deployerAddress common.A
...
 
 	for _, subnetInfo := range subnetsInfo {
-		client := subnetInfo.ChainWSClient
-
 		// Fund the deployer address
 		{
 			fundAmount := big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(10)) // 10eth
 			fundDeployerTx := createNativeTransferTransaction(ctx, subnetInfo, fundedAddress, fundedKey, deployerAddress, fundAmount)
-			SendTransactionAndWaitForAcceptance(ctx, subnetInfo.ChainWSClient, fundDeployerTx)
+			SendTransactionAndWaitForAcceptance(ctx, subnetInfo.ChainWSClient, subnetInfo.ChainRPCClient, fundDeployerTx)
```
michaelkaplan13, line 226: With these changes, I think of a few of the methods under `utils` here have become specific to running the E2E tests against a local network. Is there a way we could make that more clear?

tests/network/example_fuji_network.go
cam-schultz: Changed both this and the `LocalNetwork` receiver variable to `n` for network.

(The `g` came from `GinkgoNetwork`, which I changed to `LocalNetwork`)

tests/utils/network_setup.go
cam-schultz: All of the methods here are specific to the local network setup. I've renamed this file `local_network_setup.go` to make that clear. I've also moved some of the functions only used by the local network test from `utils.go` to a new file `local_network_utils.go`

.gitignore
```diff
@@ -36,3 +36,6 @@ docker-compose-run-local.yml
 main.log
 server.log
 tests.test
+
+# Forge docs
+contracts/docs/
```
minghinmatthewlam, line 41: I don't think this should be ignored

README.md
```diff
@@ -169,6 +169,21 @@ Next, provide the path to the `subnet-evm` repository and the path to a writeabl
 ./scripts/local/e2e_test.sh --local --subnet-evm ./subnet-evm --data-dir ~/tmp/e2e-test
 ```
 
+#### Run the E2E tests on another network
+
+The E2E tests can be run on another network by implementing the `Network` interface in `package network`. For example, the type `FujiNetwork` in `example_fuji_network.go` implements this interface, pointing to the Amplify, Bulletin, and Conduit Fuji subnets. After implementing this interface, you can run the E2E tests on this network by running a program such as:
```
minghinmatthewlam, line 174: might be worth linking where to find the fuji ABC subnet explorers

tests/basic_one_way_send.go
```diff
@@ -43,12 +46,13 @@ func BasicOneWaySend() {
 		AllowedRelayerAddresses: []common.Address{},
 		Message:                 []byte{1, 2, 3, 4},
 	}
+	log.Info("debug", "fundedAddress", fundedAddress, "fundedKey", hex.EncodeToString(fundedKey.D.Bytes()), "teleporterContractAddress", teleporterContractAddress)
```
minghinmatthewlam, line 49:
```suggestion
	log.Info("fundedAddress", fundedAddress, "fundedKey", hex.EncodeToString(fundedKey.D.Bytes()), "teleporterContractAddress", teleporterContractAddress)
```

tests/e2e_test.go
```diff
@@ -42,5 +42,5 @@ var _ = ginkgo.BeforeSuite(func() {
 var _ = ginkgo.AfterSuite(testUtils.TearDownNetwork)
 
 var _ = ginkgo.Describe("[Teleporter integration tests]", func() {
-	ginkgo.It("Send a message from Subnet A to Subnet B", BasicOneWaySend)
+	ginkgo.It("Send a message from Subnet A to Subnet B", BasicOneWaySendGinkgo)
```
minghinmatthewlam, line 45: Instead of separate `BasicOneWaySendGinkgo` function, can we call `BasicOneWaySend(&network.LocalNetwork{})` here?

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,153 @@
...
+		panic(err)
+	}
+	conduitBlockchainID, err = ids.FromString(conduitBlockchainIDStr)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Implements Network, pointing to subnets deployed on Fuji
+type FujiNetwork struct{}
```
minghinmatthewlam, line 80: can you add top of file check that `FujiNetwork` implements `Network` interface?

Also in go generally like to follow structs at top

tests/network/local_network.go
```diff
@@ -0,0 +1,34 @@
...
+	"crypto/ecdsa"
+	"math/big"
+
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/teleporter/tests/utils"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Implements Network, pointing to the network setup in local_network_setup.go
+type LocalNetwork struct{}
```
minghinmatthewlam, line 14: can you add top of file check that `LocalNetwork` implements `Network` interface?

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,153 @@
...
+	bulletinBlockchainID    ids.ID
+
+	conduitSubnetIDStr     = "wW7JVmjXp8SKrpacGzM81RBXdfcLDVY6M2DkFyArEXgtkyozK"
+	conduitBlockchainIDStr = "9asUA3QckLh7vGnFQiiUJGPTx8KE4nFtP8c1wTWJuP8XiWW75"
+	conduitWSURI           = "wss://subnets.avax.network/conduit/testnet/ws"
+	conduitRPCURI          = "https://subnets.avax.network/conduit/testnet/rpc"
+	conduitSubnetID        ids.ID
+	conduitBlockchainID    ids.ID
+
+	userAddress = common.HexToAddress("0xF0Aa98fDE5f1d08F0CCEC68A7d7A7Eae31c5E9C9")
```
minghinmatthewlam, line 44: should we also leave this empty and comment `to be supplied by user`?

tests/utils/network_setup.go
minghinmatthewlam: thoughts on moving the `local_network_setup` and `local_network_utils` files under `tests/network`? Could possibly even make it `tests/network/local/` and `tests/network/fuji/`

tests/basic_one_way_send.go
```diff
@@ -2,30 +2,33 @@ package tests
...
 	. "github.com/onsi/gomega"
 )
 
-// Ginkgo describe node that acts as a container for the teleporter e2e tests. This test suite
-// will run through the following steps in order:
-// 1. Send a transaction to the Teleporter contract on Subnet A
-// 2. Aggregate signatures and send the Warp message to Subnet B
-// 3. Verify receipt of the message on Subnet B
-func BasicOneWaySend() {
+func BasicOneWaySendGinkgo() {
```
gwen917, line 17: We have wrapper function for `BasicOneWaySend`. Is it because we do not allow to pass parameter when calling the test in [e2e_test.go](https://github.com/ava-labs/teleporter/pull/93/files#diff-311b8f116ec381569dfecd92d10675b9ff9acffcbc86d66ca4bc0aca79b6e991R45)?

tests/basic_one_way_send.go
```diff
@@ -71,7 +75,7 @@ func BasicOneWaySend() {
 		To:   &teleporterContractAddress,
 		Data: data,
 	}
-	result, err := subnetBInfo.ChainWSClient.CallContract(context.Background(), callMessage, nil)
+	result, err := subnetBInfo.ChainRPCClient.CallContract(context.Background(), callMessage, nil)
```
gwen917, line 78: What is the difference of using `ChainRPCClient` and `ChainWSClient` in here?

tests/network/example_fuji_network.go
gwen917: `var _ Network = &FujiNetwork{}`

tests/e2e_test.go
cam-schultz: No, we can't provide any arguments to the function. Ginkgo only supports functions with no parameters or return values in this interface.

tests/basic_one_way_send.go
cam-schultz: Removed this log.

.gitignore
cam-schultz: Removed to decouple this change from https://github.com/ava-labs/teleporter/pull/87. We don't check in the docs in that PR since the docs themselves contain the commit from which they were generated. This makes it difficult to keep them up to date in an automated way.

README.md
cam-schultz: Good call. Added. Though we will need to make sure to update this when these subnets are replaced.

tests/basic_one_way_send.go
cam-schultz: Yes, exactly. The different `network.Network` implementations point to the local network used in the E2E tests, as well as an example Fuji network that can be run against a live Fuji network.

tests/basic_one_way_send.go
cam-schultz: The local network nodes are configured to accept both ws and rpc traffic on the same port, so the ws client can be used execlusively. On Fuji, however, nodes are often configured to separate ws and rpc connections on different ports, so we need two distinct clients in order for that to work. In general, the ws client should only be used for `eth_subscribe` or other sub methods, while the rpc clients should be used for any RPC API calls.

tests/network/example_fuji_network.go
geoff-vball: I think the check should be here. What if we changed the implementation for both local and fuji to subscribe to events on the destination chain and wait for the given message ID until some timeout instead of checking the success the of tx we send ourselves in the local case?

tests/utils/utils.go
geoff-vball: So since we have multiple functions where there is a "correct" type of endpoint to use, it might be a good idea to make two type aliases `RPCClient` and `WSClient` that both alias `ethclient.Client`, and then our functions should accept one or the other. Right now you could pass the wrong type of client to any of these functions.

Doesn't have to be done now, I'd be happy to implement this later.

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,155 @@
...
+	Expect(err).Should(BeNil())
+	bulletinChainIDInt, err := bulletinRPCClient.ChainID(context.Background())
+	Expect(err).Should(BeNil())
+
+	conduitWSClient, err := ethclient.Dial(conduitWSURI)
+	Expect(err).Should(BeNil())
+	conduitRPCClient, err := ethclient.Dial(conduitRPCURI)
+	Expect(err).Should(BeNil())
+	conduitChainIDInt, err := conduitRPCClient.ChainID(context.Background())
+	Expect(err).Should(BeNil())
```
geoff-vball, line 104: I think we might avoid future headaches by creating an easy way to close these connections. I think adding `[]utils.SubnetTestInfo` to the `FujiNetwork` struct, and then having a function like `NewFujiNetwork` create these connections and return the `FujiNetwork` object with the connections. Then we can implement `CloseNetworkConnections` which will iterate over all the `SubnetTestInfo`s and close both the RPC and WS connections. Then we can call `defer CloseNetworkConnections()` right after `NewFujiNetwork()`.

geoff-vball (commented):
Just left a couple refactoring suggestions that might help. This is actually amazing!!

tests/e2e_test.go
minghinmatthewlam: can we rename to `BasicOneWaySendLocal` or something similar?

tests/deliver_to_wrong_chain.go
```diff
@@ -0,0 +1,80 @@
...
+
+	event, err := utils.GetSendEventFromLogs(receipt.Logs, subnetATeleporterMessenger)
+	Expect(err).Should(BeNil())
+	Expect(event.DestinationChainID[:]).Should(Equal(ids.Empty[:]))
+
+	//
+	// Relay the message to the destination
+	//
+
+	network.RelayMessage(ctx, receipt, subnetAInfo, subnetBInfo, false)
```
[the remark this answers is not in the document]
cam-schultz: The receipt is already checked in the implementation of `RelayMessage`. The final parameter, `expectSuccess` checks for a failed receipt status if `false`, and a success status if `true`.

To check the revert reason, we'd have to first make an `eth_call` and decode the return value. I'm not a big fan of this approach, since the generated Go ABI files don't include the error strings in the revert statements, so we'd have to include those strings in two separate places in order to check them. I can add this if you think it's important to confirm the specific revert reason.

tests/utils/local_network_utils.go
```diff
@@ -0,0 +1,160 @@
+package utils
```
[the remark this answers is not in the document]
cam-schultz: Sure thing! Reorganized.

tests/utils/utils.go
cam-schultz: My thought was to add `type RPCClient ethclient.Client` and `type WSClient ethclient.Client`, but Go doesn't differentiate those aliases in function arguments. I've instead replaced all function parameters that accept `ethclient.Client` with `SubnetsTestInfo`, and leave it up to function implementation to pick between the RPC and WS clients.

tests/network/example_fuji_network.go
cam-schultz: Good call. Made these change.

tests/network/example_fuji_network.go
cam-schultz: I added a timeout to the Fuji `RelayMessage` implementation, but not the local network implementation. There isn't really anything asynchronous in the local network implementation that could fail quietly, so I don't think a timeout is necessary. We can always add that in later if we start to observe tests hanging at that point.

README.md
```diff
@@ -187,7 +187,24 @@ Next, provide the path to the `subnet-evm` repository and the path to a writeabl
 ./scripts/local/e2e_test.sh --local --subnet-evm ./subnet-evm --data-dir ~/tmp/e2e-test
 ```
 
-## ABI Bindings
+#### Run the E2E tests on another network
+
+The E2E tests can be run on another network by implementing the `Network` interface in `package network`. For example, the type `FujiNetwork` in `example_fuji_network.go` implements this interface, pointing to the [Amplify](https://subnets-test.avax.network/amplify), [Bulletin](https://subnets-test.avax.network/bulletin), and [Conduit](https://subnets-test.avax.network/conduit) Fuji subnets. After implementing this interface, you can run the E2E tests on this network by running a program such as:
```
minghinmatthewlam, line 192: Is it worth linking `pakcage network` too?

minghinmatthewlam: and/or `example_fuji_network`

tests/add_fee_amount.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/ava-labs/subnet-evm/accounts/abi/bind"
+	teleportermessenger "github.com/ava-labs/teleporter/abi-bindings/go/Teleporter/TeleporterMessenger"
+	"github.com/ava-labs/teleporter/tests/network"
+	"github.com/ava-labs/teleporter/tests/utils"
+	localUtils "github.com/ava-labs/teleporter/tests/utils/local-network-utils"
+	"github.com/ethereum/go-ethereum/common"
+	. "github.com/onsi/gomega"
+)
+
+func AddFeeAmountGinkgo() {
```
minghinmatthewlam, line 16: nit: instead of `Ginkgo` can we rename to be `Local` or `_Local`?

tests/add_fee_amount.go
```diff
@@ -0,0 +1,86 @@
...
+	fundedAddress, fundedKey := network.GetFundedAccountInfo()
+	ctx := context.Background()
+
+	subnetATeleporterMessenger, err := teleportermessenger.NewTeleporterMessenger(teleporterContractAddress, subnetAInfo.ChainRPCClient)
+	Expect(err).Should(BeNil())
+	subnetBTeleporterMessenger, err := teleportermessenger.NewTeleporterMessenger(teleporterContractAddress, subnetBInfo.ChainRPCClient)
+	Expect(err).Should(BeNil())
+
+	// Use mock token as the fee token
+	mockTokenAddress, mockToken := localUtils.DeployMockToken(context.Background(), fundedAddress, fundedKey, subnetAInfo)
```
minghinmatthewlam, line 34: nit: one is deploy "mock" token, one is "exampel erc20", should they be same?

tests/utils/local-network-utils/local_network_setup.go
```diff
@@ -0,0 +1,402 @@
...
+
+	// Wait for the transaction to be mined
+	receipt, err := bind.WaitMined(ctx, source.ChainRPCClient, txn)
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	return address, mockToken
+}
+
+func DeployExampleCrossChainMessenger(ctx context.Context, fundedAddress common.Address, fundedKey *ecdsa.PrivateKey, source utils.SubnetTestInfo) (common.Address, *examplecrosschainmessenger.ExampleCrossChainMessenger) {
```
minghinmatthewlam, line 332: some utils functions pass in fundedKey/address, while some others use the var set at top, should we stick with one of them throughout?

tests/add_fee_amount.go
cam-schultz: `AddFeeAmount(network.Network)` can still be run against a local network by providing a `&network.LocalNetwork` as the argument. `AddFeeAmountGinkgo()` on the other hand is intended only to be used by a Ginkgo spec, so I think the current naming convention is the most accurate.

tests/utils/local-network-utils/local_network_setup.go
cam-schultz: Good call. Adjusted all functions that use these values to include them as parameters

tests/add_fee_amount.go
cam-schultz: Renamed `DeployMockToken` to `DeployExampleERC20`

README.md
cam-schultz: Added 👍

tests/utils/network_setup.go
minghinmatthewlam: bumping above ^

.github/workflows/linter.yml
```diff
@@ -1,27 +1,50 @@
...
+        path: teleporter
+        submodules: recursive
+
+    - name: Setup Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.20.10'
+        check-latest: true
+
+    - name: golangci-lint
```
minghinmatthewlam, line 44: should we add golangci-lint for `lint.sh` script

tests/utils/local-network-utils/local_network_setup.go
```diff
@@ -0,0 +1,442 @@
...
+	return []utils.SubnetTestInfo{
+		*subnetsInfo[subnetA],
+		*subnetsInfo[subnetB],
+	}
+}
+
+func GetTeleporterContractAddress() common.Address {
+	return teleporterContractAddress
+}
+func GetFundedAccountInfo() (common.Address, *ecdsa.PrivateKey) {
```
minghinmatthewlam, line 66: I think it's strange to have the vars `fundedAddress` and `fundedKey` in this file, as the other functions have it passed in. imo the fudned address and key should live in `local_network`

tests/utils/local-network-utils/local_network_utils.go
```diff
@@ -0,0 +1,188 @@
...
+		ctx, unsignedWarpMessageID, params.WarpQuorumDenominator,
+	)
+	Expect(err).Should(BeNil())
+
+	return signedWarpMessageBytes
+}
+
+// Constructs the aggregate signature, packs the Teleporter message, and relays to the destination
+// Returns the receipt on the destination chain
+func RelayMessage(
```
minghinmatthewlam, line 118: same as https://github.com/ava-labs/teleporter/pull/93#discussion_r1382293274, I think the `utils` that are specifically for local network, should be moved to `/tests/network/local` and some of which can take a receive of a local network too

minghinmatthewlam: That would also help clean up some extra steps, where the LocalNetwork functions are just directly calling the local_network_utils functions, when they can just live with local network

tests/add_fee_amount.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Relay message from SubnetA to SubnetB
+	network.RelayMessage(ctx, sendCrossChainMsgReceipt, subnetAInfo, subnetBInfo, false, true)
+
+	// Check Teleporter message received on the destination
+	delivered, err := subnetBTeleporterMessenger.MessageReceived(&bind.CallOpts{}, subnetAInfo.BlockchainID, messageID)
+	Expect(err).Should(BeNil())
+	Expect(delivered).Should(BeTrue())
+
+	// Send message from SubnetB to SubnetA
```
minghinmatthewlam, line 96: can you elaborate that this message back will include the Teleporter receipt for previous delivery, etc.

tests/basic_one_way_send.go
```diff
@@ -43,39 +54,24 @@ func BasicOneWaySend() {
...
-
-	bind, err := teleportermessenger.NewTeleporterMessenger(teleporterContractAddress, subnetAInfo.ChainWSClient)
-	Expect(err).Should(BeNil())
-	event, err := utils.GetEventFromLogs(receipt.Logs, bind.ParseSendCrossChainMessage)
-	Expect(err).Should(BeNil())
-	Expect(event.DestinationChainID[:]).Should(Equal(subnetBInfo.BlockchainID[:]))
-
-	teleporterMessageID = event.Message.MessageID
+	var receipt *types.Receipt
+	receipt, teleporterMessageID = utils.SendCrossChainMessageAndWaitForAcceptance(
```
minghinmatthewlam, line 59: no need extra var declaration for both these variables, we can delcare new ones here with `:=`

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+		teleporterMessageID *big.Int
+	)
+
+	subnets := network.GetSubnetsInfo()
+	subnetAInfo := subnets[0]
+	subnetBInfo := subnets[1]
+	teleporterContractAddress := network.GetTeleporterContractAddress()
+	fundedAddress, fundedKey := network.GetFundedAccountInfo()
+
+	deployerKey, _ := crypto.GenerateKey()
```
minghinmatthewlam, line 34: lets check this error as well for completeness

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+
+	//
+	// Check that the message was not successfully executed
+	//
+	executionFailedEvent, err := utils.GetEventFromLogs(
+		receipt.Logs,
+		subnetBTeleporterMessenger.ParseMessageExecutionFailed,
+	)
+	Expect(err).Should(BeNil())
+	Expect(executionFailedEvent.Message.MessageID).Should(Equal(deliveredTeleporterMessage.MessageID))
```
minghinmatthewlam, line 124: I think we should check against `teleporterMessageID` here instead, then we don't need the extra `receiveEvent` and `deliveredTeleporterMessage` parse from above since it's already checked in `RelayMessage`

minghinmatthewlam: doesn't hurt to leave i guess though, since for other network implementations liek Fuji, the `RelayMessage` doesn't necessarily check receive event. If decided to keep, can we still check against sent `teleporterMessageID`?

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+
+	//
+	// Deploy the contract on Subnet B
+	//
+	optsB := utils.CreateTransactorOpts(ctx, subnetBInfo, deployerAddress, deployerKey)
+	exampleMessengerContractB, tx, subnetBExampleMessenger, err :=
+		examplecrosschainmessenger.DeployExampleCrossChainMessenger(
```
minghinmatthewlam, line 131: we use the utils function one above, and can use it here too

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Confirm that it was deployed at the expected address
+	Expect(exampleMessengerContractB).Should(Equal(destinationContractAddress))
+
+	//
+	// Call retryMessageExecution on Subnet B
+	//
+
+	optsB = utils.CreateTransactorOpts(ctx, subnetBInfo, fundedAddress, fundedKey)
+	tx, err = subnetBTeleporterMessenger.RetryMessageExecution(optsB, subnetAInfo.BlockchainID, deliveredTeleporterMessage)
```
minghinmatthewlam, line 151: do we want a utils `RetryMessageExecutionAndWaitForAcceptance`?

tests/deliver_to_wrong_chain.go
```diff
@@ -0,0 +1,92 @@
...
+	Expect(err).Should(BeNil())
+
+	//
+	// Send a transaction to Subnet A to issue a Warp Message from the Teleporter contract to a chain other than Subnet B
+	//
+	ctx := context.Background()
+
+	sendCrossChainMessageInput := teleportermessenger.TeleporterMessageInput{
+		DestinationChainID: ids.Empty, // Some other chain ID
+		DestinationAddress: fundedAddress,
```
minghinmatthewlam, line 54: nit: can we note this is a random address, or just use address(0)

tests/deliver_to_wrong_chain.go
```diff
@@ -0,0 +1,92 @@
...
+	log.Info(
+		"Sending Teleporter transaction on source chain",
+		"destinationChainID", subnetBInfo.BlockchainID,
+		"txHash", signedTx.Hash(),
+	)
+	receipt := utils.SendTransactionAndWaitForAcceptance(ctx, subnetAInfo, signedTx, true)
+
+	event, err := utils.GetEventFromLogs(receipt.Logs, subnetATeleporterMessenger.ParseSendCrossChainMessage)
+	Expect(err).Should(BeNil())
+	Expect(event.DestinationChainID[:]).Should(Equal(ids.Empty[:]))
```
minghinmatthewlam, line 76: how does this differ from `SendCrossChainMessageAndWaitForAcceptance`

tests/example_messenger.go
```diff
@@ -0,0 +1,97 @@
...
+	//
+	// Call the example messenger contract on Subnet A
+	//
+	message := "Hello, world!"
+	optsA := utils.CreateTransactorOpts(ctx, subnetAInfo, fundedAddress, fundedKey)
+	tx, err := subnetAExampleMessenger.SendMessage(
+		optsA,
+		subnetBInfo.BlockchainID,
+		exampleMessengerContractB,
+		fundedAddress,
```
minghinmatthewlam, line 59: think it's confusing to use fundedAddress for fee contract address here, think we can use `address(0)` since fee amount is 0

tests/example_messenger.go
```diff
@@ -0,0 +1,97 @@
...
+		big.NewInt(0),
+		big.NewInt(300000),
```
minghinmatthewlam, line 61: think we should have this requiredGaslimit set in `ExampleCrossChainMessenger` as a constant

tests/insufficient_gas.go
```diff
@@ -0,0 +1,93 @@
...
+	// Deploy ExampleMessenger to Subnets A
+	_, subnetAExampleMessenger := localUtils.DeployExampleCrossChainMessenger(ctx, fundedAddress, fundedKey, subnetAInfo)
+
+	// Deploy ExampleMessenger to Subnets B
+	exampleMessengerContractB, _ := localUtils.DeployExampleCrossChainMessenger(ctx, fundedAddress, fundedKey, subnetBInfo)
+
+	// Send message from SubnetA to SubnetB with 0 execution gas, which should fail to execute
+	message := "Hello, world!"
+	optsA := utils.CreateTransactorOpts(ctx, subnetAInfo, fundedAddress, fundedKey)
+	tx, err := subnetAExampleMessenger.SendMessage(
```
minghinmatthewlam, line 46: nit: in `example_messenger.go` the same call the parameters are indented. Think the go linter should check for this

tests/insufficient_gas.go
```diff
@@ -0,0 +1,93 @@
...
+	// Check message execution failed event
+	failedMessageExecutionEvent, err := utils.GetEventFromLogs(
+		receipt.Logs, subnetBTeleporterMessenger.ParseMessageExecutionFailed,
+	)
+	Expect(err).Should(BeNil())
+	Expect(failedMessageExecutionEvent.MessageID).Should(Equal(messageID))
+	Expect(failedMessageExecutionEvent.OriginChainID[:]).Should(Equal(subnetAInfo.BlockchainID[:]))
+
+	// Retry message execution. This will execute the message with as much gas as needed
+	// (up to the transaction gas limit), rather than using the required gas specified in the message itself.s
```
minghinmatthewlam, line 79:
```suggestion
	// (up to the transaction gas limit), rather than using the required gas specified in the message itself.
```

tests/insufficient_gas.go
```diff
@@ -0,0 +1,93 @@
...
+		failedMessageExecutionEvent.Message,
+		fundedAddress,
+		fundedKey,
+		subnetBTeleporterMessenger,
+	)
+	executedEvent, err := utils.GetEventFromLogs(receipt.Logs, subnetBTeleporterMessenger.ParseMessageExecuted)
+	Expect(err).Should(BeNil())
+	Expect(executedEvent.MessageID).Should(Equal(messageID))
+	Expect(executedEvent.OriginChainID[:]).Should(Equal(subnetAInfo.BlockchainID[:]))
+}
```
minghinmatthewlam, line 93: can we also check example cross chain messengerB's `GetCurrentMessage` and make sure it's updated to message sent?

tests/utils/utils.go
```diff
@@ -48,26 +54,142 @@ var (
...
+
+	log.Info("Send AddFeeAmount transaction on source chain",
+		"messageID", messageID,
+		"sourceChainID", source.BlockchainID,
+		"destinationChainID", destination.BlockchainID)
+
+	return receipt
+}
+
+func RetryMessageExecutionAndWaitForAcceptance(
```
minghinmatthewlam, line 140: shouldn't this function take into some form of `RequireGasLimit` to make sure the transaction gets sent with gas > requiredGasLimit to pass?

tests/utils/utils.go
```diff
@@ -323,25 +451,23 @@ func calculateTxParams(ctx context.Context, wsClient ethclient.Client, fundedAdd
 	return gasFeeCap, gasTipCap, nonce
```
minghinmatthewlam, line 451: on line 449: where does the `2500000000` come from?

tests/receive_message_twice.go
```diff
@@ -0,0 +1,112 @@
...
+	log.Info(
+		"Sending Teleporter transaction on source chain",
+		"destinationChainID", subnetBInfo.BlockchainID,
+		"txHash", signedTx.Hash(),
+	)
+	receipt := utils.SendTransactionAndWaitForAcceptance(ctx, subnetAInfo, signedTx, true)
+
+	event, err := utils.GetEventFromLogs(receipt.Logs, subnetATeleporterMessenger.ParseSendCrossChainMessage)
+	Expect(err).Should(BeNil())
+	Expect(event.DestinationChainID[:]).Should(Equal(subnetBInfo.BlockchainID[:]))
```
minghinmatthewlam, line 71: can this be replaced with `SendCrossChainMessageAndWaitForAcceptance`?

tests/receive_message_twice.go
```diff
@@ -0,0 +1,112 @@
...
+		To:         &teleporterContractAddress,
+		Gas:        teleporterTx.Gas(),
+		GasFeeCap:  gasFeeCap,
+		GasTipCap:  gasTipCap,
+		Value:      big.NewInt(0),
+		Data:       teleporterTx.Data(),
+		AccessList: teleporterTx.AccessList(),
+	})
+
+	signedTx = utils.SignTransaction(secondTeleporterTx, fundedKey, subnetBInfo.ChainIDInt)
```
minghinmatthewlam, line 110: should we try to `RelayMessage` with the same receipt as above and make sure it fails?

tests/resubmit_altered_message.go
```diff
@@ -0,0 +1,80 @@
...
+	alteredMessage := make([]byte, len(teleporterMessage.Message))
+	copy(alteredMessage, teleporterMessage.Message)
+	alteredMessage[0] = ^alteredMessage[0]
+	Expect(alteredMessage[:]).ShouldNot(Equal(teleporterMessage.Message[:]))
+	teleporterMessage.Message = alteredMessage
+
+	// Resubmit the altered message
+	log.Info("Submitting the altered Teleporter message on the destination")
+	opts := utils.CreateTransactorOpts(ctx, subnetAInfo, fundedAddress, fundedKey)
+	_, err = subnetATeleporterMessenger.RetrySendCrossChainMessage(opts, subnetBInfo.BlockchainID, teleporterMessage)
```
minghinmatthewlam, line 78: can we wait for transaction to be mined, and check that the receipt has transaction failed?

tests/retry_successful_execution.go
```diff
@@ -0,0 +1,109 @@
...
+	Expect(err).Should(BeNil())
+	Expect(currMessage).Should(Equal(message))
+
+	//
+	// Attempt to retry message execution, which should fail
+	//
+	optsB := utils.CreateTransactorOpts(ctx, subnetBInfo, fundedAddress, fundedKey)
+	tx, err = subnetBTeleporterMessenger.RetryMessageExecution(optsB, subnetAInfo.BlockchainID, deliveredTeleporterMessage)
+	Expect(err).Should(Not(BeNil()))
+	Expect(tx).Should(BeNil())
```
minghinmatthewlam, line 108: why should the transaction be `nil`? Shouldn't we still be able to check transactipn receipt and make sure receipt status is failed?

tests/send_specific_receipts.go
```diff
@@ -0,0 +1,147 @@
...
+	amount, err := subnetATeleporterMessenger.CheckRelayerRewardAmount(&bind.CallOpts{}, fundedAddress, mockTokenAddress)
+	Expect(err).Should(BeNil())
+	Expect(amount).Should(Equal(totalAccumulatedRelayerFee))
+
+	// Send message from Subnet B to Subnet A to trigger the "regular" method of delivering receipts.
+	// The next message from B->A will contain the same receipts that were manually sent in the above steps,
+	// but they should not be processed again on Subnet A.
+	sendCrossChainMessageInput = teleportermessenger.TeleporterMessageInput{
+		DestinationChainID: subnetAInfo.BlockchainID,
+		DestinationAddress: fundedAddress,
```
minghinmatthewlam, line 122: lets make these random addresses some other set random address var or address(0) to not confuse with fundedAddress.

tests/send_specific_receipts.go
```diff
@@ -0,0 +1,147 @@
...
+		AllowedRelayerAddresses: []common.Address{},
+		Message:                 []byte{1, 2, 3, 4},
+	}
+
+	// This message will also have the same receipt as the previous message
+	receipt, messageID = utils.SendCrossChainMessageAndWaitForAcceptance(
+		ctx, subnetBInfo, subnetAInfo, sendCrossChainMessageInput, fundedAddress, fundedKey, subnetBTeleporterMessenger)
+
+	// Relay message from Subnet B to Subnet A
+	network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, false, true)
```
minghinmatthewlam, line 137: can we unpack the Teleporer message that we relayed here, and check that receipts include the duplicated receipt for message IDs above?

tests/unallowed_relayer.go
```diff
@@ -0,0 +1,90 @@
...
+	Expect(err).Should(BeNil())
+	Expect(event.DestinationChainID[:]).Should(Equal(subnetBInfo.BlockchainID[:]))
+
+	teleporterMessageID = event.Message.MessageID
+
+	//
+	// Relay the message to the destination
+	//
+
+	network.RelayMessage(ctx, receipt, subnetAInfo, subnetBInfo, false, false)
```
minghinmatthewlam, line 80: can we also check that relayer with the allowed relayer is successful?

tests/validator_churn.go
```diff
@@ -0,0 +1,158 @@
...
+	signedWarpMessageBytes := localUtils.ConstructSignedWarpMessageBytes(ctx, receipt, subnetAInfo, subnetBInfo)
+
+	//
+	// Modify the validator set on Subnet A
+	//
+
+	// Add new nodes to the validator set
+	log.Info("Adding nodes to the validator set")
+	var nodesToAdd []string
+	for i := 15; i <= 19; i++ {
```
minghinmatthewlam, line 85: why do we start from 15? I believe we just have two subnets, and 5 nodes each. Can we get the total number of nodes for both subnets and start from there insted of hardcoding the value?

tests/validator_churn.go
```diff
@@ -0,0 +1,158 @@
...
+		sendEvent.Message.RequiredGasLimit,
+		teleporterContractAddress,
+		fundedAddress,
+		fundedKey,
+		subnetBInfo,
+		false,
+	)
+
+	log.Info("Sending transaction to destination chain")
+	receipt = utils.SendTransactionAndWaitForAcceptance(ctx, subnetBInfo, signedTx, false)
```
minghinmatthewlam, line 122: I think we can just use `network.RelayMessage` here and expect failure

tests/utils/local-network-utils/local_network_utils.go
```diff
@@ -0,0 +1,188 @@
...
+	ctx context.Context,
+	sourceReceipt *types.Receipt,
+	source utils.SubnetTestInfo,
+	destination utils.SubnetTestInfo,
+	alterMessage bool,
```
geoff-vball, line 123: I don't think `alterMessage` belongs here as a parameter, just to satisfy one test. If we chunk the internal parts of this function into smaller helpers, then we can more easily create a `relayAlteredMessage` function and use that in the one place we need it.

tests/utils/utils.go
```diff
@@ -128,7 +275,43 @@ func CreateSendCrossChainMessageTransaction(
...
+		Nonce:     nonce,
+		To:        &teleporterContractAddress,
+		Gas:       gasLimit,
+		GasFeeCap: gasFeeCap,
+		GasTipCap: gasTipCap,
+		Value:     DefaultTeleporterTransactionValue,
+		Data:      data,
+	})
+
+	return SignTransaction(tx, fundedKey, subnetInfo.ChainIDInt)
```
geoff-vball, line 314: This should be refactored to use t`eleportermessenger.NewTeleporterMessenger` from the abi-bindings

geoff-vball (commented):
Hey! This is looking great! I went through and mostly commented on how things could be better structured to make everything easier, more readable, and less error-prone. I didn't really check any of the tests for correctness - I'll do that after this has been refactored a bit. 

I'm also more than willing to implement any of the changes I suggested. Let me know if you have any questions :)

tests/utils/network_setup.go
cam-schultz: Elected to organize the files according to this comment: https://github.com/ava-labs/teleporter/pull/93#discussion_r1393131946

tests/e2e_test.go
```diff
@@ -36,11 +36,24 @@ var _ = ginkgo.BeforeSuite(func() {
...
 	testUtils.DeployTeleporterContracts(teleporterDeployerTransaction, teleporterDeployerAddress, teleporterContractAddress)
+	testUtils.DeployTeleporterRegistryContracts(teleporterContractAddress)
 	log.Info("Set up ginkgo before suite")
 })
 
 var _ = ginkgo.AfterSuite(testUtils.TearDownNetwork)
 
 var _ = ginkgo.Describe("[Teleporter integration tests]", func() {
-	ginkgo.It("Send a message from Subnet A to Subnet B", BasicOneWaySend)
+	// Teleporter tests
```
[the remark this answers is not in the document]
cam-schultz: Deferring this to https://github.com/ava-labs/teleporter/issues/132

.github/workflows/linter.yml
cam-schultz: Good call. Added.

tests/deliver_to_nonexistent_contract.go
cam-schultz: We need the `deliveredTeleporterMessage` in any case so we can retry execution later. Will leave as is since it's functionally the same.

tests/deliver_to_wrong_chain.go
cam-schultz: Added a dummy address

tests/deliver_to_wrong_chain.go
cam-schultz: It doesn't Modified to use the util.

tests/example_messenger.go
cam-schultz: Added `constants.go` in `abi-bindings/go/CrossChainApplications/ExampleCrossChainMessenger`

tests/insufficient_gas.go
cam-schultz: No, I don't think so. There are separate formatting tools to enforce whitespace style, but that's outside the scope of this PR. I'll create a ticket to investigate and apply that to our repos

tests/utils/utils.go
cam-schultz: The transactor already estimates the tx gas and sets that as the gas limit. I don't think checking that enough gas is provided is necessary.

tests/utils/utils.go
cam-schultz: Replaced with `gasUtils.MaxPriorityFeePerGas`

tests/validator_churn.go
cam-schultz: Here, we want to deliver the message signed by the old validator set. If we use `network.RelayMessage`, it will construct a new aggregate signature using the updated validator set.

README.md
```diff
@@ -187,7 +187,24 @@ Next, provide the path to the `subnet-evm` repository and the path to a writeabl
...
+```go
+func main() {
+  // Register a failure handler that panics
+	gomega.RegisterFailHandler(func(message string, callerSkip ...int) {
+		panic(message)
+	})
+
+  // Run the test, composing it with the Network implementation
+  network := network.NewFujiNetwork()
+  defer network.CloseNetworkConnections()
```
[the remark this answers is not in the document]
cam-schultz: Might be a github ui quirk, but they are indented in my editor.

tests/basic_one_way_send.go
```diff
@@ -4,28 +4,39 @@ import (
...
+	fundedAddress, fundedKey := network.GetFundedAccountInfo()
+
+	subnetATeleporterMessenger, err := teleportermessenger.NewTeleporterMessenger(
+		teleporterContractAddress, subnetAInfo.ChainRPCClient,
+	)
+	Expect(err).Should(BeNil())
+	subnetBTeleporterMessenger, err := teleportermessenger.NewTeleporterMessenger(
+		teleporterContractAddress, subnetBInfo.ChainRPCClient,
+	)
+	Expect(err).Should(BeNil())
```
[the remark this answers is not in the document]
cam-schultz: Good call. Made this change.

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+	//
+	// Fund the deployer address on Subnet B
+	//
+	ctx := context.Background()
+
+	fundAmount := big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(10)) // 10eth
+	fundDeployerTx := utils.CreateNativeTransferTransaction(
+		ctx, subnetBInfo, fundedAddress, fundedKey, deployerAddress, fundAmount,
+	)
+	utils.SendTransactionAndWaitForAcceptance(ctx, subnetBInfo, fundDeployerTx, true)
```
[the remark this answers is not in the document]
cam-schultz: Good idea. I'll make an issue to track this. I don't know if it'll ever be worth the effort to parallelize the tests though.

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+	teleporterMessageID = sendEvent.Message.MessageID
+
+	//
+	// Relay the message to the destination
+	//
+
+	receipt = network.RelayMessage(ctx, receipt, subnetAInfo, subnetBInfo, false, true)
+	receiveEvent, err := utils.GetEventFromLogs(receipt.Logs, subnetATeleporterMessenger.ParseReceiveCrossChainMessage)
+	Expect(err).Should(BeNil())
+	deliveredTeleporterMessage := receiveEvent.Message
```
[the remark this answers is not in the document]
cam-schultz: Same consideration here.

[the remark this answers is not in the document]
cam-schultz: Same consideration here.

tests/deliver_to_wrong_chain.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+func DeliverToWrongChainGinkgo() {
+	DeliverToWrongChain(&network.LocalNetwork{})
+}
+
+func DeliverToWrongChain(network network.Network) {
+	var (
+		teleporterMessageID *big.Int
+	)
```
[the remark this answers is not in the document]
cam-schultz: Good call. Removed.

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,284 @@
...
+func (n *FujiNetwork) GetTeleporterContractAddress() common.Address {
+	return teleporterContractAddress
+}
+
+func (n *FujiNetwork) GetFundedAccountInfo() (common.Address, *ecdsa.PrivateKey) {
+	key, err := crypto.HexToECDSA(skHex)
+	Expect(err).Should(BeNil())
+
+	return userAddress, key
+}
```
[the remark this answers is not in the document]
cam-schultz: Good call. Also removed `fundedAddress` from the parameter list in the functions in `utils.go` in favor of deriving it from the `fundedKey`

tests/utils/utils.go
cam-schultz: Added `PackRetryMessageExecution` to package `teleportermessenger`

tests/utils/local-network-utils/local_network_utils.go
cam-schultz: Agreed. Refactored.

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,284 @@
...
+
+			receiveEvent, err := utils.GetEventFromLogs(l, destinationSubnetTeleporterMessenger.ParseReceiveCrossChainMessage)
+			Expect(err).Should(BeNil())
+			if receiveEvent.MessageID.Cmp(teleporterMessageID) == 0 {
+				return &types.Receipt{
+					Status: types.ReceiptStatusSuccessful,
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
cam-schultz: Let's save this for a future PR. Will create an issue.

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+	Expect(err).Should(BeNil())
+
+	// Wait for the transaction to be mined
+	receipt, err := bind.WaitMined(ctx, subnetAInfo.ChainRPCClient, tx)
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	sendEvent, err := utils.GetEventFromLogs(receipt.Logs, subnetATeleporterMessenger.ParseSendCrossChainMessage)
+	Expect(err).Should(BeNil())
+	Expect(sendEvent.DestinationChainID[:]).Should(Equal(subnetBInfo.BlockchainID[:]))
```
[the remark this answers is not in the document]
cam-schultz: Said helper would just wrap the destination ID check. I'll leave this as is for now.

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+		destinationContractAddress,
+		fundedAddress,
+		big.NewInt(0),
+		big.NewInt(300000),
+		message,
+	)
+	Expect(err).Should(BeNil())
+
+	// Wait for the transaction to be mined
+	receipt, err := bind.WaitMined(ctx, subnetAInfo.ChainRPCClient, tx)
```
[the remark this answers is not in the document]
cam-schultz: I experimented with this a bit, but wasn't able to get it working as expected. I'll create an issue and we can tackle it later.

tests/utils/local-network-utils/local_network_setup.go
cam-schultz: I'm going to keep them here, since that address is specific to the local network set up by the Ginkgo suite. A separate standalone local network wouldn't have the same set of funded keys, for instance.

cam-schultz: I'm going to keep them here, since that address is specific to the local network set up by the Ginkgo suite. A separate standalone local network wouldn't have the same set of funded keys, for instance.

tests/utils/local-network-utils/local_network_utils.go
cam-schultz: Again, I'm going to shoot to keep `local_network.go` as lightweight as possible

tests/utils/local-network-utils/local_network_setup.go
cam-schultz: I'm going to opt to keep `local_network.go` as lightweight as possible for the time being. We can revisit later if `local_network_utils.go` proves to be too verbose.

tests/resubmit_altered_message.go
cam-schultz: The first return value is `nil` in this case. I believe this is because warp message verification happens when the predicate is built, and since it fails, the block is never built.

tests/retry_successful_execution.go
cam-schultz: That's a good question. The warp message should still pass verification, so I'd expect the tx to be mined. I left a TODO and will create an issue to investigate.

tests/unallowed_relayer.go
cam-schultz: IMO that's sufficiently covered in other test cases. Adding that here would require modifying the relayer interface to accept a private key, which I don't think is worth it for an unnecessary change.

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,165 @@
...
+	//
+	// Check that the message was successfully relayed
+	//
+	delivered, err := subnetBTeleporterMessenger.MessageReceived(
+		&bind.CallOpts{},
+		subnetAInfo.BlockchainID,
+		teleporterMessageID,
+	)
+	Expect(err).Should(BeNil())
+	Expect(delivered).Should(BeTrue())
```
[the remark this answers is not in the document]
cam-schultz: Doing so would require further changes to the RelayMessage signature, so I'm going to leave it as is.

tests/deliver_to_nonexistent_contract.go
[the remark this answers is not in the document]
cam-schultz: Similar to the above, I think the current interface of RelayMessage delivering the message then returning that message's receipt makes logical sense. Any further operations on the receipt are left up to the test.

tests/deliver_to_wrong_chain.go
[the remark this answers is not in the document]
michaelkaplan13: SGTM

tests/e2e_test.go
michaelkaplan13: I'd propose getting rid of the `XXXGinkgo()` methods all together, and instead update these lines to be along the line of:
```
ginkgo.It("Send a message from Subnet A to Subnet B", func() { flows.BasicOneWaySend(&network.LocalNetwork{}) })
```

I think that will likely be the clearest for new developers looking at the code for the first time, compared to possible confusion of why the `XXXGinkgo()` methods are needed or should be used currently.

tests/network/example_fuji_network.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/ava-labs/teleporter/tests/utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/log"
+	. "github.com/onsi/gomega"
+)
+
+var _ Network = &FujiNetwork{}
+
+// Amplify, Bulletin, Conduit subnet constants
```
michaelkaplan13, line 23: Can we add a note here somewhere that the tests will not currently works against these networks due to incompatibility with the `subnet-evm` version deployed on them?

We will update in #132 either way.

tests/network/network.go
```diff
@@ -0,0 +1,22 @@
...
+	"crypto/ecdsa"
+
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/teleporter/tests/utils"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Defines the interface for the network setup functions used in the E2E tests
+type Network interface {
+	GetSubnetsInfo() []utils.SubnetTestInfo
```
michaelkaplan13, line 14: Thoughts on updating this interface to have `GetSubnetsInfo` instead return individual `SubnetTestInfo` instances? 

I think having it return a variable length slice/array makes it error prone for callers to check the result length and then index into the result afterwards.

.github/workflows/linter.yml
```diff
@@ -1,27 +1,50 @@
...
+      run: |
+        ./scripts/lint.sh
+
+  golangci:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        path: teleporter
```
minghinmatthewlam, line 35: nit: don't think we need to specify teleporter here

.github/workflows/linter.yml
minghinmatthewlam: was this added? Was thinking `/scripts/lint.sh` would also run golangci-lint

.golangci.yml
```diff
@@ -0,0 +1,29 @@
+# This file configures github.com/golangci/golangci-lint.
+
+run:
+  timeout: 3m
+  tests: true
+  # skip auto-generated files.
+  skip-files:
```
minghinmatthewlam, line 7: double cchecking if we want/is it already applying to abi bindings generated files

.github/workflows/linter.yml
cam-schultz: This follows the pattern in the same job in awm-relayer: https://github.com/ava-labs/awm-relayer/blob/main/.github/workflows/linter.yml#L19

You're probably right, but let's keep this as is for consistency.

.golangci.yml
cam-schultz: Not at the golangci-lint level, but the CI job only lints the files in `teleporter/tests` by setting that as the target directory.

cam-schultz: @gwen917 Can you add `abi-bindings/go` to this list?

tests/network/network.go
geoff-vball: @michaelkaplan13 Not sure exactly what you mean here? Do you mean create a copy of all the internal fields, ex: dialing a new wsclient?

tests/retry_successful_execution.go
geoff-vball: That's just general `go` convention to not return anything when you have an error. I would imagine we could check the error for some sort of relevant information.

cam-schultz: `RetryMessageExecution` issues a transaction to call `retryMessageExecution` on the `TeleporterMessenger` contract. I'd expect the transaction issuance to succeed (and the Go call here to return a non-nil tx), but the receipt to have a failed status after it has been mined. For example, in [SendTransactionAndWaitForAcceptance](https://github.com/ava-labs/teleporter/blob/9f4c142f9735fc8a87428e8c713776f92d469231/tests/utils/utils.go#L56) the `SendTransaction` call is successful, even if the expected receipt status is `ReceiptStatusFailed`.

Following the call to `RetryMessageExecution` eventually leads to [here](https://github.com/ava-labs/subnet-evm/blob/8beb366405008ee0f40337069e7bf124f8e23140/accounts/abi/bind/base.go#L418), which resolves to the same `ethclient.Client.SendTransaction` call as the SendTransactionAndWaitForAcceptance example I mentioned above. For some reason though, that produces an error, while the other example does not.

tests/network/network.go
gwen917: do you mean something like `GetSubnetInfoByChainID(chainID) SubnetTestInfo`?

.golangci.yml
gwen917: For abi-bindings, some of the go files are auto generated. Do you want to do lint on autogened files? It may cause issue when we do run the abi checker CI. @cam-schultz

tests/retry_successful_execution.go
geoff-vball: Yeah, I guess it's the way they've implemented it in `subnet-evm`. If `BoundContract.transact()` ever errors, it does not return the tx. If we want the behaviour you're expecting here, we could write a wrapper `alwaysReturnTx` that takes in any function that returns `(*types.Transaction, error)`. We could then call the function, and if it returns an error, set `opts.NoSend = true`, and call the function again to get the tx. It's a little bit of work and not sure if it's worth it, but the option is there.

The other option is just filing a ticket in `subnet-evm` to request the change in `BoundContract.transact()`

geoff-vball: FYI the error contains the message `"execution reverted: TeleporterMessenger: message not found"`

cam-schultz: Nice, thanks for checking the error. That gives me confidence that the test is behaving as expected. It still might be worth investigating how we can unify these behaviors, but as long as the test is correct, that is very low priority IMO.

tests/network/network.go
michaelkaplan13: I was thinking `GetSubnetInfo() (utils.SubnetTestInfo, utils.SubnetTestInfo)` such that it could be used as:
```
subnetA, subnetB := network.GetSubnetInfo()
```
rather than having to worry about bounds checking, additional assignments, etc.

`GetSubnetInfoByChainID(chainID) SubnetTestInfo` works too, though the caller may not know the blockchain ID they are looking for.

tests/add_fee_amount.go
```diff
@@ -0,0 +1,103 @@
...
+	mockTokenAddress, mockToken := localUtils.DeployExampleERC20(
+		context.Background(),
+		fundedKey,
+		subnetAInfo,
+	)
+	localUtils.ExampleERC20Approve(
+		ctx,
+		mockToken,
+		teleporterContractAddress,
+		big.NewInt(0).Mul(big.NewInt(1e18),
```
michaelkaplan13, line 38: Can this just be `big.NewInt(1e18)`? Unless there's a reason it needs to be larger, I think that's a cleaner option.

gwen917: fixed

tests/network/network.go
cam-schultz: The tests assume symmetry for all subnets that they interact with, so it doesn't really matter what we call subnet A vs subnet B in each test. IMO adding an `Expect` statement that the list returned by `GetSubnetsInfo` contains at least enough subnets to run the test is sufficient. 

My 2c.

gwen917: Agree. As long as there is enough subnets for the test, it should good. Added `Expect` check after `GetSubnetsInfo`.

.github/workflows/linter.yml
cam-schultz: I did add it to `lint.sh`, but then realized that the solhint job would fail since it does not install golangci-lint. It didn't make much sense to install golangci-lint in the solidity lint job, so I reverted that change.

It makes logical sense to have separate Go and Solidity lint jobs. It also makes sense to have a single unified lint.sh scripts that covers both languages. If we can make it clear what linter is failing, my vote would be to unify the lint.sh script and the lint job.

minghinmatthewlam: maybe we can do this with command line flags to `lint.sh`

tests/erc20_bridge_multihop.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	return receipt, event.Message.MessageID
+}
+
+func approveBridgeToken(
+	ctx context.Context,
+	source utils.SubnetTestInfo,
+	bridgeTokenAddress common.Address,
+	transtor *bridgetoken.BridgeToken,
```
minghinmatthewlam, line 435: nit:typo same for other functions in this file

.github/workflows/linter.yml
```diff
@@ -1,27 +1,54 @@
...
+    - name: Checkout teleporter
+      uses: actions/checkout@v4
+
+    - name: Install solhint
+      run: |
+        npm install solhint -g
+        solhint --version
+
+    # "solhint **/*.sol" runs differently than "solhint '**/*.sol'", where the latter checks sol files
+    # in subdirectories. The former only checks sol files in the current directory and directories one level down.
```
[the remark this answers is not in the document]
gwen917: fixed

tests/erc20_bridge_multihop.go
gwen917: fixed

geoff-vball (approved):
LGTM!

tests/basic_send_receive.go
```diff
@@ -0,0 +1,120 @@
...
+		subnetAInfo,
+		subnetBInfo,
+		sendCrossChainMessageInput,
+		fundedKey,
+	)
+
+	//
+	// Relay the message to the destination
+	//
+
```
bernard-avalabs, line 73: minor nit: remove empty line.

tests/basic_send_receive.go
```diff
@@ -0,0 +1,120 @@
...
+	delivered, err := subnetBInfo.TeleporterMessenger.MessageReceived(
+		&bind.CallOpts{}, subnetAInfo.BlockchainID, teleporterMessageID,
+	)
+	Expect(err).Should(BeNil())
+	Expect(delivered).Should(BeTrue())
+
+	//
+	// Send a transaction to Subnet B to issue a Warp Message from the Teleporter contract to Subnet A
+	//
+
```
bernard-avalabs, line 88: minor nit: remove empty line.

tests/deliver_to_wrong_chain.go
```diff
@@ -0,0 +1,84 @@
...
+		subnetAInfo,
+		subnetBInfo,
+		sendCrossChainMessageInput,
+		fundedKey,
+	)
+
+	//
+	// Relay the message to the destination
+	//
+
```
bernard-avalabs, line 69: minor nit: remove empty line.

tests/receive_message_twice.go
```diff
@@ -0,0 +1,74 @@
...
+		"destinationChainID", subnetBInfo.BlockchainID,
+	)
+	receipt, teleporterMessageID := utils.SendCrossChainMessageAndWaitForAcceptance(
+		ctx, subnetAInfo, subnetBInfo, sendCrossChainMessageInput, fundedKey,
+	)
+
+	//
+	// Relay the message to the destination
+	//
+
```
bernard-avalabs, line 52: minor nit: remove empty line

tests/network/example_fuji_network.go
gwen917: added [here](https://github.com/ava-labs/teleporter/pull/93/files#diff-73cd34956e9a9bb0832f207ea9a0c60067ea99aeec1509dc291f1de930606adaR24-R25)

tests/unallowed_relayer.go
```diff
@@ -0,0 +1,69 @@
...
+		"destinationChainID", subnetBInfo.BlockchainID,
+	)
+	receipt, teleporterMessageID := utils.SendCrossChainMessageAndWaitForAcceptance(
+		ctx, subnetAInfo, subnetBInfo, sendCrossChainMessageInput, fundedKey,
+	)
+
+	//
+	// Relay the message to the destination
+	//
+
```
bernard-avalabs, line 54: minor nit: remove empty line

tests/e2e_test.go
gwen917: fixed.

bernard-avalabs (approved):
LGTM. I left some (very) minor formatting comments. I'm not sure if the extra empty lines are intentional. Please disregard if they are.

This is a very large PR. We may want to break them up into smaller PRs in the future if possible.

michaelkaplan13 (approved):
LGTM! 

A handful of the outstanding TODOs should be taken care of in https://github.com/ava-labs/teleporter/issues/132 once I'm able to update it with the latest changes from here.

geoff-vball (dismissed):

minghinmatthewlam (dismissed):

geoff-vball (dismissed):

minghinmatthewlam (approved):
LGTM

geoff-vball (approved):
👍

PR #101, opened 2023-11-02
pull from ava-labs/foundry

cam-schultz:
## Why this should be merged
Updates the integration tests to use a stable tagged release of `ava-labs/foundry` instead of the volatile nightly releases from `foundry-rs/foundry`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,6 +37,7 @@ ENV LANG en_US.utf8
 # Install foundry from specific commit
 SHELL ["/bin/bash", "-c"]
 RUN curl -L https://raw.githubusercontent.com/ava-labs/foundry/v0.1.0/foundryup/install > /tmp/foundry-install-script && \
+    sed -i 's/\/ava-labs\/foundry\/master\/foundryup/\/ava-labs\/foundry\/v0.1.0\/foundryup/g' /tmp/foundry-install-script && \
     cat /tmp/foundry-install-script | bash && \
     echo "export PATH=\"$PATH:/$HOME/.foundry/bin\"">> ~/.bashrc && \
     source ~/.bashrc && \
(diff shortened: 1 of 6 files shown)


docker/Dockerfile
```diff
@@ -36,13 +36,12 @@ ENV LANG en_US.utf8
 
 # Install foundry from specific commit
 SHELL ["/bin/bash", "-c"]
-RUN curl -L https://raw.githubusercontent.com/foundry-rs/foundry/037b3bc9cebd545fe3a4f05a32bf7efb82afdbd8/foundryup/install > /tmp/foundry-install-script && \
-    sed -i 's/\/foundry-rs\/foundry\/master\/foundryup/\/foundry-rs\/foundry\/037b3bc9cebd545fe3a4f05a32bf7efb82afdbd8\/foundryup/g' /tmp/foundry-install-script && \
+RUN curl -L https://raw.githubusercontent.com/ava-labs/foundry/master/foundryup/install > /tmp/foundry-install-script && \
```
michaelkaplan13, line 39: Still not in love with the fact that changes to the `foundryup` script in the primary branch will get pulled in without an explicit update. Could we use the `v0.1.0` tag for that too?

cam-schultz: Yup, good call

michaelkaplan13: To the same end, we should probably keep the `sed` statement to use the specific commit/tag here: https://github.com/ava-labs/foundry/blob/v0.1.0/foundryup/install#L11

michaelkaplan13: Or if you don't think that's necessary, would recommend piping the CURL output directly into `bash` rather than writing to a temp file.

michaelkaplan13:
This was a great idea @cam-schultz! 

Want to call out that we also use a `foundryup` installation in `abi_binding.sh`. We may want to update that installation also, though it may require adding more build images that can be installed from our fork for other platforms.

michaelkaplan13 (approved):

minghinmatthewlam (approved):

PR #102, opened 2023-11-02
Add max warnings check for linting

minghinmatthewlam:
## Why this should be merged
Previously linter passes when there are warnings but no errors, adding a max warning of 0 check.

## How this works
Update `lint.sh`, and disabled `private-vars-leading-underscore` for two libraries.


geoff-vball (approved):

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -22,6 +22,7 @@ library ReceiptQueue {
     // The maximum number of receipts to include in a single message.
     uint256 private constant _MAXIMUM_RECEIPT_COUNT = 5;
 
+    // solhint-disable private-vars-leading-underscore
```
cam-schultz, line 25: What's the motivation for removing this check here? Same question in SafeERC20TransferFrom

minghinmatthewlam: I placed it there so the rule still applies for state variable naming, but think it's more strange using library code in `TeleporterMessenger` with the underscore for functions like `queue._size()` and `queue._enqueue()`, etc. So removed them for the functions of the two libraries.

Thoughts here? Would be down to just change it all for libraries as well.

cam-schultz (approved):

PR #103, opened 2023-11-02
Remove WarpProtocolRegistry

### LINKED ISSUE #77: Remove WarpProtocolRegistry and refactor TeleporterRegistry
**Context and scope**
Current implementation has `WarpProtocolRegistry`, which was meant to be inherited by other protocols on top of Warp, but Teleporter repo wouldn't be the appropriate place to leave the contract, also we don't have other protocols built on top of Warp right now.

Remove the `WarpProtocolRegistry` and move logic into TeleporterRegistry.

**Open questions**
Do we want to keep an external->internal matching pattern for functions, or just public? 
Would Teleporter anycasting need a separate registry?

minghinmatthewlam:
## Why this should be merged
Currently we have an abstract `WarpProtocolRegistry`, but this contract shouldn't be saved in Teleporter repo. We will remove contract for now, and can refactor out an abstract contract in future if needed. Fixes #77 

## How this works
Move all previous logic handled by inheriting `WarpProtocolRegistry` into `TeleporterRegistry` directly. Also revisits some of the function visiblity, previously we had an `external` function that directly calls its `internal` function that holds the logic, which can be harder to read. Instead, since `TeleporterRegistry` isn't meant to be inherited we can go with `public`, also removing `getLatestVersion()` in favor of public `latestVersion` state variable.

## How this was tested
updated unit tests, CI

## How is this documented
comments, README


contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,109 @@
 
 pragma solidity 0.8.18;
 
-import "../../WarpProtocolRegistry.sol";
 import "../ITeleporterMessenger.sol";
+import "@subnet-evm-contracts/interfaces/IWarpMessenger.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
```
cam-schultz, line 10: Per the OZ audit, let's explicity import the types of interest rather than the entire contract

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,109 @@
...
+        WarpMessenger(0x0200000000000000000000000000000000000005);
+
+    bytes32 public immutable blockchainID;
+
+    // The latest protocol version. 0 means no protocol version has been added, and isn't a valid version.
+    uint256 public latestVersion;
+
+    // Mappings that keep track of the protocol version and corresponding contract address.
+    mapping(uint256 => address) internal _versionToAddress;
+    mapping(address => uint256) internal _addressToVersion;
```
cam-schultz, line 40: Also per the audit, let's name the map key value parameters

cam-schultz: Also should we make these private since `TeleporterRegistry` isn't intended to be inherited from?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +118,81 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+        return _addressToVersion[protocolAddress];
+    }
+
+    /**
+     * @dev Adds the new protocol version address to the registry.
+     * Updates latest version if the version is greater than the current latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Note: `protocolAddress` doesn't have to be a contract address, this is primarily
+     * to support the case we want to register a new protocol address meant for a security patch
```
cam-schultz, line 169:
```suggestion
     * to support the case in which we want to register a new protocol address meant for a security patch
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +118,81 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+        require(
+            version <= latestVersion,
+            "TeleporterRegistry: invalid version"
+        );
+        return _versionToAddress[version];
+    }
+
+    /**
+     * @dev Gets the version of the given `protocolAddress`.
+     * If `protocolAddress` is not a registered protocol address, returns 0, which is an invalid version.
```
cam-schultz, line 155: In `getAddressFromVersion`, we revert if an invalid version is provided, but here we return 0 if an invalid address is provided. Do we want to revert in this case as well?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,109 @@
...
+
+        for (uint256 i = 0; i < initialEntries.length; i++) {
+            _addToRegistry(initialEntries[i]);
+        }
+    }
+
+    /**
+     * @dev Gets and verifies a Warp off-chain message, and adds the new protocol version
+     * address to the registry.
+     * If a version is greater than the current latest version, it will be set as the latest version.
```
bernard-avalabs, line 66: It might be worth mentioning in the comment here that the version is in the off-chain message (entry.version).

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,109 @@
...
+     * Emits a {AddProtocolVersion} event when successful.
      * Requirements:
      *
-     * - `version` must be a valid version, i.e. greater than 0 and not greater than the latest version.
+     * - a valid Warp off-chain message must be provided.
+     * - source chain ID must be the same as the blockchain ID of the registry.
+     * - origin sender address must be the same as the `VALIDATORS_SOURCE_ADDRESS`.
+     * - destination chain ID must be the same as the blockchain ID of the registry.
+     * - destination address must be the same as the address of the registry.
+     * - version must not be zero.
```
bernard-avalabs, line 78: It might also be worth mentioning which of the requirements are checked here, and which are checked in _addToRegistry.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +118,81 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+    }
+
+    /**
+     * @dev Adds the new protocol version address to the registry.
+     * Updates latest version if the version is greater than the current latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Note: `protocolAddress` doesn't have to be a contract address, this is primarily
+     * to support the case we want to register a new protocol address meant for a security patch
+     * before the contract is deployed, to prevent the vulnerabilitiy from being exposed before registry update.
```
bernard-avalabs, line 170: nit: "the" registry update.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +118,81 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+     *
+     * - `version` is not zero
+     * - `version` is not already registered
+     * - `protocolAddress` is not zero address
+     */
+    function _addToRegistry(ProtocolRegistryEntry memory entry) private {
+        require(entry.version != 0, "TeleporterRegistry: zero version");
+        // Check that the version has not previously been registered.
+        require(
+            _versionToAddress[entry.version] == address(0),
```
bernard-avalabs, line 181: Should we allow multiple versions pointing to the same protocolAddress? If not, should we check _addressToVersion[entry.protocolAddress] != 0?

bernard-avalabs (commented):
Generally looks good to me. I've left a few minor comments.

contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
```diff
@@ -81,9 +81,7 @@ contract ERC20BridgeTest is Test {
 
         vm.expectCall(
             MOCK_TELEPORTER_REGISTRY_ADDRESS,
-            abi.encodeWithSelector(
-                WarpProtocolRegistry.getLatestVersion.selector
-            )
+            abi.encodeWithSignature("latestVersion()")
```
michaelkaplan13, line 84: Doesn't matter to me either way, but just curious if it's possible to use `TeleporterRegistry.latestVersion.selector` version here or not?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,109 @@
...
+     * - destination address must be the same as the address of the registry.
+     * - version must not be zero.
+     * - version must not already be registered.
+     * - protocol address must not be zero address.
      */
-    function getTeleporterFromVersion(
-        uint256 version
-    ) external view returns (ITeleporterMessenger) {
-        return ITeleporterMessenger(_getAddressFromVersion(version));
+    function addProtocolVersion(uint32 messageIndex) external {
```
michaelkaplan13, line 82: When you update this branch with `main`, `message.destinationAddress` will no longer be available.

I think we need to have some part of the off-chain message contain `address(this)` still though, in order to support multiple registry contracts on a single chain that should be updated independently. My suggestion would be for the message payload to be something like `abi.encode(registryAddress, ProtocolRegistryEntry);` so then we can decode and check the registryAddress the message is intended for.

An alternative would be to require that `message.originSenderAddress == address(this)`, since it isn't possible for this contract to send messages either. That would remove the need for `VALIDATORS_SOURCE_ADDRESS`, but may be slightly less clear convention.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: Yup we should, in the case we want to revert to a previous protocol address and set it as "latestVersion", users will use the same previoius address

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: updated to private added key value naming

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: same thoughts on the first approach. Think the latter is more confusing in terms of "why is this contract sending a message to itself", etc.

contracts/src/CrossChainApplications/ERC20Bridge/tests/ERC20BridgeTests.t.sol
minghinmatthewlam: updated, for the public state variable getter functions, the access to their function is different, needed to use an isntance of the contract

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: hmm ye I thought about this too, but `getAddressFromVersion` is used to check for `onlyAllowedTeleporter` in `TeleporterUpgradeable` as well. If we add an extra revert here that'd still be okay for the modifier, but it'd revert with a different error message. No strong pref here, thoughts?

Added a check that the parameter `protocolAddress` is not zero address.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +111,85 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+            "TeleporterRegistry: zero protocol address"
+        );
+
+        _versionToAddress[entry.version] = entry.protocolAddress;
+        _addressToVersion[entry.protocolAddress] = entry.version;
+        emit AddProtocolVersion(entry.version, entry.protocolAddress);
+
+        // Set latest version if the version is greater than the current latest version.
+        if (entry.version > latestVersion) {
+            latestVersion = entry.version;
```
michaelkaplan13, line 192: Should we have a `LatestVersionUpdated` event?

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +111,85 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+     * @dev Gets the address of a protocol version.
+     * Requirements:
+     *
+     * - `version` must be a valid version, i.e. greater than 0 and not greater than the latest version.
+     */
+    function getAddressFromVersion(
+        uint256 version
+    ) public view returns (address) {
+        require(version != 0, "TeleporterRegistry: zero version");
+        require(
```
michaelkaplan13, line 139: The current behavior here is to revert if `version > latestVersion`, but not revert otherwise if `version` isn't contained in `_versionToAddress`. I think we should have the same behavior in all cases the `version` isn't explicit set in `_versionToAddress`.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
michaelkaplan13: My preference would be to have the same pattern of behavior for both methods in the event they are given either a version or address that isn't set. No strong preference between them both reverting or both returning the default value (`0` or `address(0)`) in those cases though.

minghinmatthewlam: checked it so that both functions check for registered inputs, and reverts if the input given is not registered.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: added

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: good point, mentioned above but changed it to check for registered inputs, and to revert if not.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +119,92 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+        return version;
+    }
+
+    /**
+     * @dev Adds the new protocol version address to the registry.
+     * Updates latest version if the version is greater than the current latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Note: `protocolAddress` doesn't have to be a contract address, this is primarily
+     * to support the case in which we want to register a new protocol address meant for a security patch
```
michaelkaplan13, line 179: Lets remove any use of "we" in the contract comments.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +119,92 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+        uint256 version = _addressToVersion[protocolAddress];
+        require(version != 0, "TeleporterRegistry: protocol address not found");
+        return version;
+    }
+
+    /**
+     * @dev Adds the new protocol version address to the registry.
+     * Updates latest version if the version is greater than the current latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
```
michaelkaplan13, line 177: We should update the comment to note `LatestVersionUpdated` here and above.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,110 @@
...
+        latestVersion = 0;
+
+        for (uint256 i = 0; i < initialEntries.length; i++) {
+            _addToRegistry(initialEntries[i]);
+        }
+    }
+
+    /**
+     * @dev Gets and verifies a Warp off-chain message, and adds the new version to procotol address mapping,
+     * specified in the Warp off-chain message, to the registry.
```
michaelkaplan13, line 74: Receives a Warp off-
```suggestion
     * @dev Receives a Warp off-chain message containing a new protocol version and address to be registered,
     * and adds the the new values to the respective mappings.
```

Technically the verification of the message happens in the precompile, not here :)

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -5,28 +5,111 @@
...
+
+    /**
+     * @dev Receives a Warp off-chain message containing a new protocol version and address to be registered,
+     * and adds the the new values to the respective mappings.
+     * If a version is greater than the current latest version, it will be set as the latest version.
+     * If a version is less than the current latest version, it is added to the registry, but
+     * doesn't change the latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Emits a {LatestVersionUpdated} event when a new protocol version great than the current latest version is added.
```
michaelkaplan13, line 80:
```suggestion
     * Emits a {LatestVersionUpdated} event when a new protocol version greater than the current latest version is added.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +120,93 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+        require(version != 0, "TeleporterRegistry: protocol address not found");
+        return version;
+    }
+
+    /**
+     * @dev Adds the new protocol version address to the registry.
+     * Updates latest version if the version is greater than the current latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Emits a {LatestVersionUpdated} event when a new protocol version great than the current latest version is added.
```
michaelkaplan13, line 179:
```suggestion
     * Emits a {LatestVersionUpdated} event when a new protocol version greater than the current latest version is added.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -37,6 +120,93 @@ contract TeleporterRegistry is WarpProtocolRegistry {
...
+
+    /**
+     * @dev Adds the new protocol version address to the registry.
+     * Updates latest version if the version is greater than the current latest version.
+     *
+     * Emits a {AddProtocolVersion} event when successful.
+     * Emits a {LatestVersionUpdated} event when a new protocol version greater than the current latest version is added.
+     * Note: `protocolAddress` doesn't have to be a contract address, this is primarily
+     * to support the case of registering a new protocol address meant for a security patch
+     * before the contract is deployed, and prevent the vulnerabilitiy from being exposed before the registry update.
```
michaelkaplan13, line 182:
```suggestion
     * Note: `protocolAddress` doesn't have to be a contract address. This allows for registering a new protocol 
     * address before the contract is deployed. This may be the desired order of events to prevent publishing the 
     * patch for a vulnerability in the current version before the update takes effect. 
```

Nit: The current wording seemed a little cumbersome.

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -61,26 +70,23 @@ contract TeleporterRegistry {
     }
 
     /**
-     * @dev Gets and verifies a Warp off-chain message, and adds the new protocol version
-     * address to the registry.
+     * @dev Receives a Warp off-chain message containing a new protocol version and address to be registered,
+     * and adds the the new values to the respective mappings.
```
cam-schultz, line 74:
```suggestion
     * and adds the new values to the respective mappings.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -61,26 +70,23 @@ contract TeleporterRegistry {
...
-     * - destination chain ID must be the same as the blockchain ID of the registry.
-     * - destination address must be the same as the address of the registry.
-     * - version must not be zero.
-     * - version must not already be registered.
-     * - protocol address must not be zero address.
+     * - destination address must be the same as the address of this registry.
      */
     function addProtocolVersion(uint32 messageIndex) external {
-        // Get and validate for a Warp off-chain message.
+        // Get the verified Warp message, and check that it was sent off-chain to this registry.
```
cam-schultz, line 89:
```suggestion
        // Get the verified Warp message, and check that it was sent to this registry via a Warp off-chain message.
```

Suggest the above change to maintain consistent terminology

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -125,7 +127,7 @@ contract TeleporterRegistry {
      * @dev Gets the {ITeleporterMessenger} contract of the given `version`.
      * Requirements:
      *
-     * - `version` must be a valid version, i.e. greater than 0 and not greater than the latest version.
+     * - `version` must be a valid version registered.
```
cam-schultz, line 130:
```suggestion
     * - `version` must be a valid registered version.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -137,37 +139,47 @@ contract TeleporterRegistry {
      * @dev Gets the address of a protocol version.
      * Requirements:
      *
-     * - `version` must be a valid version, i.e. greater than 0 and not greater than the latest version.
+     * - `version` must be a valid version registered.
```
cam-schultz, line 142:
```suggestion
     * - `version` must be a valid registered version.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -137,37 +139,47 @@ contract TeleporterRegistry {
...
+            "TeleporterRegistry: version not found"
         );
-        return _versionToAddress[version];
+        return protocolAddress;
     }
 
     /**
      * @dev Gets the version of the given `protocolAddress`.
-     * If `protocolAddress` is not a registered protocol address, returns 0, which is an invalid version.
+     * Rrequirements:
```
cam-schultz, line 158:
```suggestion
     * Requirements:
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -137,37 +139,47 @@ contract TeleporterRegistry {
...
+     *
+     * - `protocolAddress` must be a valid protocol address registered.
```
cam-schultz, line 160:
```suggestion
     * - `protocolAddress` must be a valid registered protocol address.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -137,37 +139,47 @@ contract TeleporterRegistry {
...
+        return version;
     }
 
     /**
      * @dev Adds the new protocol version address to the registry.
      * Updates latest version if the version is greater than the current latest version.
      *
      * Emits a {AddProtocolVersion} event when successful.
+     * Emits a {LatestVersionUpdated} event when a new protocol version greater than the current latest version is added.
      * Note: `protocolAddress` doesn't have to be a contract address, this is primarily
```
cam-schultz, line 180:
```suggestion
     * Note: `entry.protocolAddress` doesn't have to be a contract address, this is primarily
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -137,37 +139,47 @@ contract TeleporterRegistry {
...
-     * to support the case we want to register a new protocol address meant for a security patch
-     * before the contract is deployed, to prevent the vulnerabilitiy from being exposed before registry update.
+     * to support the case of registering a new protocol address meant for a security patch
+     * before the contract is deployed, and prevent the vulnerabilitiy from being exposed before the registry update.
```
cam-schultz, line 182:
```suggestion
     * before the contract is deployed, and prevent the vulnerability from being exposed before the registry update.
```

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
cam-schultz, line 180: I'm not sure this note is necessary to include here. I think what it's describing is that the contract to be registered does not have to actually be deployed. But the registry doesn't call any of the registered contracts directly, so I don't think the language in the registry contract needs to make the distinction between deployed and undeployed contracts at all.

If you feel like this comment should stay, I'd suggest simplifying it to something like "entry.protocolAddress does not need to be a deployed contract. This allows the registry to be updated ahead of a protocol upgrade being deployed"

cam-schultz (commented):
A handful of small nits, typos, and clarification suggestions. The code LGTM though

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: updated

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
minghinmatthewlam: I wanted to leave the note about this being possibly used in case where its a patch for the current version, updated the wording.

bernard-avalabs (approved):
LGTM

contracts/src/Teleporter/upgrades/TeleporterRegistry.sol
```diff
@@ -177,14 +177,12 @@ contract TeleporterRegistry {
      *
      * Emits a {AddProtocolVersion} event when successful.
      * Emits a {LatestVersionUpdated} event when a new protocol version greater than the current latest version is added.
-     * Note: `protocolAddress` doesn't have to be a contract address, this is primarily
-     * to support the case of registering a new protocol address meant for a security patch
-     * before the contract is deployed, and prevent the vulnerabilitiy from being exposed before the registry update.
+     * Note: `entry.protocolAddress` doesn't have to be a contract address, allowing a new protocol address to be registered before the contract is deployed. This may be the desired order of events to prevent publishing the patch for a vulernability in the current version before the registry is updated to the new version.
```
cam-schultz, line 180: I don't love the use of the word "vulnerability" here, as it could cause unnecessary FUD. I think we should remove the second sentence altogether, and not speculate too much on use cases.
```suggestion
     * Note: `entry.protocolAddress` doesn't have to be a contract address, allowing a new protocol address to be registered before the contract is deployed.
```

minghinmatthewlam: hmm ye I'm good with the first line covering how the registry can be used.

cam-schultz (approved):
LGTM

michaelkaplan13 (approved):

Issue #94, opened 2023-10-31
Add Teleporter E2E test cases


cam-schultz:
**Context and scope**
There are a number of Teleporter-specific edge cases that we should add end-to-end tests for. Many of these are implicitly covered by unit tests already, but we should add integration level tests as well. Below are the test cases to add. As they are implemented, we should check the box and link the corresponding PR.

(Check next to test case means it's in progress; a check next to the PR means it's been merged)
- [x] Message execution runs out of gas, and execution needs to be retried
  - [x] PR: https://github.com/ava-labs/teleporter/pull/110
- [x] Adding an extra fee amount after a message is submitted, checking relayer gets total amount as reward
  - [x] PR: https://github.com/ava-labs/teleporter/pull/109 
- [x] Relayer manually retries sending of specific receipts.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/113
- [x] Test receiving a receipt multiple times has no effect
  - [x] PR: https://github.com/ava-labs/teleporter/pull/113 
- [x] Submit a message, large validator set turn over before the message is delivered. Message can be resubmitted on the sending chain to be retried
  - Requires changes to https://github.com/ava-labs/avalanche-network-runner
  - [x] PR: https://github.com/ava-labs/teleporter/pull/133
- [x] Try resubmitting a message that has been slightly altered, ensure it doesn't work.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/121
- [x] Try receiving the same message multiple times, ensure only the first one succeeds.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/105
- [x] Try delivering a message from a relayer that isn't allowed, ensure it doesn't work.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/105
- [x] Try re-executing a message that has previously succeeded, ensure it doesn't work.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/105
- [x] Deliver message destined for contract that doesn't exist, should fail. Then deploy the contract to the intended address and retry message execution, should succeed.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/104 
- [x] Try delivering message to a different chain than it was meant for, should fail.
  - [x] PR: https://github.com/ava-labs/teleporter/pull/104 
- [x] Relayer modifies a message in flight. This should fail warp verification
  - [x] PR: https://github.com/ava-labs/teleporter/pull/121

[...]

PR #40, opened 2023-09-26
Native to Native Token Bridge

geoff-vball:
## How is this documented

Issue #27, opened 2023-09-19
Author integration tests using Ginkgo


cam-schultz:
**Action items**
The Ginkgo test framework has already been implemented in other tickets, for example https://github.com/ava-labs/teleporter/issues/94. Much of the discussion in the following sections have already been resolved and implemented, but I'll leave the text there for context.

The main remaining action item to close this ticket out is to port the existing bash integration tests to our Ginkgo framework. Below are the test cases we need to port. 

- [x] example_messenger.sh
  - PR: https://github.com/ava-labs/teleporter/pull/93 
- [x] basic_send_receive.sh
  - PR: https://github.com/ava-labs/teleporter/pull/139
- [x] erc20_bridge_multihop.sh
  - PR: https://github.com/ava-labs/teleporter/pull/140 
- [x] send_specified_receipts.sh
  - PR: https://github.com/ava-labs/teleporter/pull/93 
- [x] block_hash_publisher_receiver.sh
  - PR: https://github.com/ava-labs/teleporter/pull/163
- [ ] Remove bash integration test setup
  - PR: https://github.com/ava-labs/teleporter/pull/168

**Context and scope**
The current integration test setup consists of a set of docker containers that:
1. Spin up a local avalanche network and deploys subnets and Teleporter contracts
2. Runs an `awm-relayer` instance
3. Executes integration tests

These steps are performed within the containers using a set of bash scripts. For example, https://github.com/ava-labs/teleporter/blob/main/docker/run_setup.sh starts the networks, and https://github.com/ava-labs/teleporter/blob/main/scripts/local/test.sh runs the integration tests.

**Discussion and alternatives**
[Ginkgo](https://onsi.github.io/ginkgo/) is a test framework written in Go that performs very similar steps to our current setup, but in a much more user friendly and less error prone way. Ginkgo:
- handles container setup and teardown
- is easier to avoid bugs when authoring tests due to Go's static typing
- produces much more readable tests with clearly defined expected behavior
- is under active development

To convert our bash-based integration tests to Ginkgo, the following development strategy is proposed:
1. Port `awm-relayer` E2E tests written in Ginkgo to `teleporter`. The high-level changes here are:
  a. Implement basic send and receive integration test.
  b. Modify message relaying to either build and run `awm-relayer` directly, or manually aggregate signatures and deliver the message from within the E2E test. `subnet-evm` Warp E2E tests take the latter approach.
[...]

PR #104, opened 2023-11-02
Teleporter test cases - part 1

cam-schultz:
## Why this should be merged
- Deliver message destined for contract that doesn't exist, should fail. Then deploy the contract to the intended address and retry message execution, should succeed.
- Try delivering message to a different chain than it was meant for, should fail.

Also implements the ExampleCrossChainMessenger integration test in Ginkgo. This is already implemented in bash, but all bash tests will eventually be ported to Ginkgo. This partially addresses https://github.com/ava-labs/teleporter/issues/27

Ports a handful of E2E utility changes from https://github.com/ava-labs/teleporter/pull/40

## How this works
See above

## How this was tested
New E2E test cases

## How is this documented
N/A


tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,146 @@
...
+	fundDeployerTx := utils.CreateNativeTransferTransaction(ctx, subnetBInfo, fundedAddress, fundedKey, deployerAddress, fundAmount)
+	utils.SendTransactionAndWaitForAcceptance(ctx, subnetBInfo.ChainWSClient, subnetBInfo.ChainRPCClient, fundDeployerTx, true)
+
+	//
+	// Deploy ExampleMessenger to Subnet A, but not to Subnet B
+	// Send a message that should fail to be executed on Subnet B
+	//
+
+	optsA := utils.CreateTransactorOpts(ctx, subnetAInfo, fundedAddress, fundedKey)
+	_, tx, subnetAExampleMessenger, err := examplecrosschainmessenger.DeployExampleCrossChainMessenger(optsA, subnetAInfo.ChainRPCClient, subnetAInfo.TeleporterRegistryAddress)
```
gwen917, line 57: Do we want to move contract deployment to utils? so that it can be re-used.

tests/deliver_to_nonexistent_contract.go
```diff
@@ -0,0 +1,146 @@
...
+	signedTx := utils.CreateRetryMessageExecutionTransaction(
+		ctx,
+		subnetBInfo,
+		subnetAInfo.BlockchainID,
+		deliveredTeleporterMessage,
+		fundedAddress,
+		fundedKey,
+		teleporterContractAddress,
+	)
+	utils.SendTransactionAndWaitForAcceptance(ctx, subnetBInfo.ChainWSClient, subnetBInfo.ChainRPCClient, signedTx, true)
```
gwen917, line 138: do we want to use abi binding to call `RetryMessageExecution` directly?

cam-schultz: Good call, fixed now

tests/deliver_to_nonexistent_contract.go
cam-schultz: Good call. We use this contract frequently to test Teleporter edge cases, so it makes sense to have this a widely available utility. I've added this in.

gwen917 (approved):

PR #105, opened 2023-11-03
Teleporter test cases 2

cam-schultz:
## Why this should be merged
- Try receiving the same message multiple times, ensure only the first one succeeds.
- Try delivering a message from a relayer that isn't allowed, ensure it doesn't work.
- Try re-executing a message that has previously succeeded, ensure it doesn't work.

## How this works
See above

## How this was tested
New E2E test cases

## How is this documented
N/A


tests/receive_message_twice.go
```diff
@@ -0,0 +1,106 @@
...
+
+	log.Info("debug",
+		"gas", secondTeleporterTx.Gas(),
+		"gasFeeCap", secondTeleporterTx.GasFeeCap(),
+		"gasTipCap", secondTeleporterTx.GasTipCap(),
+		"data", secondTeleporterTx.Data(),
+		"accesslist", secondTeleporterTx.AccessList(),
+	)
+	signedTx = utils.SignTransaction(secondTeleporterTx, fundedKey, subnetBInfo.ChainIDInt)
+	utils.SendTransactionAndWaitForAcceptance(ctx, subnetBInfo.ChainWSClient, subnetBInfo.ChainRPCClient, signedTx, false)
```
gwen917, line 105: is there anything we need to check after transaction sent?

tests/retry_successful_execution.go
```diff
@@ -0,0 +1,111 @@
...
+	signedTx := utils.CreateRetryMessageExecutionTransaction(
+		ctx,
+		subnetBInfo,
+		subnetAInfo.BlockchainID,
+		deliveredTeleporterMessage,
+		fundedAddress,
+		fundedKey,
+		teleporterContractAddress,
+	)
+	utils.SendTransactionAndWaitForAcceptance(ctx, subnetBInfo.ChainWSClient, subnetBInfo.ChainRPCClient, signedTx, false)
```
gwen917, line 110: same here. Do we want to use the abi bindings to call `RetryMessageExecution` directly?

cam-schultz: Changed

tests/receive_message_twice.go
cam-schultz: The final `bool` parameter of `SendTransactionAndWaitForAcceptance` indicates whether the tx is expected to fail, which it is in this case.

gwen917 (approved):

PR #106, opened 2023-11-03
OpenZeppelin Audit Cleanup Items

michaelkaplan13:
## Why this should be merged

Addresses clean up items from the first version of the OpenZeppelin audit report (to be published at a later date).

## How this works
- Sets the visibility level to private for non-constant and non-immutable state variables in `abstract` contracts to avoid inheriting contracts from being able to override their behavior.
- Adds docstrings to all structs.
- Adds security contact link to contract source code.
- Removes unused state variables.
- Adds named parameters of mappings.
- Removes unused named return variables in favor of explicitly returned values.
- Uses explicit imports for contracts under `Teleporter/`
- Makes `relayerRewardAddresses` and `relayerRewardAmounts` mappings `internal` to avoid redundant getter functions. Preferable than removing the explicit getter functions which are defined in the interface.
- Fixes typos.
- Adds `getNextMessageID` to the `ITeleporterMessenger` interface.
- Updates documentation comments of `SafeERC20TransferFrom`
- Removes unused external `checkIsAllowedRelayer` function.
- Uses `Math` library for `min` calculation in receipt array sizes

## How this was tested

Unit tests and CI.

## How is this documented

N/A


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,48 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // a receipt for. The messages are tracked per subnet and keyed by message ID.
-    // The first key is the subnet ID, the second key is the message ID, and the value is the info
+    // Key is the blockchain ID of the other subnet, and the value is a queue of pending receipts for messages
+    // received from that subnet.
+    mapping(bytes32 sourceChainID => ReceiptQueue.TeleporterMessageReceiptQueue receiptQueue)
+        public receiptQueues;
+
+    // Tracks the message hash and fee information for each message sent that has yet to be acknowledged
+    // with a receipt. The messages are tracked per subnet and keyed by message ID.
+    // The first key is the blockchain ID, the second key is the message ID, and the value is the info
```
minghinmatthewlam, line 58:
```suggestion
    // Tracks the message hash and fee information for each message sent that has yet to be confirmed delivery
    // with a receipt. The messages are tracked per chain and keyed by message ID.
    // The first key is the blockchain ID, the second key is the message ID, and the value is the info
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,48 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+    mapping(bytes32 sourceChainID => mapping(uint256 messageID => address relayerRewardAddress))
+        internal _relayerRewardAddresses;
 
-    // The blockchain ID of the chain the contract is deployed on. Initialized lazily when receiveCrossChainMessage() is called,
-    // if the value has not already been set.
-    bytes32 public blockchainID;
+    // Tracks the fee amounts for a given asset able to be redeemed by a given relayer.
+    // The first key is the relayer address, the second key is the fee token contract address,
+    // and the value is the amount of the asset redeemable by the relayer.
+    mapping(address relayerAddress => mapping(address feeTokenContract => uint256 redeemableRewardAmount))
```
minghinmatthewlam, line 80: can we rename this to `relayerRewardAddress` and update comments?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -146,8 +141,8 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             existingMessageInfo.feeInfo
         );
 
-        // Resubmit the message to the warp message precompile now that we know the exact message was
-        // already submitted in the past.
+        // Resubmit the message to the warp message precompile now that it is known that
```
minghinmatthewlam, line 144:
```suggestion
        // Resubmit the message to the warp message precompile now that it is confirmed that
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -287,7 +282,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         // Check the message has not been delivered before by checking that there is no relayer reward
         // address stored for it already.
         require(
-            relayerRewardAddresses[warpMessage.sourceChainID][
+            _relayerRewardAddresses[warpMessage.sourceChainID][
```
minghinmatthewlam, line 285: I think instead of checking `_relayerRewardAddresses` directly here, we should make an internal `_messageReceived` and use that check here to make it more clear.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -438,7 +429,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         for (uint256 i = 0; i < messageIDs.length; i++) {
             uint256 receivedMessageID = messageIDs[i];
             // Check the relayer reward address for this message.
```
minghinmatthewlam, line 431:
```suggestion
            // Check that the message has been received.
```

contracts/src/Teleporter/tests/CheckAllowedRelayerTests.t.sol
```diff
@@ -6,27 +6,16 @@
...
 
-contract CheckIsAllowedRelayerTest is TeleporterMessengerTest {
-    // The state of the contract gets reset before each
-    // test is run, with the `setUp()` function being called
-    // each time after deployment.
-    function setUp() public virtual override {
-        TeleporterMessengerTest.setUp();
-    }
-
+contract CheckIsAllowedRelayerTest is TeleporterMessenger, Test {
```
minghinmatthewlam, line 11: do we want to test `internal` functions of `TeleproterMessenger` by making derived contracts? I would lean towards `_checkIsAllowedRealyer` being tested in `TeleporterMessengerTests`

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -5,7 +5,7 @@
 
 pragma solidity 0.8.18;
 
-import "./TeleporterRegistry.sol";
+import {TeleporterRegistry} from "./TeleporterRegistry.sol";
 
 /**
  * @dev TeleporterUpgradeable provides upgrade utility for applications built on top
```
minghinmatthewlam, line 11: can't comment on line, but confirming for `minTeleporterVersion` we can't change visbility, since derived contracts will need to implement `updateMinTeleporterVersion` and have access to the state variable. Trying to think how we can only allow updating in `updateMinTeleporterVersion`

contracts/src/WarpProtocolRegistry.sol
```diff
@@ -37,8 +36,10 @@ abstract contract WarpProtocolRegistry {
     uint256 internal _latestVersion;
 
     // Mappings that keep track of the protocol version and corresponding contract address.
-    mapping(uint256 => address) internal _versionToAddress;
-    mapping(address => uint256) internal _addressToVersion;
+    mapping(uint256 version => address protocolAddress)
```
minghinmatthewlam, line 39: since `WarpProtcoolRegistry` is also abstract could consider changing visiblity for the state variables too, but we are looking to remove this contract in favor of `TeleporterRegistry` anyways

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -5,11 +5,19 @@
...
 struct TeleporterMessageReceipt {
     uint256 receivedMessageID;
     address relayerRewardAddress;
 }
 
+// Represents all of the information required for submitting a Teleporter message
+// to be sent to the given destination chain ID and address. Includes the fee
+// information for the message, the amount of gas the relayer must provide to execute
+// the message on the destination chain, the relayer accounts allowed to deliver the
+// message, and the message data itself.
```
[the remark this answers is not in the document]
minghinmatthewlam: I see you changed the `_sendTeleporterMessage` input to be `TeleporterMessageInput`, but thoughts on removing it as requirement for the interface, and moving to struct to `TeleporterMessenger`? Not a big deal either way imo

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -5,11 +5,19 @@
 
 pragma solidity 0.8.18;
 
+// A message receipt identifies the message ID that was
+// delivered and address that should be able to redeem the
+// reward for that message.
```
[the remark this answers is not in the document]
michaelkaplan13: Nice catch. I think singular "reward" is correct in this case, but fixed the other.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -5,18 +5,26 @@
 
 pragma solidity 0.8.18;
 
-import "./ITeleporterMessenger.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
```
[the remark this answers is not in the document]
michaelkaplan13: I'm not opposed to enabling the `no-global-imports` linting rule and correcting everywhere needed now. It will put us in the best position moving forward I think.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -533,7 +526,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     function getFeeInfo(
         bytes32 destinationChainID,
         uint256 messageID
-    ) external view returns (address feeAsset, uint256 feeAmount) {
+    ) external view returns (address, uint256) {
```
geoff-vball, line 529: Either leave these in or describe the return values in the comment above.

contracts/src/Teleporter/SafeERC20TransferFrom.sol
```diff
@@ -5,22 +5,29 @@
...
  * Note: A reentrancy guard must always be used when calling token.safeTransferFrom in order to
  * prevent against possible "before-after" pattern vulnerabilities.
+ *
+ * @custom:security-contact https://github.com/ava-labs/teleporter/blob/main/SECURITY.md
  */
 library SafeERC20TransferFrom {
     using SafeERC20 for IERC20;
 
+    /**
+     * @dev Checks the balance of the recipient before and after the call to transferFrom, and
```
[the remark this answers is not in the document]
michaelkaplan13: I'd rather keep it for completeness. Updating the wording now though 👍

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,48 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
     // Tracks the outstanding receipts to send back to a given subnet in subsequent messages sent to it.
-    // Key is the subnet ID of the other subnet, and the value is a queue of pending receipts for messages
-    // we have received from that subnet.
-    mapping(bytes32 => ReceiptQueue.TeleporterMessageReceiptQueue)
-        public outstandingReceipts;
-
-    // Tracks the message hash and fee information for each message sent that we have not yet received
-    // a receipt for. The messages are tracked per subnet and keyed by message ID.
-    // The first key is the subnet ID, the second key is the message ID, and the value is the info
+    // Key is the blockchain ID of the other subnet, and the value is a queue of pending receipts for messages
```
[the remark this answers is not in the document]
michaelkaplan13: I think "other chain" is clear here, as in "not this chain that the contract is deployed on, but the other one".

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I think "Get the relayer reward address for the message" is more clear. I'll update.

contracts/src/Teleporter/tests/CheckAllowedRelayerTests.t.sol
michaelkaplan13: What's the motivation for not wanting it to be in its own test suite? It makes it easier to find and isolate IMO.

I don't love making a derived contract to test `internal` functions, but I think it's the easy option that provides the needed coverage.

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -219,6 +235,11 @@ interface ITeleporterMessenger {
         uint256 messageID
     ) external view returns (address feeAsset, uint256 feeAmount);
 
+    /**
+     * @dev Returns the next message ID to be used to send a message to the given chain ID.
+     */
+    function getNextMessageID(bytes32 chainID) external view returns (uint256);
```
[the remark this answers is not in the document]
michaelkaplan13: You're right that it is referring to a blockchain ID, but throughout this file the rest the usages are also all `chainID`, so I'm going to leave this as is. Updating the naming was not mentioned in the audit, so I'd consider it out of scope for this PR.

contracts/src/CrossChainApplications/ExampleMessenger/ExampleCrossChainMessenger.sol
```diff
@@ -32,7 +32,7 @@ contract ExampleCrossChainMessenger is
         string message;
     }
 
-    mapping(bytes32 => Message) private _messages;
+    mapping(bytes32 originChainID => Message message) private _messages;
```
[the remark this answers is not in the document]
michaelkaplan13: Nope, we're matching an origin blockchain ID to the latest message received from that chain.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,48 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+    // received from that subnet.
+    mapping(bytes32 sourceChainID => ReceiptQueue.TeleporterMessageReceiptQueue receiptQueue)
+        public receiptQueues;
+
+    // Tracks the message hash and fee information for each message sent that has yet to be acknowledged
+    // with a receipt. The messages are tracked per subnet and keyed by message ID.
+    // The first key is the blockchain ID, the second key is the message ID, and the value is the info
     // for the uniquely identified message.
-    mapping(bytes32 => mapping(uint256 => SentMessageInfo))
+    mapping(bytes32 destinationChainID => mapping(uint256 messageID => SentMessageInfo messageInfo))
```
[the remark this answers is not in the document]
michaelkaplan13: Same as https://github.com/ava-labs/teleporter/pull/106#discussion_r1383647205

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,48 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    mapping(bytes32 => mapping(uint256 => address))
-        public relayerRewardAddresses;
-
     // Tracks the hash of messages that have been received but whose execution has never succeeded.
     // Enables retrying of failed messages with higher gas limits. Message execution is guaranteed to
-    // succeed at most once.  The first key is the subnet ID, the second key is the message ID, and
+    // succeed at most once. The first key is the blockchain ID, the second key is the message ID, and
     // the value is the hash of the uniquely identified message whose execution failed.
-    mapping(bytes32 => mapping(uint256 => bytes32))
+    mapping(bytes32 sourceChainID => mapping(uint256 messageID => bytes32 messageHash))
```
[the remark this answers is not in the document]
michaelkaplan13: Same as https://github.com/ava-labs/teleporter/pull/106#discussion_r1383647205

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -90,21 +90,16 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
      */
     function sendCrossChainMessage(
         TeleporterMessageInput calldata messageInput
-    ) external senderNonReentrant returns (uint256 messageID) {
+    ) external senderNonReentrant returns (uint256) {
```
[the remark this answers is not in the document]
michaelkaplan13: See audit report 👍

contracts/src/Teleporter/ReceiptQueue.sol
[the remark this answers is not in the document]
michaelkaplan13: Updated now. I think having different linting rules for different directories would likely cause more confusion that benefits.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: updated now

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: That wasn't an item called out in the audit, so I think it's out of scope for this PR.

michaelkaplan13: Gotcha, I personally am not super inclined to make the change given it increases overall contract size and gas costs and the check is properly described in the comment directly above here. If others feel strongly though I'm open to it.

minghinmatthewlam: When you say `gas costs` do you mean the increase in gas for deployment, or are you expecting an increase in gas for wrappign the check now in a function call?

Either way think the gas cost is minor in favor of more readable code so would be my vote.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Just went through and double checked that we've removed all named returned parameters to stay consistent, as recommend in the audit report. I also updated some of the "@return" comments in the interface definition. To avoid duplication I'll leave the comments here that refer to the documentation in the interface as is 👍

geoff-vball: Ahhh okay. Didn't realize they were in the interface. Thanks!

michaelkaplan13: They we're complete as is, so thanks for calling it out 👍

contracts/src/Teleporter/tests/CheckAllowedRelayerTests.t.sol
minghinmatthewlam: Good with the derived contract and own test suite, just want to make sure consistent with how other internal functions were tested.

geoff-vball (dismissed):
LGTM!

contracts/src/Teleporter/SafeERC20TransferFrom.sol
```diff
@@ -5,22 +5,28 @@
...
  * Note: A reentrancy guard must always be used when calling token.safeTransferFrom in order to
  * prevent against possible "before-after" pattern vulnerabilities.
+ *
+ * @custom:security-contact https://github.com/ava-labs/teleporter/blob/main/SECURITY.md
  */
 library SafeERC20TransferFrom {
     using SafeERC20 for IERC20;
 
+    /**
+     * @dev Checks the balance of the contract before and after the call to safeTransferFrom, and returns balance
```
cam-schultz, line 27:
```suggestion
     * @dev Checks the balance of the contract before and after the call to safeTransferFrom, and returns the balance
```

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
cam-schultz: Perhaps we should add a comment at that state variable explaining this special case and why it diverges from Solidity best practices.

scripts/abi_bindings.sh
```diff
@@ -54,6 +54,12 @@ forge build --force --extra-output-files abi
 
 contract_names=($CONTRACT_LIST)
 
+# Set the CGO flags to use the portable version of BLST
+#
+# We use "export" here instead of just setting a bash variable because we need
+# to pass this flag to all child processes spawned by the shell.
+export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
```
cam-schultz, line 61: Can we instead put this in a `constants.sh` and `source` that file in all our scripts? This will be consistent with our other repos, e.g. `awm-relayer`, `subnet-evm`, `avalanchego`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,47 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // Note that these values are also used to determine if a given message has been delivered or not.
-    // The first key is the subnet ID, the second key is the message ID, and the value is the reward address
-    // provided by the deliverer of the uniquely identified message.
-    mapping(bytes32 => mapping(uint256 => address))
-        public relayerRewardAddresses;
-
     // Tracks the hash of messages that have been received but whose execution has never succeeded.
     // Enables retrying of failed messages with higher gas limits. Message execution is guaranteed to
-    // succeed at most once.  The first key is the subnet ID, the second key is the message ID, and
+    // succeed at most once. The first key is the blockchain ID, the second key is the message ID, and
```
cam-schultz, line 64:
```suggestion
    // guaranteed to succeed at most once. The first key is the blockchain ID of the sending chain, the second key is the message ID, and
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,52 +37,47 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+
+    // Tracks the outstanding receipts to send back to a given chain in subsequent messages sent to that chain.
+    // Key is the blockchain ID of the other chain, and the value is a queue of pending receipts for messages
+    // received from that chain.
+    mapping(bytes32 sourceChainID => ReceiptQueue.TeleporterMessageReceiptQueue receiptQueue)
+        public receiptQueues;
+
+    // Tracks the message hash and fee information for each message sent that has yet to be acknowledged
+    // with a receipt. The messages are tracked per chain and keyed by message ID.
+    // The first key is the blockchain ID, the second key is the message ID, and the value is the info
```
cam-schultz, line 57:
```suggestion
    // The first key is the blockchain ID of the destination chain, the second key is the message ID, and the value is the info
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -664,33 +670,33 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     /**
      * @dev Marks the receipt of a message from the given destination chain ID with the given message ID.
      *
-     * It is possible that the receipt was already received for this message, in which case we return early.
-     * If existing message is found and not yet delivered, we delete it from state and increment the fee/reward
+     * If the receipt was already previously received for this message, returns early.
+     * If existing message is found and not yet delivered, deletes it from state and increment the fee/reward
```
cam-schultz, line 674:
```suggestion
     * If existing message is found and not yet delivered, delete it from state and increment the fee/reward
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -732,9 +738,8 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         }
 
         // Call the destination address of the message with the formatted call data.
-        // We only provide the required gas limit to the sub-call because we know that
-        // we have sufficient gas left to cover that amount, and do not want to allow the
-        // end application to consume arbitrary gas.
+        // Only provide the required gas limit to the sub-call so that the end application
+        // cannot consume an arbitrary amount gas.
```
cam-schultz, line 742:
```suggestion
        // cannot consume an arbitrary amount of gas.
```

cam-schultz (commented):
A handful of typographical corrections, otherwise LGTM

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -664,33 +670,33 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     /**
      * @dev Marks the receipt of a message from the given destination chain ID with the given message ID.
      *
-     * It is possible that the receipt was already received for this message, in which case we return early.
-     * If existing message is found and not yet delivered, we delete it from state and increment the fee/reward
+     * If the receipt was already previously received for this message, returns early.
```
bernard-avalabs, line 673: nit: sentence may read better if we flip it around: "Function returns early if the receipt..."

bernard-avalabs (commented):
Generally makes sense to me. I've left a few minor comments mostly about wording in comments.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Fair enough, yeah I checked the `forge test --gas-report` before and after making those change, and it was a very minor increase. 

I added an internal `_messageReceived` now.

contracts/src/Teleporter/tests/CheckAllowedRelayerTests.t.sol
michaelkaplan13: This was actually the first internal method in the `TeleporterMessenger` contract. 👍 

They theoretically could be tested through various inputs to the public/external methods that call them, but in this case I think I'd rather keep the isolated test coverage for it specifically.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
michaelkaplan13: Good call. Adding now

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I think "deletes" is correct in this case. "If the existing message is found and not yet delivered, this function deletes it from state" (consistent with "returns early" above). I'm updating the wording otherwise thought because reading it again now I think it's confusing as is.

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -56,16 +64,11 @@ library ReceiptQueue {
     ) internal returns (TeleporterMessageReceipt[] memory result) {
         // Get the current outstanding receipts for the given chain ID.
         // If the queue contract doesn't exist, there are no outstanding receipts to send.
-        uint256 resultSize = size(queue);
+        uint256 resultSize = Math.min(_MAXIMUM_RECEIPT_COUNT, size(queue));
```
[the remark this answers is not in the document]
michaelkaplan13: It should be the same gas usage roughly, but using the library is recommend as the less-error prone approach 👍

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -162,9 +175,9 @@ interface ITeleporterMessenger {
     /**
      * @dev Sends the receipts for the given `messageIDs`.
      *
-     * Sends the receipts of the specified messages in a new message (with an empty payload) back to the origin chain.
-     * This is intended to be used if the message receipts were originally included in messages that were dropped
-     * or otherwise not delivered in a timely manner.
+     * Sends the specified message receipts in a new message (with an empty payload) back to the origin chain.
+     * This is intended to be used if the message receipts are not being delivered in a timely manner by the
```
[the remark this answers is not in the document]
michaelkaplan13: Thanks, I agree your suggestion is much better phrasing 🙏

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -56,16 +64,11 @@ library ReceiptQueue {
     ) internal returns (TeleporterMessageReceipt[] memory result) {
         // Get the current outstanding receipts for the given chain ID.
         // If the queue contract doesn't exist, there are no outstanding receipts to send.
```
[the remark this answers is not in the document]
michaelkaplan13: Good call. I don't think the above comment applies any more either since technically the queue exists for every chain ID in the mapping (with the default value if it has never been set previously). I'll update it.

contracts/src/CrossChainApplications/README.md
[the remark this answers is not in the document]
michaelkaplan13: Good catch. We may want to refresh that guide to reflect all of the recent changes, but not for this PR.

scripts/abi_bindings.sh
michaelkaplan13: Yep, updated now. I updated `abi_binding.sh` and `e2e_test.sh` to first set `TELEPORTER_PATH` and then source the constants file from there, and removed unused constants.

scripts/constants.sh
```diff
@@ -2,18 +2,11 @@
 # Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 # See the file LICENSE for licensing terms.
 
-# Use lower_case variables in the scripts and UPPER_CASE variables for override
-# Use the constants.sh for env overrides
-
-TELEPORTER_PATH=$(
```
cam-schultz, line 8: What's the reason for removing this and putting it in the other scripts? My thinking was that we could put this here and source constants.sh everywhere that needs it.

scripts/local/e2e_test.sh
```diff
@@ -70,6 +69,12 @@ fi
 
 cd $cwd
 
+# Set the CGO flags to use the portable version of BLST
+#
+# We use "export" here instead of just setting a bash variable because we need
+# to pass this flag to all child processes spawned by the shell.
+export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
```
cam-schultz, line 76: Does this still need to be here now that it's included in constants.sh?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -33,46 +37,47 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // Note that these values are also used to determine if a given message has been delivered or not.
-    // The first key is the chain ID, the second key is the message ID, and the value is the reward address
-    // provided by the deliverer of the uniquely identified message.
-    mapping(bytes32 => mapping(uint256 => address))
-        public relayerRewardAddresses;
-
     // Tracks the hash of messages that have been received but whose execution has never succeeded.
     // Enables retrying of failed messages with higher gas limits. Message execution is guaranteed to
-    // succeed at most once.  The first key is the chain ID, the second key is the message ID, and
+    // succeed at most once. The first key is the blockchain ID, the second key is the message ID, and
```
cam-schultz, line 64:
```suggestion
    // succeed at most once. The first key is the blockchain ID of the sending chain, the second key is the message ID, and
```

cam-schultz (commented):
A couple of small comments, including an amendment to a suggestion I previously made.

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: The first time I suggested a change on this line, I had inadvertently suggested a change that was already included.

scripts/local/e2e_test.sh
michaelkaplan13: Nope, thanks for catching.

scripts/constants.sh
michaelkaplan13: That was my originally thinking also, but then I realized having `TELEPORTER_PATH` set in each individual script is consistent with the pattern used in the `subnet-evm` scripts.

The reason is that sourcing the `constants.sh` file from a script requires knowing where the script is in relation to to the current working directory the script is being called from. So it's less error prone to set the relative path to the base of the repo from each script individual (no matter where it's called from), and then source `constants.sh` relative from there.

cam-schultz: Pesky chickens and eggs 😆 Makes sense

cam-schultz (dismissed):
LGTM!

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ on:
       - "*"
 
 env:
-  RELAYER_VERSION: v0.2.2
+  RELAYER_VERSION: v0.2.3
```
minghinmatthewlam, line 12: can we also update the other references to relayer version `v0.2.2.` in docker run/test

michaelkaplan13: Good catch 👍

minghinmatthewlam (dismissed):
LGTM 👍

cam-schultz (approved):

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -577,108 +595,102 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         // received for this message (it's possible for receipts to be sent more than once)
-        // or that the other chain sent an invalid receipt. We return early since this is an expected
+        // or that the other chain sent an invalid receipt. Return early since this is an expected
         // case where there is no fee to be paid for the given message.
         if (messageInfo.messageHash == bytes32(0)) {
             return;
         }
 
-        // Delete the message information from state now that we know it has been delivered.
+        // Delete the message information from state now that it is known to be delivered.
```
bernard-avalabs, line 688: nit: "it is known to be" seems a bit awkward. Can we just say "... now that the message has been delivered"?

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -17,6 +17,14 @@ import "./TeleporterRegistry.sol";
  */
 abstract contract TeleporterUpgradeable {
     TeleporterRegistry public immutable teleporterRegistry;
+
+    /**
+     * @dev The minimum required Teleporter version that the contract is allowed
+     * to receive messages from. Should only be update through the `updateMinTeleporterVersion`
```
bernard-avalabs, line 23: nit: update -> updated

bernard-avalabs (approved):
LGTM

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
michaelkaplan13: fixed 👍

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I'd prefer to leave it as because technically the fact that the message is delivered isn't enough to delete it from state here. The sending chain has to be informed of the delivery via a receipt.

cam-schultz (approved):

minghinmatthewlam (approved):

bernard-avalabs (approved):

PR #113, opened 2023-11-06
Test send specific receipts (Part 2)

gwen917:
## Why this should be merged
- Relayer manually retries sending of specific receipts.

## How this works
See above

## How this was tested
New E2E test cases

## How is this documented


tests/send_specific_receipts.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/ava-labs/teleporter/tests/utils"
+	"github.com/ethereum/go-ethereum/common"
+	. "github.com/onsi/gomega"
+)
+
+func SendSpecificReceiptsGinkgo() {
+	sendSpecificReceipts(&network.LocalNetwork{})
+}
+
+func sendSpecificReceipts(network network.Network) {
```
cam-schultz, line 19: Let's export this function

tests/send_specific_receipts.go
```diff
@@ -0,0 +1,99 @@
...
+	// Check delivered
+	delivered, err = subnetATeleporterMessenger.MessageReceived(&bind.CallOpts{}, subnetBInfo.BlockchainID, messageID)
+	Expect(err).Should(BeNil())
+	Expect(delivered).Should(BeTrue())
+
+	// Check the reward amount. The reward amount should be 10
+	amount, err := subnetATeleporterMessenger.CheckRelayerRewardAmount(&bind.CallOpts{}, fundedAddress, mockTokenAddress)
+	Expect(err).Should(BeNil())
+	Expect(amount).Should(Equal(big.NewInt(10)))
+
```
cam-schultz, line 98: Can you add an additional step of sending an arbitrary message from Subnet B -> Subnet A? This message will also include the receipts that were explicitly sent in the above steps. We should check that the relayer reward amount is unchanged, even after the receipt is delivered again.

gwen917: sure. Then, will it cover `Test receiving a receipt multiple times has no effect` case?

tests/send_specific_receipts.go
gwen917: fixed

tests/send_specific_receipts.go
cam-schultz: Yup it will! I pointed it out for completeness here, but that will also cover that test case.

cam-schultz (approved):

PR #115, opened 2023-11-06
Set blst to portable mode for abi check

geoff-vball:
## Why this should be merged
Our tests involving `subnet-evm` are flakily failing with `Caught SIGILL in blst_cgo_init, consult <blst>/bindings/go/README.md.` every once in a while. I think it's an issue with cross-platform compatibility with `blst` - I would imagine some of our runners are on the right OS, but some might be on an incompatible version. 

This *should* fix the situation, but it's hard to tell with flaky tests.

Also moves `constants.sh` from `scripts/local` to `scripts`, because it is more general than just local tests.

## How is this documented


geoff-vball:
Closing in favor of this being fixed in https://github.com/ava-labs/teleporter/pull/106

PR #119, opened 2023-11-08
Use assembly for low level call

### LINKED ISSUE #107: Consider using assembly for low level call to avoid return bombs
**Context and scope**
When a message is delivered, an external call is made to the `destinationAddress` specified by the message sender where the amount of gas the call can use is specified in the message by the `requiredGasLimit` value. The return data from this call is not used and only the `success` of the call is recorded. Although a gas limit is set for the call, the external contract could force the transaction to use gas in excess of the specified limit due to an [open issue with the Solidity compiler](https://github.com/ethereum/solidity/issues/12306). This is possible because the `TeleporterMessenger` contract pays for memory expansion to accommodate the return data, but the external contract controls the size of that return data. Consider using assembly to explicitly prevent any memory from being allocated for the return data from the external call.

See reference [here](https://github.com/nomad-xyz/ExcessivelySafeCall/tree/main).

For instance, the change could look like:
```
        // Call the destination address of the message with the formatted call data.
        // Only provide the required gas limit to the sub-call so that the end application
        // cannot consume an arbitrary amount gas.
        bytes memory payload = abi.encodeCall(
            ITeleporterReceiver.receiveTeleporterMessage,
            (originChainID, message.senderAddress, message.message)
        );
        address target = message.destinationAddress;
        uint256 requiredGas = message.requiredGasLimit;
        bool success;
        assembly ("memory-safe") {
            success := call(
               requiredGas, // gas provided to the call
                target, // call target
                0, // zero value
                add(payload, 0x20), // input data
                mload(payload), // input data length
                0, // output
                0 // output size
            )
        }

        // If the execution failed, store a hash of the message in state such that its
        // execution can be retried again in the future with a higher gas limit (paid by whoever
        // retries).Either way, the message will now be considered "delivered" since the relayer
        // provided enough gas to meet the required gas limit.
        if (!success) {
            _storeFailedMessageExecution(originChainID, message);
            return;
        }
```

**Open questions**
Does `call` in assembly return false if the contract being call hits a `revert` statement?

michaelkaplan13:
## Why this should be merged

Fixes #107 

## How this works

See description [here](https://github.com/ava-labs/teleporter/issues/107). TL;DR, uses assembly to avoid memory expansion of unused return data.

## How this was tested

Unit tests and CI.

## How is this documented

Comments in code (no interface or behavior changes).


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -377,19 +377,37 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         // message can be retried again if desired.
-        // solhint-disable-next-line avoid-low-level-calls
-        (bool success, ) = message.destinationAddress.call(
-            abi.encodeCall(
-                ITeleporterReceiver.receiveTeleporterMessage,
-                (originChainID, message.senderAddress, message.message)
+        //
+        // Use assembly to make the low-level call to avoid unnecessary memory expansion of the return data
+        // due to an issue with the version of Solidity used. This prevents possible "return bomb" vectors
+        // where the external contract could for the caller to use an arbitrary amount of gas.
```
cam-schultz, line 396: The wording here seems a bit off.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -377,19 +377,37 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        address target = message.destinationAddress;
+        bool success;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            success := call(
+                gas(), // gas provided to the call
+                target, // call target
+                0, // zero value
+                add(payload, 0x20), // input data
+                mload(payload), // input data length
```
cam-schultz, line 406: From the Yul [documentation](https://docs.soliditylang.org/en/latest/yul.html#evm-dialect), these arguments correspond to the input and input size. Can you explain how the `add` and `mload` operations correspond to this? I'd expect to be something like `mload(payload)` to get the input pointer, then `msize(payload)` to get the payload size.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -377,19 +377,37 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-                (originChainID, message.senderAddress, message.message)
+        //
+        // Use assembly to make the low-level call to avoid unnecessary memory expansion of the return data
+        // due to an issue with the version of Solidity used. This prevents possible "return bomb" vectors
+        // where the external contract could for the caller to use an arbitrary amount of gas.
+        address target = message.destinationAddress;
+        bool success;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            success := call(
```
cam-schultz, line 401: Does Solidity variable scoping assign the result of this call to the `success` variable in the outer scope? I'm thrown off by the similarity to Golang syntax, which would not update the outer scope variable using `:=` syntax.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -731,19 +749,35 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+            ITeleporterReceiver.receiveTeleporterMessage,
+            (originChainID, message.senderAddress, message.message)
         );
 
+        // Call the destination address of the message with the formatted call data. Only provide the required
+        // gas limit to the sub-call so that the end application cannot consume an arbitrary amount of gas.
+        //
+        // Use assembly to make the low-level call to avoid unnecessary memory expansion of the return data
+        // due to an issue with the version of Solidity used. This prevents possible "return bomb" vectors
+        // where the external contract could for the caller to use an arbitrary amount of gas.
```
cam-schultz, line 764: Same awkward wording here.

cam-schultz (commented):
I have a few comments and questions, mostly around unfamiliarity with Solidity language semantics, especially w.r.t. assembly blocks.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Yep, that's correct. This is based on the example provided here: https://gist.github.com/steventhornton/ac2adf96aeb2f2ee29fc225d940eaa9a

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -377,19 +377,37 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-        // solhint-disable-next-line avoid-low-level-calls
-        (bool success, ) = message.destinationAddress.call(
-            abi.encodeCall(
-                ITeleporterReceiver.receiveTeleporterMessage,
-                (originChainID, message.senderAddress, message.message)
+        //
+        // Use assembly to make the low-level call to avoid unnecessary memory expansion of the return data
+        // due to an issue with the version of Solidity used. This prevents possible "return bomb" vectors
+        // where the external contract could for the caller to use an arbitrary amount of gas.
+        address target = message.destinationAddress;
```
minghinmatthewlam, line 397: nit: don't think we need additional variable here and below

michaelkaplan13: They are actually needed to make them easily available within the `assembly` block. I'll add a comment explaining this.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Updated 👍 Let me know if you think it's clearer now or can think of a better wording.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: To be honest, I pulled this from a handful of examples I found online and from OpenZeppelin. Dug into it a bit more now, and added comments on the usage of `add` and `mload`. Thanks for calling out.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Updated

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: Thanks, yeah those comments clarify things. Do you know if the array length is always guaranteed to be in the first slot of the array? How about for an empty array?

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: I'm not sure I understand this either. Could we just refer to `message.destinationAddress` directly in the below `call()`?

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: From the [Solidity docs](https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html#bytes-and-string):

"""
bytes and string are encoded identically. In general, the encoding is similar to bytes1[], in the sense that there is a slot for the array itself and a data area that is computed using a keccak256 hash of that slot’s position. However, for short values (shorter than 32 bytes) the array elements are stored together with the length in the same slot.

In particular: if the data is at most 31 bytes long, the elements are stored in the higher-order bytes (left aligned) and the lowest-order byte stores the value length * 2. For byte arrays that store data which is 32 or more bytes long, the main slot p stores length * 2 + 1 and the data is stored as usual in keccak256(p). This means that you can distinguish a short array from a long array by checking if the lowest bit is set: short (not set) and long (set).
"""

From this, it seems like we need to handle small and large payloads differently: small payloads encode the length and payload bytes in the same slot, where large payloads put the payload bytes at the keccack hash of the length pointer.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Nope, the relevant documentation about variable access in assembly is here: https://docs.soliditylang.org/en/latest/assembly.html#access-to-external-variables-functions-and-libraries

The TL;DR is:
```
Local variables of value type are directly usable in inline assembly. They can both be read and assigned to.

Local variables that refer to memory evaluate to the address of the variable in memory, not the value itself.
```
We could access that memory directly using other assembly instructions, but it ends up being much easier to make those values available by assigning them to local value types prior to the assembly.

cam-schultz: Makes sense, thanks for the docs link

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I think the docs you link refer to the layout of variables in storage, not in memory. I'm checking to see if I can find the correct documentation for encoding of arrays in memory.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -377,19 +377,38 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        // Re-encode the payload by ABI encoding a call to the {receiveTeleporterMessage} function
+        // defined by the {ITeleporterReceiver} interface.
+        bytes memory payload = abi.encodeCall(
+            ITeleporterReceiver.receiveTeleporterMessage,
+            (originChainID, message.senderAddress, message.message)
+        );
+
         // Reattempt the message execution with all of the gas left available for execution of this transaction.
         // Use all of the gas left because this message has already been successfully delivered, and it is the
         // responsibility of the caller to provide as much gas is needed. Compared to the initial delivery, where
```
minghinmatthewlam, line 389:
```suggestion
        // Use all of the gas left because this message has already been successfully delivered, and it is the
        // caller's responsibility to provide as much gas as is needed. Compared to the initial delivery, where
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -377,19 +377,38 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         // the relayer should still receive their reward even if the message execution takes more gas than expected.
         // Require that the call be successful such that in the failure case this transaction reverts and the
         // message can be retried again if desired.
-        // solhint-disable-next-line avoid-low-level-calls
-        (bool success, ) = message.destinationAddress.call(
-            abi.encodeCall(
-                ITeleporterReceiver.receiveTeleporterMessage,
-                (originChainID, message.senderAddress, message.message)
+        //
+        // Assembly is used for the low-level call to avoid unnecessary memory expansion of the return data.
```
minghinmatthewlam, line 394: should we attach the github issue here as a link?

michaelkaplan13: I thought about it, but thought that the link is more likely to break at some point than not, so stayed away from putting in the comment of the contract.

michaelkaplan13: I'm happy to add if people think otherwise though

minghinmatthewlam: i'd still prefer to add

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -731,19 +750,36 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        // This prevents possible "return bomb" vectors where the external contract could force the caller
+        // to use an arbitrary amount of gas. First need to define the {target}, {requiredGas}, and {success}
+        // variables to make them easily accesible within and follow the assembly block.
+        address target = message.destinationAddress;
+        uint256 requiredGas = message.requiredGasLimit;
+        bool success;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
```
minghinmatthewlam, line 771: the two assembly calls are pretty similar, any chance we can abstract to a function and use for both? 

Also, any unit tests for the assembly call that can be added?

michaelkaplan13: Good call on moving the assembly to a helper 👍 

Our unit tests actually already cover the assembly usage. The expected behavior is unchanged, so the tests aren't altered. I'm not 100% sure if it's possible to test the "return bomb" case given that `receiveTeleporterMessage` doesn't have any return value

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: included

minghinmatthewlam (approved):
LGTM

cam-schultz (approved):
LGTM

PR #121, opened 2023-11-09
Test relay altered message

gwen917:
## Why this should be merged
- Try resubmitting a message that has been slightly altered, ensure it doesn't work.
- Relayer modifies a message in flight. This should fail warp verification

## How this works
See above

## How this was tested
New E2E test cases

## How is this documented


tests/relayer_modifies_message.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/ava-labs/teleporter/tests/utils"
+	"github.com/ethereum/go-ethereum/common"
+	. "github.com/onsi/gomega"
+)
+
+func RalyerModifiesMessageGinkgo() {
+	RalyerModifiesMessage(&network.LocalNetwork{})
+}
+
+func RalyerModifiesMessage(network network.Network) {
```
cam-schultz, line 19:
```suggestion
func RelayerModifiesMessageGinkgo() {
	RelayerModifiesMessage(&network.LocalNetwork{})
}

func RelayerModifiesMessage(network network.Network) {
```

cam-schultz (approved):
One small typo, otherwise LGTM

tests/relayer_modifies_message.go
```diff
@@ -0,0 +1,57 @@
...
+
+	receipt, messageID := utils.SendCrossChainMessageAndWaitForAcceptance(
+		ctx, subnetAInfo, subnetBInfo, sendCrossChainMessageInput, fundedAddress, fundedKey, subnetATeleporterMessenger)
+
+	// Relay the message to the destination
+	// Relayer modifies the message in flight
+	network.RelayMessage(ctx, receipt, subnetAInfo, subnetBInfo, true, false)
+
+	// Check Teleporter message was not received on the destination
+	delivered, err := subnetBTeleporterMessenger.MessageReceived(&bind.CallOpts{}, subnetAInfo.BlockchainID, messageID)
```
bernard-avalabs, line 54: Just for my own understanding: Am I correct to assume that MessageReceived is provided by AWM and does not rely on actions from a (potentially malicious) relayer?

tests/utils/utils.go
```diff
@@ -327,6 +328,17 @@ func CreateReceiveCrossChainMessageTransaction(
...
 
+	if alterMessage {
+		// Alter the message
+		modifiedMessage := make([]byte, len(signedMessage.Payload))
+		copy(modifiedMessage, signedMessage.Payload)
+		modifiedMessage[0] = ^modifiedMessage[0]
+		Expect(modifiedMessage[:]).ShouldNot(Equal(signedMessage.Payload[:]))
+		signedMessage.Payload = modifiedMessage
+
+		signedMessage.Initialize()
```
bernard-avalabs, line 339: Should we consider a related test where the signature is from the original payload? Perhaps testing incorrect signatures is already covered in a different test case.

bernard-avalabs (approved):
LGTM. Just some questions for my own understanding.

tests/utils/local_network_utils.go
```diff
@@ -126,6 +126,12 @@ func RelayMessage(
 		alterMessage,
 	)
 
+	if alterMessage {
```
cam-schultz, line 129: We should be able to instead pass `expectSuccess || alterMessage` to the call to `SendTransactionAndWaitForAcceptance` below

PR #126, opened 2023-11-10
Update TeleporterMessenger.sol

Aryan9592:
## Why this should be merged
This PR aims to enhance some 'for' loops by making them more gas efficient. Also got some typos fixed!

## How this works
For reference -> https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement

Thanks @michaelkaplan13


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -583,9 +590,12 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         }
 
         // Otherwise, the deliverer address must be included in allowedRelayers.
-        for (uint256 i = 0; i < allowedRelayers.length; ++i) {
+        for (uint256 i; i < allowedRelayers.length;) {
             if (allowedRelayers[i] == delivererAddress) {
                 return true;
+                unchecked {
```
bernard-avalabs, line 596: I assume that this unchecked increment should be outside of the if block.

Aryan9592: @bernard-avalabs  You are right about it..Thanks!

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -269,7 +269,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
             "TeleporterMessenger: invalid destination chain ID"
         );
 
-        // Check the message has not been delivered previously.
+        // Check if the message has not been delivered previously.
```
geoff-vball, line 272:
```suggestion
        // Require that the message has not been delivered previously.
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -303,14 +303,17 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-        for (uint256 i = 0; i < length; ++i) {
+        for (uint256 i; i < length;) {
             TeleporterMessageReceipt memory receipt = teleporterMessage
                 .receipts[i];
             _markReceipt(
                 warpMessage.sourceChainID,
                 receipt.receivedMessageID,
                 receipt.relayerRewardAddress
             );
+            unchecked {
```
minghinmatthewlam, line 314: can you add a comment about why we're using unchecked increment here, same for below loop. Thinking if the minimal gas optimization is worth the readability.

Aryan9592: Sure!

Aryan9592: As you say @minghinmatthewlam

michaelkaplan13 (approved):
🙏 Thanks for catching all of these and making the PR. LGTM

geoff-vball (approved):

michaelkaplan13:
I think the Snyk action is failing just because the forked branch isn't able to use the Snyk account key set for the repository. It should pass when merged now.

minghinmatthewlam:
Thanks @Aryan9592 for the fixes!

Aryan9592:
Welcome @minghinmatthewlam

PR #130, opened 2023-11-13
Add utility function for parsing generic logs

geoff-vball:
## Why this should be merged
Eliminates code duplication. Makes it very easy to parse logs from contracts.

## How this was tested
Existing E2E tests that use this

## How is this documented


cam-schultz (approved):
This is great 👍 There are a few more log helpers in active branches that can be removed as well once this is integrated.

michaelkaplan13 (approved):
🙌

PR #133, opened 2023-11-14
Test retry send message

cam-schultz:
## Why this should be merged
Implements the following test cases from https://github.com/ava-labs/teleporter/issues/94:
- Submit a message, large validator set turn over before the message is delivered. Message can be resubmitted on the sending chain to be retried

While implementing this test, I discovered a bug in `avalanche-network-runner` that is fixed in `main`, but not in a release version. Therefore, this PR shouldn't be merged until a new version of `avalanche-network-runner` is released, and that version included in a new `subnet-evm` version.

## How this works
The test works as follows:
- Send a Teleporter message and produce an aggregate signature
- Add new validators to the signing subnet such that the aggregate signature no longer represents enough stake weight
- Attempt to deliver the message (this fails)
- Call `retrySendCrossChainMessage` to trigger producing a new aggregate signature using the new validator set

## How this was tested
CI

## How is this documented
N/A


tests/utils/local_network_setup.go
```diff
@@ -163,55 +169,110 @@ func SetupNetwork(warpGenesisFile string) {
...
-	blockchainIDA = subnetADetails.BlockchainID
-	chainANodeURIs = append(chainANodeURIs, subnetADetails.ValidatorURIs...)
+	SetSubnetAValues(subnetA)
 
 	subnetB = subnetIDs[1]
-	subnetBDetails, ok := manager.GetSubnet(subnetB)
+	SetSubnetBValues(subnetB)
+}
+
+func SetSubnetBValues(subnetID ids.ID) {
```
gwen917, line 204: Is it possible to use the same function to set the subnet values? `SetSubnetBValues` and `SetSubnetAValues` are similar.

gwen917 (commented):
LGTM. What is the bug in the avalanche-network-runner?

tests/utils/local_network_setup.go
geoff-vball: We can totally clean this all up into a loop. I don't mind making a PR to general-e2e after this is merged in.

cam-schultz: Agreed. Let's save general cleanup work for https://github.com/ava-labs/teleporter/pull/93

cam-schultz:
> LGTM. What is the bug in the avalanche-network-runner?

It was an issue with removing subnet validators from permissioned subnets. The bug fix is here: https://github.com/ava-labs/avalanche-network-runner/pull/671

Additionally, there is no function to add validators to a permissioned subnet in the most recent release. That feature was added here: https://github.com/ava-labs/avalanche-network-runner/pull/673

Once these changes are included in a release version, we can shift the dependency to the release rather than a commit.

tests/utils/local_network_setup.go
cam-schultz: Addressed in #93

PR #137, opened 2023-11-15
Teleporter CLI

### LINKED ISSUE #96: Teleporter Decoder
**Context and scope**
We want to provide decoding functionalities that will help make debugging Teleporter transactions easier.

- [x] Given a transaction to Teleporter address, there should be function selector, and data. Being able to decoding the selector to which Teleporter function, and decode data to function arguments
- [x] Decode Teleporter emitted log events to which event was emitted, and decode into event fields (topics and data)
- [x] Given a Teleporter message's bytes decode into Teleporter message with corresponding fields

**Open questions**
GIven a dapp contract's abi, can we decode the Teleporter message payload into the call arguments, i.e. figure out which function in the dapp is being called, and what function arguments are passed.

minghinmatthewlam:
## Why this should be merged
Fixes #96 

## How this works
Creates a cobra command cli that integrates with the Teleporter protocol. The command supports three initial subcommands:
- `event`: Passing in `--topics` and `--data`, parses into the corresponding Teleporter event, does not require Teleporter contract address or rpc url
- `message`: Passing in Teleporter message bytes, parses into corresponding `TeleporterMessage`, does not require Teleporter contract or rpc url
- `transaction`: Passing in a transaction hash, Teleporter contract address, and rpc url, filters through the transaction's receipt for any Teleporter or Warp log event.

Since the command uses the generated abi bindings, it must be used with data that was generated with same abi, i.e. same version of Teleporter.

## How this was tested
Copied and adapted cli to previous commit deployed on firedrillt/secondfire subnets, and tested with transactions on those chains.

## How is this documented
README
comments


cli/README.md
```diff
@@ -0,0 +1,17 @@
...
+To build the CLI, run `go build -o teleporter-cli` from this directory. This will create a binary called `teleporter-cli` in the current directory.
+
+## Usage
+
+The CLI has a number of subcommands. To see the list of subcommands, run `./teleporter-cli help`. To see the help for a specific subcommand, run `./teleporter-cli help <subcommand>`.
+
+The supported subcommands include:
+
+- `event`: given a log event's topics and data, attempts to decode into a Teleporter event in more readable format.
+- `message`: given a Teleporter message bytes hex encoded, attempts to decode into a Teleporter message in more readable format.
```
cam-schultz, line 16:
```suggestion
- `message`: given a Teleporter message encoded as a hex string, attempts to decode into a Teleporter message in a more readable format.
```

cli/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+To build the CLI, run `go build -o teleporter-cli` from this directory. This will create a binary called `teleporter-cli` in the current directory.
+
+## Usage
+
+The CLI has a number of subcommands. To see the list of subcommands, run `./teleporter-cli help`. To see the help for a specific subcommand, run `./teleporter-cli help <subcommand>`.
+
+The supported subcommands include:
+
+- `event`: given a log event's topics and data, attempts to decode into a Teleporter event in more readable format.
```
cam-schultz, line 15:
```suggestion
- `event`: given a log event's topics and data, attempts to decode into a Teleporter event in a more readable format.
```

cli/README.md
```diff
@@ -0,0 +1,17 @@
...
+- `message`: given a Teleporter message bytes hex encoded, attempts to decode into a Teleporter message in more readable format.
+- `transaction`: given a transaction hash, attempts to decode all relevant Teleporter and Warp log events in more readable format.
```
cam-schultz, line 17:
```suggestion
- `transaction`: given a transaction hash, attempts to decode all relevant Teleporter and Warp log events in a more readable format.
```

cli/cmd/event.go
```diff
@@ -0,0 +1,47 @@
...
+	"go.uber.org/zap"
+)
+
+var (
+	topicArgs []string
+	data      []byte
+)
+
+var eventCmd = &cobra.Command{
+	Use:   "event --topics topic1, topic2",
```
cam-schultz, line 19: From the usage string, it's not clear how to provide both topics and data. Is it
`--topics topic1 topic2 -- --data data`? Or
`--topics "topic1,topic2" --data data`? Or something else?

cli/cmd/transaction.go
```diff
@@ -0,0 +1,94 @@
...
+	"go.uber.org/zap"
+)
+
+const (
+	warpPrecompileAddress = "0x0200000000000000000000000000000000000005"
+)
+
+var (
+	rpcEndpoint     string
+	contractAddress common.Address
```
cam-schultz, line 24: Can you include "Teleporter" in this var and the corresponding flag?

cam-schultz (commented):
Generally looks good, this will be a great tool to have! I have two high level requests:

1) Can you add unit test suites for the CLI commands? This will also serve as documentation of how to call the CLI commands for easy testing.

2) Can you modify the `golang-ci lint` job to lint the `cli/` directory as well? We should still skip `abi-bindings/`.

geoff-vball (commented):
Forgot to submit these too.. So sorry 😓

cli/cmd/transaction.go
```diff
@@ -0,0 +1,94 @@
...
+
+func init() {
+	rootCmd.AddCommand(transactionCmd)
+	transactionCmd.PersistentFlags().StringVar(&rpcEndpoint, "rpc", "", "RPC endpoint to connect to the node")
+	address := transactionCmd.PersistentFlags().StringP("contract-address", "c", "", "Teleporter contract address")
+	err := transactionCmd.MarkPersistentFlagRequired("rpc")
+	cobra.CheckErr(err)
+	err = transactionCmd.MarkPersistentFlagRequired("contract-address")
+	cobra.CheckErr(err)
+	transactionCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
minghinmatthewlam: I extracted the pre run funcs into separate functions but left as a call in init, since it's easier with the args reading, otherwise would need other global var for the args.

cli/cmd/transaction.go
```diff
@@ -0,0 +1,94 @@
...
+		// Run the persistent pre-run function of the root command if it exists.
+		if rootCmd.PersistentPreRunE != nil {
+			if err := rootCmd.PersistentPreRunE(cmd, args); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
minghinmatthewlam: children only inherit if they do not declare their own persistent pre runs. There was more obscure doc on it, and also tested it out.

cli/README.md
```diff
@@ -0,0 +1,17 @@
+# Teleporter CLI
+
+This directory contains the source code for the Teleporter CLI. The CLI is a command line interface for interacting with the Teleporter contracts. It is written with [cobra](https://github.com/spf13/cobra) commands as a Go application.
+
+## Build
+
+To build the CLI, run `go build -o teleporter-cli` from this directory. This will create a binary called `teleporter-cli` in the current directory.
```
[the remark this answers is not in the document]
minghinmatthewlam: restructured to be `cmd/teleporter-cli`

cli/cmd/transaction.go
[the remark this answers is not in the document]
geoff-vball: Ah I see. In that case, let's at least call `cmd.Parent()` here instead of using `rootCmd` directly, so if we ever change the command structure, we won't have to refactor that part.

[This page](https://stackoverflow.com/questions/64055288/cobra-persistentprerun-stack-overflow) says `cmd.Parent().PersistentPreRun(cmd.Parent(), args)` might be more concise.

cli/cmd/root.go
```diff
@@ -0,0 +1,64 @@
...
+to debug Teleporter on chain activity. The CLI can help decode
+Teleporter and Warp events, as well as parsing Teleporter messages.`,
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
```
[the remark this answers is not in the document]
geoff-vball: Right now, if we hit this line we swallow the error and never print it.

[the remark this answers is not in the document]
minghinmatthewlam: did you mean with the previous `os.Exit(1)` commit? It's updated to use `panic` which should include error in exit message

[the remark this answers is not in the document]
geoff-vball: Yeah, sorry, not sure why I didn't see why this was changed. Thanks!

geoff-vball (approved):
Looks great! Only one small thing you could consider - I think most cli implementations I've seen add the child to the parent in the _parent's_ init function. This way you have all the children in one place, and you can follow the tree structure down if you want.

michaelkaplan13 (approved):
Generally LGTM. I wasn't previously familiar with cobra so not sure of all the best practices as related to it, but seems like the perfect use case for it, and playing around with the CLI feels very intuitive 👍

minghinmatthewlam:
> Generally looks good, this will be a great tool to have! I have two high level requests:
> 
> 1. Can you add unit test suites for the CLI commands? This will also serve as documentation of how to call the CLI commands for easy testing.
> 2. Can you modify the `golang-ci lint` job to lint the `cli/` directory as well? We should still skip `abi-bindings/`.

Added some basic unit tests for the CLI and other files. For the CLI unit tests running into issues when the tests are run altogether, likely due to them having shared state, and possibly the commands not reinitializing between tests. Deferring the additional test cases to new issue https://github.com/ava-labs/teleporter/issues/174

scripts/lint.sh
```diff
@@ -39,8 +39,8 @@ function golanglinter() {
     # lint e2e tests go code
     go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_LINT_VERSION}
 
-    echo "Linting Teleporter E2E tests Golang code..."
-    cd $TELEPORTER_PATH/tests
+    echo "Linting Golang code..."
+    cd $TELEPORTER_PATH/
```
cam-schultz, line 43: It looks like this will apply linting to all Go files in the repository, including the ABI generated ones. Can you filter those from the linter?

minghinmatthewlam: this is already filtered out by `.golangci.yml`, it ignores the abi gen directory

cam-schultz: Nice, didn't realize that

.github/workflows/test.yml
```diff
@@ -71,3 +71,27 @@ jobs:
...
+        run: |
+          source ./scripts/versions.sh
+          GO_VERSION=$GO_VERSION >> $GITHUB_ENV
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Run ABI Binding Unit Tests
```
michaelkaplan13, line 94: Lets update this step name since it's not just ABI binding tests

cmd/teleporter-cli/root.go
```diff
@@ -0,0 +1,81 @@
...
+
+func init() {
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
+	logLevelArg := rootCmd.PersistentFlags().StringP("log", "l", "", "Log level i.e. debug, info...")
+	rootCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		return rootPreRunE(logLevelArg)
+	}
+}
+
+func rootPreRunE(logLevelArg *string) error {
```
michaelkaplan13, line 45: Possibly dumb question, but what's the `E` suffix denote here and in `callPersistentPreRunE`, etc?

minghinmatthewlam: In `cobra` cli there's commonly functions with an extra `E` suffix as duplicates, i.e. `PersistentPreRun` vs `PersistentPreRunE`. The one's with E just mean return an error

cmd/teleporter-cli/transaction.go
```diff
@@ -0,0 +1,100 @@
...
+	Long: `Given a transaction this command looks through the transaction's receipt
+for Teleporter and Warp log events. When corresponding log events are found,
+the command parses to log event fields to a more human readable format.`,
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		receipt, err := client.TransactionReceipt(context.Background(),
+			common.HexToHash(args[0]))
+		cobra.CheckErr(err)
+		cmd.Print("HELLOOOO")
+		logger.Info("TEST")
```
geoff-vball, line 40: I think you left these here by accident :)

bernard-avalabs (commented):
I'm not familiar with Cobra, but it generally makes sense to me.

cmd/teleporter-cli/transaction.go
minghinmatthewlam: haha yes thanks for finding

michaelkaplan13 (approved):

geoff-vball (approved):

PR #138, opened 2023-11-16
Single source of truth for go version

geoff-vball:
## Why this should be merged
Makes the go.mod file the single source of truth for the go version.

Adds a github action to check the go.mod file for any changes after a `go mod tidy`

Also augments `versions.sh` to get the versions from `go.mod` so we have a single source of truth for those too!

## How this works
`go.mod` only allows `major.minor` version, so I've added a comment with the patch version that we will extract with `sed` in order to install the correct golang version in docker and CI.

## How is this documented


.github/workflows/e2e.yml
```diff
@@ -40,9 +42,6 @@ jobs:
           cd subnet-evm
           ./scripts/build.sh /tmp/e2e-test/avalanchego/plugins/srEXiWaHuhNyGwPUi444Tu47ZEDwxTWrbQiuD7FmgSAQ6X7Dy
 
-      - name: Checkout teleporter repository
```
geoff-vball, line 43: This is already done above on line 23

scripts/versions.sh
```diff
@@ -2,9 +2,21 @@
...
+  cd .. && pwd
+)
+
+# Pass in the full name of the dependency
+function getDepVersion() {
+    grep -m1 $1 $TELEPORTER_PATH/go.mod | cut -d ' ' -f2
+}
+
 # Set up the versions to be used
-AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v0.2.3'}
```
geoff-vball, line 6: This isn't used anymore

.github/workflows/gomod_checker.yml
```diff
@@ -0,0 +1,31 @@
...
+          source ./scripts/utils.sh
+          GO_VERSION=$(getGO_VERSION) >> $GITHUB_ENV
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - run: go mod tidy
+      - run: git --no-pager diff -- go.mod go.sum  # This prints the diff
+      - run: git --no-pager diff --quiet -- go.mod go.sum  g# This errors if there is a diff
```
feuGeneA, line 31:
```suggestion
      - run: git --no-pager diff --quiet -- go.mod go.sum  # This errors if there is a diff
```

scripts/utils.sh
```diff
@@ -35,4 +35,12 @@ function convertToLower() {
...
         echo $1 | sed -e 's/\(.*\)/\L\1/'
     fi
+}
+
+function setGO_VERSION() {
+    export GO_VERSION=$(getGO_VERSION)
+}
+
+function getGO_VERSION() {
+    echo $(grep -m1 go go.mod | cut -d ' ' -f2).$(grep -m1 GO_PATCH_VERSION go.mod | cut -d ' ' -f3)
```
feuGeneA, line 45: Naively, this seems more suited to `versions.sh`. Admittedly, this is a nit, but is there a reason why it should be here instead of there? And have all of the callers of `setGO_VERSION` just `source versions.sh` instead?

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/ava-labs/teleporter
 
 go 1.20
 
+// GO_PATCH_VERSION 10
```
feuGeneA, line 5: Why not this, so that `go mod tidy` can enforce it?
```suggestion
go 1.20.10
```

feuGeneA (changes requested):
Looks good! Thanks for doing this!

"Requested changes" for a single character in `gomod_checker.yml` haha. The other comments are questions for my own education as much as for improvement of this PR.

go.mod
geoff-vball: So this is a little weird... tests work locally for me when this is set, [but on CI, it fails](https://github.com/ava-labs/teleporter/actions/runs/6906754471/job/18792474985). Everything I can find has the `go` version in `go.mod` without a patch version.

It looks like we could use the [toolchain command,](https://go.dev/doc/toolchain) but it seems like too much work for now.

scripts/utils.sh
geoff-vball: Admittedly I didn't know `versions.sh` existed when I started this PR. I've moved it now :)

scripts/versions.sh
```diff
@@ -2,9 +2,27 @@
 # Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 # See the file LICENSE for licensing terms.
 
-# Set up the versions to be used
-AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v0.2.3'}
-SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION:-'v0.5.8'}
```
geoff-vball, line 7: This isn't used either.

scripts/versions.sh
cam-schultz: We do reference the AWM Relayer version in test.yml. Could we keep this var here, and reference it from that workflow?

go.mod
cam-schultz: I haven't tried it myself, but this https://github.com/nektos/act allows you to run github actions locally to help debug

cam-schultz (commented):
Generally LGTM. Made one small request.

scripts/versions.sh
geoff-vball: Sounds good!

minghinmatthewlam (commented):
Generally LGTM, looks like the `test.yml` is complaining about relayer version format

scripts/versions.sh
```diff
@@ -2,9 +2,28 @@
...
+set -e
+set -o pipefail
+
+TELEPORTER_PATH=$(
+  cd "$(dirname "${BASH_SOURCE[0]}")"
+  cd .. && pwd
+)
+
+function setGO_VERSION() {
```
michaelkaplan13, line 13: Nit: Should the capitalization here be `setGoVersion` to be consistent with `getDepVersion`?

geoff-vball: The idea is that `GO_VERSION` is a specific env var, whereas `getDepVersion` takes a param.

michaelkaplan13: Gotcha, it seems a little odd given that we haven't used that convention anywhere else, but I don't feel strongly either way.

michaelkaplan13 (approved):

scripts/versions.sh
```diff
@@ -2,9 +2,28 @@
...
+    export GO_VERSION=$(getGO_VERSION)
+}
+
+function getGO_VERSION() {
+    echo $(getDepVersion go).$(grep -m1 GO_PATCH_VERSION $TELEPORTER_PATH/go.mod | cut -d ' ' -f3)
```
minghinmatthewlam, line 18: instead of get/set functions for `GO_VERSION`, would we be able to just set the value with above `echo $(getDepVersion go).$(grep -m1 GO_PATCH_VERSION $TELEPORTER_PATH/go.mod | cut -d ' ' -f3)` then just `source scripts/versions.sh`?

docker/Dockerfile
```diff
@@ -22,8 +24,8 @@ RUN apt-get install -y wget curl git python3 make gcc build-essential
 
 # Install Golang
 RUN apt-get install -y bison
-RUN wget https://go.dev/dl/go1.20.10.linux-${ARCH}.tar.gz
-RUN rm -rf /usr/local/go && tar -C /usr/local -xzf go1.20.10.linux-${ARCH}.tar.gz
+RUN wget https://go.dev/dl/go${GO_VERSION}.linux-${ARCH}.tar.gz
```
cam-schultz, line 27: If I'm understanding `setGO_VERSION` and `getGO_VERSION` correctly, we just parse the line in `go.mod` containing the Go version. However, that line only contains the major and minor versions (not the bug fix version), so I'm concerned these links may not work. For instance `wget https://go.dev/dl/go1.21.darwin-amd64.tar.gz` returns 404.

geoff-vball: The first part, `$(getDepVersion go)` grabs the major/minor, and then `grep -m1 GO_PATCH_VERSION $TELEPORTER_PATH/go.mod | cut -d ' ' -f3` grabs the patch version I've left in the comment in `go.mod`.

scripts/versions.sh
geoff-vball: I think we *could* do this, but we would have to export it because the docker containers need it exported. I would rather anyone sourcing this file have finer-grained control over what they're exporting and what they're not.

docker/Dockerfile
cam-schultz: Nice 👍

cam-schultz (dismissed):
LGTM!

go.mod
geoff-vball: Unfortunately it seems like `act` is incompatible with `$GITHUB_ENV`, which means I can't run this locally.

michaelkaplan13 (commented):
I don't love the pattern of using a value set within a comment throughout the code base. At the least, lets add a clear explanation to the comment regarding how it must be formatted and how it's used, but I think I would be in favor of just setting the full version explicitly in `versions.sh` and having a comment there and in the go.mod that it should be updated in both places when needed. 

That's similar to how AvalancheGo approaches it here: https://github.com/ava-labs/avalanchego/blob/master/go.mod#L3

feuGeneA (dismissed):

scripts/versions.sh
```diff
@@ -2,9 +2,29 @@
...
+  cd .. && pwd
+)
+
+# Pass in the full name of the dependency.
+# Parses go.mod for a matching entry and extracts the version number.
+function getDepVersion() {
+    grep -m1 "^\s*$1" $TELEPORTER_PATH/go.mod | cut -d ' ' -f2
+}
+
+# AWM_RELAYER is needed for the bash-script integration tests, which should eventually be removed.
```
cam-schultz, line 22:
```suggestion
# AWM_RELAYER_VERSION is needed for the docker run setup, but is not a go module dependency
```

cam-schultz (approved):

cam-schultz (approved):

feuGeneA (approved):

PR #139, opened 2023-11-16
e2e test case: basic send *and receive*

feuGeneA:


geoff-vball (approved):

scripts/local/integration-tests/basic_send_receive.sh
cam-schultz, line 1: Let's hold off on deleting the bash integration tests for now. We're still deciding if there are reasonable use cases for them that the Ginkgo framework can't replicate.

tests/basic_send_receive.go
cam-schultz, line 1: After the message is delivered from B->A, can you have the relayer redeem the reward marked in the receipt? That behavior is not currently covered by any of the existing E2E tests.

cam-schultz (commented):
Two high level requests

scripts/local/integration-tests/basic_send_receive.sh
minghinmatthewlam: would agree here, lets leave the bash scripts. If we decide to remove, we can create a separate issue to remove them altogether.

feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/139/commits/98deca982b6acac8a8e4ec802e36f172e60218c8

minghinmatthewlam (commented):
LGTM, just Cam's comment https://github.com/ava-labs/teleporter/pull/139#discussion_r1397522870 for redeeming rewards and checking the reward amount is correct

tests/add_fee_amount.go
```diff
@@ -112,4 +112,27 @@ func AddFeeAmount(network network.Network) {
...
+
+	// Redeem the relayer reward amount
+	balanceBeforeRedemption, err := mockToken.BalanceOf(
+		&bind.CallOpts{}, fundedAddress,
+	)
+	Expect(err).Should(BeNil())
+	// TODO: figure out why the following is reverting with message
+	// "execution reverted: TeleporterMessenger: no reward to redeem"
+	transaction, err := subnetATeleporterMessenger.RedeemRelayerRewards(
+		&bind.TransactOpts{}, fundedAddress,
```
michaelkaplan13, line 124: I think this should use `mockTokenAddress` instead of `fundedAddress`. The parameter of `redeemRelayerRewards` is the fee asset contract to redeem the rewards for, with the "redeemer" address being the caller of the method. So we should be sure this transaction is being sent from the account specified by the relayer as the reward redeemer, and specify the mock token as the asset being redeemed.

cam-schultz (approved):
LGTM

PR #140, opened 2023-11-16
ERC20 bridge multihop

gwen917:
## Why this should be merged
Implements the following test cases from https://github.com/ava-labs/teleporter/issues/27:

Port the integration test cases in erc20_bridge_multihop.sh into Gingko.

## How this works
See above

## How this was tested
New E2E test case

## How is this documented


tests/utils/local-network-utils/local_network_setup.go
```diff
@@ -80,12 +82,15 @@ func SetupNetwork(warpGenesisFile string) {
 	// Name 10 new validators (which should have BLS key registered)
 	var subnetANodeNames []string
 	var subnetBNodeNames []string
-	for i := 1; i <= 10; i++ {
+	var subnetCNodeNames []string
```
minghinmatthewlam, line 85: I think `validate_churn` should be updated to issue txs to the propose vm for all subnets. Also the node ids collide from 15-19 right now

tests/e2e_test.go
```diff
@@ -51,6 +51,10 @@ var _ = ginkgo.BeforeSuite(func() {
 var _ = ginkgo.AfterSuite(localUtils.TearDownNetwork)
 
 var _ = ginkgo.Describe("[Teleporter integration tests]", func() {
+	// Cross-chain application tests
```
minghinmatthewlam, line 54: after the comment below, and we issue tx to propose vm to all subnets, I'd like for validator churn to not need to be added last. Do you know right now if these tests are in random order?

geoff-vball: Agreed. If we can fix it we should. Otherwise what happens if we have another test that also needs to go last? 🙃

tests/utils/local-network-utils/local_network_setup.go
gwen917: I think I tried it from 16 to 20, but it still get error. I will double check.

tests/e2e_test.go
minghinmatthewlam: instead of making the issue to proposevm calls to each subnet individually, can we iterate through all subnets in the network, get subnet info, and issue tx? To prevent in the future if we add more subnets for this test to break and need to be debugged.

tests/erc20_bridge_multihop.go
```diff
@@ -0,0 +1,448 @@
...
+		ctx,
+		nativeERC20,
+		erc20BridgeAddressA,
+		amount,
+		subnetAInfo,
+		fundedAddress,
+		fundedKey,
+	)
+
+	// Sent a transaction on Subnet A to add support for the the ERC20 token to the bridge on Subnet B
```
cam-schultz, line 77:
```suggestion
	// Send a transaction on Subnet A to add support for the the ERC20 token to the bridge on Subnet B
```

tests/erc20_bridge_multihop.go
```diff
@@ -0,0 +1,448 @@
...
+		subnetAInfo.BlockchainID,
+		erc20BridgeAddressA,
+		nativeERC20Address,
+	)
+	Expect(err).Should(BeNil())
+	Expect(bridgeTokenSubnetBAddress).ShouldNot(Equal(common.Address{}))
+	bridgeTokenB, err := bridgetoken.NewBridgeToken(bridgeTokenSubnetBAddress, subnetBInfo.ChainRPCClient)
+	Expect(err).Should(BeNil())
+
+	// Sent a transaction on Subnet B to add support for the the ERC20 token to the bridge on Subnet C
```
cam-schultz, line 116:
```suggestion
	// Send a transaction on Subnet B to add support for the the ERC20 token to the bridge on Subnet C
```

tests/erc20_bridge_multihop.go
```diff
@@ -0,0 +1,448 @@
...
+		ctx,
+		subnetCInfo,
+		bridgeTokenSubnetCAddress,
+		bridgeTokenC,
+		amount,
+		erc20BridgeAddressC,
+		fundedAddress,
+		fundedKey)
+
+	// Sent a transaction to unwrap tokens from Subnet C back to Subnet A
```
cam-schultz, line 296:
```suggestion
	// Send a transaction to unwrap tokens from Subnet C back to Subnet A
```

cam-schultz (approved):
A few more small comments, otherwiss LGTM

tests/validator_churn.go
```diff
@@ -82,7 +83,7 @@ func ValidatorChurnGinkgo() {
 	// Add new nodes to the validator set
 	log.Info("Adding nodes to the validator set")
 	var nodesToAdd []string
-	for i := 15; i <= 19; i++ {
+	for i := 16; i <= 20; i++ {
```
minghinmatthewlam, line 86: can we check the total initial node count instead of hardcoding 16 here?

tests/validator_churn.go
```diff
@@ -102,6 +104,10 @@ func ValidatorChurnGinkgo() {
 		ctx, subnetBInfo.ChainIDInt, fundedKey, subnetBInfo.ChainWSClient,
 	)
 	Expect(err).Should(BeNil())
+	err = subnetEvmUtils.IssueTxsToActivateProposerVMFork(
```
minghinmatthewlam, line 107: can we iterate through all subnets to issue txs, instead of separate calls for each subnet in case we add subnets down the line.

minghinmatthewlam (approved):
LGTM, comment about checking total nodes instead of hardcoding is addressed in PR https://github.com/ava-labs/teleporter/pull/93.

minghinmatthewlam:
just bmping to create issue related to investigating `validator_churn` failure when different ordering: https://github.com/ava-labs/teleporter/pull/140#discussion_r1397975629

bernard-avalabs (approved):
LGTM

PR #142, opened 2023-11-17
Set all containers to exit if one exits early

geoff-vball:
## Why this should be merged
Causes all docker containers to exit if one exits. This makes tracking down errors much easier.

## How this was tested
Put `exit 1` at the start of `run_setup.sh`, then run `./scripts/local/test.sh` and watch all containers exit when `run_setup.sh` fails.

## How is this documented

### DIFF
--- a/scripts/local/run.sh
+++ b/scripts/local/run.sh
@@ -48,7 +48,7 @@ setARCH
 
 if [ -z "$LOCAL_RELAYER_IMAGE" ]; then
     echo "Using published awm-relayer image"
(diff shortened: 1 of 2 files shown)


cam-schultz (approved):

michaelkaplan13:
Do you know if the exit code for other containers the same as the one that exited? Just want to make sure we don't accidentally have exit codes of 0 in unexpected cases.

geoff-vball:
@michaelkaplan13 Looks like the other containers get sent a `SIGTERM` and return code `143`.

bernard-avalabs (approved):
LGTM

PR #158, opened 2023-11-27
Update README.md mark receipts

Andyvargtz:
Change the Message Receipts flow since the message is not really marked as delivered, it only adds the relayer reward.

## Why this should be merged
Previous version was misleading

## How this works
NA
## How this was tested
NA
## How is this documented
NA


michaelkaplan13 (approved):
🙏 thank you for catching and clarifying this

geoff-vball (approved):

minghinmatthewlam (approved):

PR #160, opened 2023-11-27
Add script for installing foundry

geoff-vball:
## Why this should be merged
Unifies Dockerfile and github CI to both install foundry in the same manner from our forked repo.

## How is this documented

### DIFF
--- a/scripts/install_foundry.sh
+++ b/scripts/install_foundry.sh
@@ -4,8 +4,11 @@
 
 set -e
 
+# The foundry install script uses XDG_CONFIG_HOME as the root of the install.
+# This can vary for different environments, so it is set to $HOME for consistency.
 export XDG_CONFIG_HOME=$HOME
 
+# This installs from ava-labs fork of the foundry repo.
 FOUNDRY_VERSION=v0.1.0
 curl -L https://raw.githubusercontent.com/ava-labs/foundry/${FOUNDRY_VERSION}/foundryup/install > /tmp/foundry-install-script
 sed -i "s/\/ava-labs\/foundry\/master\/foundryup/\/ava-labs\/foundry\/${FOUNDRY_VERSION}\/foundryup/g" /tmp/foundry-install-script


scripts/install_foundry.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -e
+
+FOUNDRY_VERSION=v0.1.0
```
minghinmatthewlam, line 7: should we set this in `versions.sh`?

geoff-vball: IMO we should only use `versions.sh` for a version that is needed in more than one place. Adding indirectly here only increases complexity.

scripts/install_foundry.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -e
+
+export XDG_CONFIG_HOME=$HOME
+
+FOUNDRY_VERSION=v0.1.0
+curl -L https://raw.githubusercontent.com/ava-labs/foundry/${FOUNDRY_VERSION}/foundryup/install > /tmp/foundry-install-script
+sed -i "s/\/ava-labs\/foundry\/master\/foundryup/\/ava-labs\/foundry\/${FOUNDRY_VERSION}\/foundryup/g" /tmp/foundry-install-script
+cat /tmp/foundry-install-script | bash
+echo "export PATH=\"$PATH:$HOME/.foundry/bin\"" >> ~/.bashrc
```
minghinmatthewlam, line 13: why do we need to echo into bash here, with no source following it.

geoff-vball: Line 13? `bashrc` is a file.

scripts/abi_bindings.sh
```diff
@@ -15,6 +15,8 @@ source $TELEPORTER_PATH/scripts/versions.sh
 
 setARCH
 
+export PATH=$PATH:$HOME/.foundry/bin
```
minghinmatthewlam, line 18: can we move this to the CI step?

scripts/install_foundry.sh
minghinmatthewlam: yup why do we need to echo into `.bashrc` here? We don't have a following `source bashrc` so don't think this would take into effect, and we also export it specifically in the next line.

geoff-vball: Ohhh sorry I misunderstood. The idea is that the next time a new shell is opened, it will be able to find it. Though I think the foundry install script may do this on its own, I'll try removing it...

minghinmatthewlam (approved):

scripts/abi_bindings.sh
```diff
@@ -40,8 +40,7 @@ fi
 
 if ! command -v forge &> /dev/null; then
     echo "forge not found, installing"
-    curl -L https://foundry.paradigm.xyz | bash
-    source $HOME/.bashrc
+    $TELEPORTER_PATH/scripts/install_foundry.sh
     foundryup
```
gwen917, line 44: `foundryup` is already run in the install_foundry.sh. Should we remove from here?

geoff-vball: Oh yes, thanks, good catch!

scripts/install_foundry.sh
geoff-vball: Seems to be fine without it, so it's gone.

minghinmatthewlam (dismissed):

scripts/install_foundry.sh
cam-schultz, line 1: Can you add a comment at the top of this file explaining that this installs our fork of foundry, not the official repo?

cam-schultz (commented):
Two tiny comments, otherwise LGTM! Thanks for fixing this.

michaelkaplan13 (approved):
LGTM, thanks for making these consistent. 

Somewhat unrelated, have we published ARM64 builds in our fork?

geoff-vball:
@michaelkaplan13 No we haven't added the ARM64 binaries yet.

cam-schultz:
> @michaelkaplan13 No we haven't added the ARM64 binaries yet.

Updated https://github.com/ava-labs/foundry/releases/tag/v0.1.0 with ARM64 binaries.

PR #161, opened 2023-11-27
Update dependabot ignore

minghinmatthewlam:
## Why this should be merged
Add `subnet-evm` and `avalanchego` to dependabot ignore

## How is this documented


.github/dependabot.yml
```diff
@@ -12,6 +12,9 @@ updates:
     ignore:
       # Update go-ethereum only via subnet-evm or avalanchego
       - dependency-name: "go-ethereum"
+      # Manual integration of subnet-evm that includes compatible avalanchego version
+      - dependency-name: "subnet-evm"
+      - dependency-name: "avalanchego"
```
geoff-vball, line 17: Do we really want to ignore these? I think dependabot opening a new PR is a good reminder that there is a new version, even if it needs manual changes. I think the `1.10.16` update could have been merged as-is.

geoff-vball: Yeah I think @cam-schultz expressed the same idea in slack. I'm on board with ignoring avalanchgo but keeping subnet-evm.

geoff-vball (approved):

michaelkaplan13 (approved):

PR #162, opened 2023-11-28
Fix install foundry mac

cam-schultz:
## Why this should be merged
Fixes the install_foundry.sh to work on Macos

## How this works
Removes -i from sed argument list

## How this was tested
Confirmed the script works on Macos and Linux

## How is this documented
N/A


geoff-vball (approved):
Thanks for fixing this!

michaelkaplan13 (approved):

PR #170, opened 2023-11-30
Standalone application for running E2E test flows against live networks

### LINKED ISSUE #132: Standalone application for running E2E test against live testnets
**Context and scope**
#93 creates Teleporter integration tests that are run against local containerized subnets. Having an application for running the same tests against live testnets will make it easy to test the Teleporter deployment and integration on new networks as they are created.

The networks used for the tests should be configurable.

We could try to make the networks used by the Ginkgo containers configurable to achieve the same effective result, but that may add unnecessary complexity to the test set up, and certain tests (such as validator set churn) aren't able to be run against live testnets in any event. Thanks to @cam-schultz for calling that out.

michaelkaplan13:
## Why this should be merged
Fixes #132 

## How this works
Refactors the E2E test directory structure to allow the test flows to be run against a local network or a live testnet. For live testnets, the application reads the network values from environment variables.

## How this was tested
Set up a "live testnet" including an external using `./scripts/local/run.sh`, and then ran the test flows using `go run tests/testnet/main/run_testnet_flows.go`.

## How is this documented
Updated README


tests/flows/relayer_modifies_message.go
```diff
@@ -12,23 +12,25 @@ import (
...
+) []byte
+
 // Disallow this test from being run on anything but a local network, since it requires special behavior by the relayer
-func RelayerModifiesMessage() {
-	network := &network.LocalNetwork{}
-	subnets := network.GetSubnetsInfo()
-	Expect(len(subnets)).Should(BeNumerically(">=", 2))
-	subnetAInfo := subnets[0]
-	subnetBInfo := subnets[1]
+func RelayerModifiesMessage(network interfaces.Network, constructSignedMessageFunc constructSignedMessageFunc) {
```
cam-schultz, line 32: For tests that should only be run on local networks, what do you think about modifying the signature to accept a `localNetwork` rather than a `network`?

tests/interfaces/network.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Returns the Teleporter contract address for all subnets in this network.
+	GetTeleporterContractAddress() common.Address
+
+	// An address and corresponding key that has native tokens on each of the subnets in this network.
+	GetFundedAccountInfo() (common.Address, *ecdsa.PrivateKey)
+
+	// Whether or not the funded wallet is capable of relaying messages between subnets in this network.
+	// Intended to be true for local networks where all nodes are querable by the test application for their
+	// BLS signatures, and false for testnet networks where test application does not neccarily have
```
cam-schultz, line 39:
```suggestion
	// BLS signatures, and false for testnet networks where test application does not necessarily have
```

tests/flows/deliver_to_wrong_chain.go
```diff
@@ -0,0 +1,110 @@
...
+		ctx,
+		subnetAInfo,
+		subnetBInfo,
+		sendCrossChainMessageInput,
+		fundedKey,
+	)
+
+	if network.SupportsIndependentRelaying() {
+		//
+		// Try to relay the message to subnet B, should fail
```
cam-schultz, line 60:
```suggestion
		// Try to relay the message to subnet C, should fail
```

README.md
```diff
@@ -192,21 +192,16 @@ Then run the following command from the root of the repository:
...
-	gomega.RegisterFailHandler(func(message string, callerSkip ...int) {
-		panic(message)
-	})
-
-  // Run the test, composing it with the Network implementation
-  network := network.NewFujiNetwork()
-  defer network.CloseNetworkConnections()
-	tests.BasicOneWaySend(network)
-}
+The same E2E test flows can be executed against external network by setting the proper environment variables in `.env.testnet` and `.env`, and running the following commands:
```
cam-schultz, line 195: We check in `.env.example`, but `.env` is what's read by the tests. Can you add a step to copy `.env.example` into `.env`?

tests/flows/add_fee_amount.go
```diff
@@ -91,13 +98,19 @@ func AddFeeAmount(network network.Network) {
...
+		mockTokenAddress)
 	Expect(err).Should(BeNil())
-	Expect(amount).Should(Equal(additionalFeeAmount.Add(additionalFeeAmount, initFeeAmount)))
+	Expect(newRewardAmount).Should(Equal(new(big.Int).Add(initialRewardAmount, expectedIncrease)))
 
-	utils.RedeemRelayerRewardsAndConfirm(
-		ctx, subnetAInfo, mockToken, mockTokenAddress, fundedKey, amount,
-	)
+	// If the funded address is the one able to redeem the rewards, do so and check the reward amount is reset.
+	if fundedAddress == receiveEvent.RewardRedeemer {
```
cam-schultz, line 111: Could we use `network.SupportsIndependentRelaying` here? Perhaps renaming that method `FundedAddressIsRelayer` or similar would be more accurate.

cam-schultz: Same applies to this check in the other tests.

tests/flows/send_specific_receipts.go
```diff
@@ -0,0 +1,239 @@
...
+
+		// Relay message
+		network.RelayMessage(ctx, receipt, source, destination, true)
+
+		outstandReceiptCount = getOutstandReceiptCount(source, destination.BlockchainID)
+	}
+	log.Info("Receipt queue emptied")
+}
+
+func getOutstandReceiptCount(source interfaces.SubnetTestInfo, destinationBlockchainID ids.ID) *big.Int {
```
cam-schultz, line 192:
```suggestion
func getOutstandingReceiptCount(source interfaces.SubnetTestInfo, destinationBlockchainID ids.ID) *big.Int {
```

tests/flows/send_specific_receipts.go
```diff
@@ -0,0 +1,239 @@
...
+	tokenAddress common.Address,
+) {
+	// Check the reward amounts.
+	if receiveEvent1.RewardRedeemer == receiveEvent2.RewardRedeemer {
+		amount, err := sourceSubnet.TeleporterMessenger.CheckRelayerRewardAmount(
+			&bind.CallOpts{},
+			receiveEvent1.RewardRedeemer,
+			tokenAddress)
+		Expect(err).Should(BeNil())
+		Expect(amount).Should(Equal(new(big.Int).Mul(relayerFeePerMessage, big.NewInt(2))))
```
cam-schultz, line 224: Can we have the expected amount as a function parameter?

tests/flows/send_specific_receipts.go
```diff
@@ -0,0 +1,239 @@
...
+	relayerFeePerMessage = big.NewInt(5)
+)
+
+func SendSpecificReceipts(network interfaces.Network) {
+	subnetAInfo, subnetBInfo, _ := utils.GetThreeSubnets(network)
+	teleporterContractAddress := network.GetTeleporterContractAddress()
+	_, fundedKey := network.GetFundedAccountInfo()
+	ctx := context.Background()
+
+	// Clear the receipt queue from Subnet B -> Subnet A to have a clean slate for the test flow.
```
cam-schultz, line 28: We might want to disable this test in testnet/main/run_testnet_flows.go since it can result in an arbitrary number of Teleporter messages to clear to queue.

Instead of clearing the receipt queue, could we instead:
- Call `sendSpecificReceipts` and confirm we receive them on Subnet A, as we're doing
- Check the receipt queue on Subnet B directly, and confirm that the receipts are still there

I think any way we write this test, there's going to be a race condition on Fuji since the specified receipts may be delivered by by anybody in between the call to `sendSpecificReceipts` and checking the receipt queue (however we do it) within the test. I'm sure similar Fuji-specific races exist in may of the other tests as well. Something we should keep in mind when running these tests on Fuji.

tests/interfaces/network.go
```diff
@@ -0,0 +1,52 @@
...
+	"math/big"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/subnet-evm/core/types"
+	"github.com/ava-labs/subnet-evm/ethclient"
+	teleportermessenger "github.com/ava-labs/teleporter/abi-bindings/go/Teleporter/TeleporterMessenger"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type SubnetTestInfo struct {
```
cam-schultz, line 15: nit: Since this type is used often, I'd prefer to have it in its own file in the same package.

tests/interfaces/network.go
```diff
@@ -0,0 +1,52 @@
...
+	// For implementations where SupportsIndependentRelaying() is true, relays the specified message between the
+	// two subnets,and returns the receipt of the transaction the message was delivered in.
+	// For implementations where SupportsIndependentRelaying() is false, waits for the specific message to be relayed
+	// by an external relayer, and returns the receipt of the transaction the message was delivered in.
+	RelayMessage(ctx context.Context,
+		sourceReceipt *types.Receipt,
+		source SubnetTestInfo,
+		destination SubnetTestInfo,
+		expectSuccess bool) *types.Receipt
+}
```
cam-schultz, line 52: I'm wondering if it would be worth having a separate `Network` implementation for `deliver_to_wrong_chain.go` that implements `RelayMessage` in that test's unique way for Fuji.  That way, we could remove `SupportsIndependentRelaying` altogether, rather than have two tightly coupled methods.

tests/testnet/network.go
```diff
@@ -0,0 +1,277 @@
...
+	if len(logs) == 0 {
+		return nil, errors.New("Failed to find ReceiveCrossChainMessage log for relayed message")
+	} else if len(logs) > 1 {
+		return nil, errors.New("Found multiple ReceiveCrossChainMessage logs for relayed message")
+	}
+
+	return destination.RPCClient.TransactionReceipt(ctx, logs[0].TxHash)
+}
+
+func (n *testNetwork) SupportsIndependentRelaying() bool {
```
cam-schultz, line 239: nit: Can we group exported methods together?

cam-schultz (commented):
I like this new structure a lot! I left a handful of nits, and a few suggestions for further cleaning things up and making the tests more usable.

tests/flows/relayer_modifies_message.go
michaelkaplan13: I thought of doing that, but unfortunately it creates circular imports between the `local` and `flows` packages.

Another option that I'm open to switching to would be to add the two functions currently passed as parameters to flows to the `network` interface, and then have the implementation of them always error for networks they are not supported on. Thoughts on that?

cam-schultz: That would work, though I'm not a huge fan of adding functions that are only used in one place to an interface.

One other alternative suggestion: What if we added a `LocalNetwork` interface to `package interfaces`? It could just be composed of a `Network`, but that should allow Go's type system to make the distinction. Then we change this function to accept a `LocalNetwork`.

Lots of ways to skin a cat ofc, I don't have too strong an opinion either way since this is all just for testing.

michaelkaplan13: I like even better, good call. Updated it now.

Also I've never heard "lots of way to skin a cat" before, but I like it haha 😅

geoff-vball:
@michaelkaplan13 Are we still planning on removing the integration scripts? If so, it might be easiest to remove those files in this PR to eliminate the big diff that is currently there.

tests/flows/add_fee_amount.go
michaelkaplan13: I considered it, but I think this is actually the more accurate check. An implementation could theoretically support independent relaying but not be the only relayer for that network. In the event that a different relayer was the one to deliver the message, the rewards would still not be redeemable.

tests/flows/send_specific_receipts.go
michaelkaplan13: Nice catch

tests/interfaces/network.go
michaelkaplan13: I think it'd be odd to have a `Network` implementation for a specific test, rather than keeping it one per "network type" (i.e. local network run by the tests themselves, or independent live network).

I don't love having two coupled methods, but think it's reasonable it's reasonable in this case, especially since one is just a boolean flag.

cam-schultz: Yeah, I agree with that. Better to have coupled methods within the same interface than to have tests depend on specific interface implementations (with the notable exception that certain tests only run on local networks)

tests/interfaces/network.go
cam-schultz: Good point, and probably not worth adding a whole other package just for this type. I'm cool leaving this here.

tests/flows/add_fee_amount.go
cam-schultz: Makes sense 👍

tests/flows/relayer_modifies_message.go
cam-schultz: Haha definitely one of the odder idioms

tests/flows/send_specific_receipts.go
michaelkaplan13: Thanks for calling this out. I realized it when getting this flow to work correctly against external networks, but was just focused on getting it working first.

Checking that the receipts are still in the queue also isn't necessarily possible for external testnets because unrelated messages may have already removed them.

Instead, for now I opted to add a `IsExternalNetwork` function to the `Network` interface, and skip the parts of this flow that shouldn't be run against networks that aren't fully controlled by the test application itself.

tests/utils/utils.go
```diff
@@ -196,7 +196,7 @@ func RedeemRelayerRewardsAndConfirm(
 	updatedRewardAmount, err :=
 		subnet.TeleporterMessenger.CheckRelayerRewardAmount(&bind.CallOpts{}, relayerAddress, feeTokenAddress)
 	Expect(err).Should(BeNil())
-	Expect(updatedRewardAmount).Should(Equal(big.NewInt(0)))
+	Expect(updatedRewardAmount.Cmp(big.NewInt(0))).Should(Equal(0))
```
cam-schultz, line 199: FYI @geoff-vball added a function in https://github.com/ava-labs/teleporter/pull/40 that does this: https://github.com/ava-labs/teleporter/blob/4a6059a85f4e4c601d511a077ea29d6411a3c983/tests/utils/utils.go#L538

cam-schultz (dismissed):
LGTM!

tests/interfaces/network.go
michaelkaplan13: I actually went ahead and moved it. If anyone feels strongly let me know, but I'll leave it as is now otherwise.

michaelkaplan13:
> > @michaelkaplan13 Are we still planning on removing the integration scripts? If so, it might be easiest to remove those files in this PR to eliminate the big diff that is currently there.
> 
> Yes, but I'll leave that for #168. If that gets merged first, I'll update this branch (or the other way around)

Cam's PR was merged and this branch is updated with main, so they are removed now 👍

cam-schultz (dismissed):

tests/testnet/network.go
```diff
@@ -0,0 +1,281 @@
...
+// The implementation checks for the deliver of the given message on the destination
+// within a time window of {relayWaitTime} seconds, and returns the receipt of the
+// transaction that delivered the message.
+func (n *testNetwork) RelayMessage(ctx context.Context,
+	sourceReceipt *types.Receipt,
+	source interfaces.SubnetTestInfo,
+	destination interfaces.SubnetTestInfo,
+	expectSuccess bool) *types.Receipt {
+	// Set the context to expire after 20 seconds
+	var cancel context.CancelFunc
```
geoff-vball, line 186: I don't think we need this line

tests/flows/validator_churn.go
```diff
@@ -142,3 +131,16 @@ func ValidatorChurn() {
 	// If this changes in the future, this test will need to perform cleanup by removing the nodes that were added
 	// and re-adding the nodes that were removed.
 }
+
+// Each subnet is assumed that have 5 nodes names nodeN-bls, where N is unique
```
geoff-vball, line 135: nit: change `5` to `nodesPerSubnet`

tests/flows/send_specific_receipts.go
```diff
@@ -0,0 +1,248 @@
...
+) bool {
+	for _, receipt := range receipts {
+		if receipt.ReceivedMessageID.Cmp(expectedMessageID) == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func checkExpectedRewardAmounts(
```
geoff-vball, line 219: Can we either rename this to something that mentions it's for a round-trip or leave a comment saying that this checks the cumulative rewards for two separate receive events for the same tokenAddress?

tests/utils/utils.go
```diff
@@ -451,20 +449,161 @@ func SignTransaction(tx *types.Transaction, key *ecdsa.PrivateKey, chainID *big.
...
+	receipt, err := bind.WaitMined(ctx, subnet.RPCClient, tx)
+	Expect(err).Should(BeNil())
+	Expect(receipt.Status).Should(Equal(types.ReceiptStatusSuccessful))
+
+	return address, exampleMessenger
+}
+
+func DeployERC20Bridge(
+	ctx context.Context,
+	fundedKey *ecdsa.PrivateKey,
```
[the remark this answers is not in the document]
michaelkaplan13: Yeah good call. I updated them all to be `senderKey`, which I think is slightly more self-describing for all cases.

README.md
```diff
@@ -192,21 +192,16 @@ Then run the following command from the root of the repository:
...
+```bash
+set -a
+source .env
+source .env.testnet
+go run tests/testnet/main/run_testnet_flows.go
```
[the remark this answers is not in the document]
michaelkaplan13: Added now!

tests/flows/send_specific_receipts.go
geoff-vball: I figured I might be wrong but thought it would help illustrate my point 😆

docker/run_setup.sh
```diff
@@ -81,10 +81,14 @@ if [ ! -e $dir_prefix/NETWORK_RUNNING ]; then
 
     export PATH="$PATH:$HOME/.foundry/bin"
 
-    subnet_a_url="http://127.0.0.1:9650/ext/bc/$subnet_a_chain_id/rpc"
-    subnet_b_url="http://127.0.0.1:9650/ext/bc/$subnet_b_chain_id/rpc"
-    subnet_c_url="http://127.0.0.1:9650/ext/bc/$subnet_c_chain_id/rpc"
-    c_chain_url="http://127.0.0.1:9650/ext/bc/C/rpc"
+    subnet_a_rpc_url="http://127.0.0.1:9650/ext/bc/$subnet_a_chain_id/rpc"
+    subnet_a_ws_url="ws://127.0.0.1:9650/ext/bc/$subnet_a_chain_id/ws"
```
minghinmatthewlam, line 85: are these webscoket urls used anywhere, or is it added just incase we use in future?

scripts/testnet/run_testnet_e2e_flows.sh
```diff
@@ -0,0 +1,11 @@
+set -e
+
```
minghinmatthewlam, line 2: can you add copyright

docker/run_setup.sh
michaelkaplan13: They're not used anywhere in this script (like most of the variables actually), but they are written out to `var.sh` so that they can be easily used by other scripts/containers if needed.

tests/flows/basic_send_receive.go
```diff
@@ -108,7 +104,11 @@ func BasicSendReceive(network network.Network) {
 	Expect(err).Should(BeNil())
 	Expect(delivered).Should(BeTrue())
 
-	utils.RedeemRelayerRewardsAndConfirm(
-		ctx, subnetAInfo, feeToken, feeTokenAddress, fundedKey, feeAmount,
-	)
+	// If the reward address of the message from A->B is the funded address able to send
```
minghinmatthewlam, line 107:
```suggestion
	// If the reward address of the message from A->B is the funded address, which is able to send
```

tests/flows/erc20_bridge_multihop.go
```diff
@@ -172,37 +189,17 @@ func ERC20BridgeMultihop(network network.Network) {
...
-	Expect(err).Should(BeNil())
-	Expect(actualSymbol).Should(Equal("EXMP"))
-
-	actualDecimals, err := bridgeTokenB.Decimals(&bind.CallOpts{})
-	Expect(err).Should(BeNil())
-	Expect(actualDecimals).Should(Equal(uint8(18)))
-
 	// Check the recipient balance of the new bridge token.
 	actualRecipientBalance, err := bridgeTokenB.BalanceOf(&bind.CallOpts{}, fundedAddress)
 	Expect(err).Should(BeNil())
```
minghinmatthewlam, line 205: can't comment on line 208, but typo `spent`

tests/flows/validator_churn.go
```diff
@@ -132,9 +132,9 @@ func ValidatorChurn(network interfaces.LocalNetwork) {
 	// and re-adding the nodes that were removed.
 }
 
-// Each subnet is assumed that have 5 nodes names nodeN-bls, where N is unique
-// across each subnet. Nodes to be added should thus be named nodeN-bls where
-// N starts one greater than the current total number of nodes.
+// Each subnet is assumed that have {nodesPerSubnet} nodes named nodeN-bls, where
```
cam-schultz, line 135:
```suggestion
// Each subnet is assumed to have {nodesPerSubnet} nodes named nodeN-bls, where
```

cam-schultz (commented):
One tiny nit, otherwise LGTM

tests/flows/erc20_bridge_multihop.go
```diff
@@ -244,6 +247,7 @@ func ERC20BridgeMultihop(network network.Network) {
 
 	// Relay message from SubnetB to SubnetA
 	receipt = network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, true)
```
minghinmatthewlam, line 249: can you add additional comments about how this `receipt` will include the `SendCrossChainmessage` event from A->C since it's not reached its destination, possibly adding to line 260 or here.

cam-schultz (approved):

minghinmatthewlam (commented):
Some nits, but otherwise LGTM

.env.testnet
```diff
@@ -1,21 +1,21 @@
 # subnet A (Amplify)
-subnet_a_url=https://subnets.avax.network/amplify/testnet/rpc
+subnet_a_rpc_url=https://subnets.avax.network/amplify/testnet/rpc
 subnet_a_subnet_id=2PsShLjrFFwR51DMcAh8pyuwzLn1Ym3zRhuXLTmLCR1STk2mL6
 subnet_a_chain_id=2nFUad4Nw4pCgEF6MwYgGuKrzKbHJzM8wF29jeVUL41RWHgNRa
 subnet_a_chain_id_hex=ea70d815f0232f5419dabafe36c964ffe5c22d17ac367b60b556ab3e17a36458
```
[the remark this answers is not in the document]
michaelkaplan13: They were already there and not touched here, so I'd defer remove them to a separate PR

cam-schultz (approved):

geoff-vball (approved):

bernard-avalabs (commented):
LGTM.

One question: Do we have a test where we have more than one relayer competing to relay a message? If not, it might be helpful to test for possible race conditions there.

PR #176, opened 2023-12-04
Upgradeability audit fixes

### LINKED ISSUE #169: Upgradeability audit fixes
**Context and scope**
There were multiple suggestions for fixes from a recent upgradeability audit. We want to address the issues found including:
- Teleporter version can be overwritten to a lower value
- The constructor initializes the _lastestVersion to 0 unnecessarily
- Messages could become permanently stuck if minTeleporterVersion is wrongly set
- The TeleporterUpgradeable lacks helper functions
- The updateMinTeleporterVersion function is marked as external
- Potential race condition between new Teleporter version and updateMinTeleporterVersion, block/pausable will be addresed in #172 
- The MinTeleporterVersionUpdated event might not be emitted during specific updates of the minTeleporterVersion.
- The minTeleporterVersion value could be cached to save gas
- The blockchainID variable could be cached to save gas

minghinmatthewlam:
## Why this should be merged
Fixes #169 

## How this works
- `TeleporterUpgradeable` now provides a base function of `updateMinTeleporterVersion` that takes in a parameter `version`, which much be greater than previous minimum version and a valid version.
- `updateMinTeleporterVersion` now calls new function `_checkTeleporterUpgradeAccess()`, which all derived contracts need to have implemented. `TeleporterOwnerUpgradeable` implements the access check for owners.
- `TeleporterUpgradeable` also now provides a base implementation for `ITeleporterReceiver.receiveTeleporterMessage` that checks whether the caller is an allowed Teleporter version. Derived contracts implement `_receiveTeleporterMessage` for their logic of handling received Teleporter messages.
- If the same protocol address is registered in the registry, `_addressToVersion` saves the highest version number corresponding to the address.
- Gas optimiziations from state variable reads.

## How this was tested
- New unit tests
- CI

## How is this documented
Updated README


contracts/src/.solhint.json
```diff
@@ -1,41 +1,35 @@
...
-        ],
-        "func-named-parameters": [
-            "error",
-            5
-        ],
-        "one-contract-per-file": "off"
-    }
-}
\ No newline at end of file
+  "extends": "solhint:recommended",
```
geoff-vball, line 2: Did you mean to reformat this file?

contracts/src/CrossChainApplications/ERC20Bridge/ERC20Bridge.sol
```diff
@@ -385,6 +311,78 @@ contract ERC20Bridge is
...
     }
 
+    /**
+     * @dev See {TeleporterUpgradeable-receiveTeleporterMessage}.
+     *
+     * Receives a Teleporter message and routes to the appropriate internal function call.
+     */
+    function _receiveTeleporterMessage(
+        bytes32 originBlockchainID,
+        address originSenderAddress,
```
geoff-vball, line 321: We've renamed these to `origin`, which I think is better than `native`. Can we replace the rest of the variables in this contract to align as well?

geoff-vball: Looks like we also use `source` in places. Might be a good idea to align what we can, though some places different terms might make more sense.

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -45,23 +41,30 @@ contract ERC20Bridge is
...
+The `TeleporterUpgradeable.updateMinTeleporterVersion` function updates the `minTeleporterVersion` used to check which Teleporter versions can deliver messages to the dapp, and emits the `MinTeleporterVersionUpdated` event. The `updateMinTeleporterVersion` function should **ONLY** be called by the dapp when it completes delivery of messages from the old Teleporter contract, and now wants to update the `minTeleporterVersion` to only allow the new Teleporter version. By defaut, `updateMinTeleporterVersion` can only be called with a version greater than the current `minTeleporterVersion` and less than `latestVersion` in the Teleporter registry. So once this function is called, the dapp will no longer be able to receive messages from the old Teleporter contract version, unless the old version's Teleporter address was registered in the registry again with a new version.
```
geoff-vball, line 61:
```suggestion
The `TeleporterUpgradeable.updateMinTeleporterVersion` function updates the `minTeleporterVersion` used to check which Teleporter versions can deliver messages to the dapp, and emits the `MinTeleporterVersionUpdated` event. The `updateMinTeleporterVersion` function should **ONLY** be called by the dapp when it completes delivery of messages from the old Teleporter contract, and now wants to update the `minTeleporterVersion` to only allow the new Teleporter version. By default, `updateMinTeleporterVersion` can only be called with a version greater than the current `minTeleporterVersion` and less than `latestVersion` in the Teleporter registry. So once this function is called, the dapp will no longer be able to receive messages from the old Teleporter contract version, unless the old version's Teleporter address was registered in the registry again with a new version.
```

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -56,16 +44,95 @@ abstract contract TeleporterUpgradeable {
...
+     * - `version` must be less than or equal to the latest Teleporter version.
+     * - `version` must be greater than the current minimum Teleporter version.
+     *
+     */
+    function _setMinTeleporterVersion(uint256 version) internal virtual {
+        uint256 latestTeleporterVersion = teleporterRegistry.latestVersion();
+        uint256 oldMinTeleporterVersion = _minTeleporterVersion;
+
+        require(
+            version <= latestTeleporterVersion,
```
michaelkaplan13, line 109: Instead of checking that `version <= latestTeleporterVersion`, should we consider explicitly requiring that `getAddressFromVersion(version) != address(0)`? I believe this is a more restrictive check.

geoff-vball: I think it's a design decision but I think I lean slightly towards the current implementation here.

Imagine we have a min version of 2, latest version of 4, but skipped version 3 entirely. The current implementation lets us set the min version to 3, which might be preferable in the case where someone wants to disallow version 2, but is fine with version 3 if it is later added to the registry.

This might cause bugs elsewhere here though, if we assume that the min version should exist. We should document this very clearly.

cam-schultz (dismissed):
LGTM. I left a suggestion to add a UML diagram to illustrate a common usage of these contracts.

contracts/src/CrossChainApplications/ERC20Bridge/ERC20Bridge.sol
minghinmatthewlam: I took a look through `ERC20Bridge` on the different places we use origin/source, and I think the rest of places that use `native` look appropriate. Will likely defer consolidating naming of `origin` vs `source` for blockchainID and senderAddress to another issue.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: I am also leaning towards allowing the dapp to update to a version that is not yet deployed, but within `minTeleporterVersion` and `latestVersion`.

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -45,23 +41,30 @@ contract ERC20Bridge is
             .getBlockchainID();
     }
     ...
+    function _receiveTeleporterMessage(
```
bernard-avalabs, line 44: Perhaps add some comments to the example application? I expect many readers will jump straight to the example code, so adding comments here would be helpful.

bernard-avalabs (dismissed):
LGTM

contracts/src/Teleporter/upgrades/tests/TeleporterRegistryTests.t.sol
```diff
@@ -377,6 +435,71 @@ contract TeleporterRegistryTest is Test {
...
+            ),
+            abi.encode(warpMessage, true)
+        );
+        vm.expectCall(
+            WARP_PRECOMPILE_ADDRESS,
+            abi.encodeCall(
+                IWarpMessenger.getVerifiedWarpMessage,
+                (messageIndex)
+            )
+        );
```
michaelkaplan13, line 464: This `mockCall` and then `expectCall` pattern looks to be similar in a few places. If it's easy, may want to consider making it a helper function, but no strong preference either way.

michaelkaplan13 (commented):
Looks like there a couple small open comments, but otherwise LGTM!

We should also fix the ABI check prior to merging.

michaelkaplan13 (approved):
LGTM!

I'd be curious to hear @meaghanfitzgerald and @martineckardt's thoughts on the UML diagram we added. I think it's great to include, but also doesn't necessarily make it immediately clear how the contract is intended to be used (the explanation in the README below it does),just my 2 cents.

geoff-vball (dismissed):
Needs a merge resolution, but otherwise LGTM!

geoff-vball (approved):

PR #183, opened 2023-12-07
Update subnet evm

geoff-vball:
## Why this should be merged
Updates subnet-evm

## How this works
subnet-evm's `GetMessageAggregateSignature` now takes a `subnetIDStr` as a parameter. This is used in the case of the C-Chain, so that signatures from subnetA can be fetched when aggregating signatures for a message that needs to be sent to subnetA. 

Should we be checking if the sender is the c-chain here and setting the parameter accordingly? If so, how do we check if the sender is the c-chain?

## How is this documented


gwen917 (approved):

michaelkaplan13:
> Should we be checking if the sender is the c-chain here and setting the parameter accordingly? If so, how do we check if the sender is the c-chain?

Were you able to get an answer to this question? I'm not 100% if the implementation of `GetMessageAggregateSignature` handles this case or not, but we should handle it if not.

cam-schultz:
> > Should we be checking if the sender is the c-chain here and setting the parameter accordingly? If so, how do we check if the sender is the c-chain?
> 
> Were you able to get an answer to this question? I'm not 100% if the implementation of `GetMessageAggregateSignature` handles this case or not, but we should handle it if not.

I imagine we'll have to update other test utilities, as the existing suite were all written with subnet <-> subnet Teleporter messages in mind.

michaelkaplan13:
Based on [this PR](https://github.com/ava-labs/subnet-evm/pull/1008/files), it looks like the API expects to be given the subnet ID of the validators to be queried, so when we add E2E tests that include messages to/from the C-chain, we'll need to update this to set the subnet ID of the destination chain if the origin chain is the C-chain.

Shouldn't be too difficult, but also fine with just adding a comment for now explaining the necessary change.

PR #185, opened 2023-12-07
Pause Teleporter address interaction

### LINKED ISSUE #172: Consider how Teleporter apps can stop receiving messages from vulnerable version
**Context and scope**
Right now for receiving Teleporter messages there is a `minTeleporterVersion`, and when a new version gets published to the registry, the app should wait for all previous version messages to be delivered, then call `updateMinTeleporterVersion`. 

What if the latest version has a vulnerability and shouldn't be used for sending/receiving? Currently the only way is to add a new version to the registry with a Warp off-chain message, and perform an update to `minTeleporterVersion`. 

The registry allows adding old protocol addresses as a new version, so the chain can the latest version that is not vulnerable back to the registry, apps call `updateMinTeleporterVersion`, and they can still receive previous Teleporter messages from that version. What if the chain adds a new protocol address for the new version? Then when apps call `updateMinTeleporterVersion` to stop receiving/sending from a vulnerable version, they'll no longer be able to receive Teleporter messages of the previous valid version that haven't been delivered yet.

We don't want to require chains to add the latest valid Teleporter address to the registry, before adding a new Teleporter address. We also don't want apps to have to rely on the registry update, which requires a Warp off-chain message so needs validator consensus that can take long, to stop receiving/sending from a vulnerable Teleporter version. So need to implement some mechanism in `TeleporterUpgradeable` that allows apps to stop receiving/sending from a specific Teleporter version.

There are two approaches currently in mind, blocking a specific Teleporter version, or pausing Teleporter interaction altogether (all versions). 

With pausing the app would be pausing all versions, so would need to figure out how apps can still receive Teleporter messages from the previous valid versions. If this can only done by relying on a registry update with a previous valid protocol address but new version, then this is less viable. 

For blocking a version, should the app be allowed to unblock a version. Definitely not unblock a version < `minTeleporterVersion`, but can it unblock `version > minTeleporterVersion` in case apps block the wrong version? But blocking is already having the same trust assumption as `updateMinTeleporterVersion`, since if the admins for the app call `updateMinTeleporterVersion` incorrectly they'd also stop receiving Teleporter messages.

For `updateMinTeleporterVersion` we now have `_checkTeleporterUpgradeAccess`, that should be overriden for different access pattern checks, and both approaches should use the same check.

**Open questions**
How do we handle pausing/blocking for sending Teleporter messages? Currently we by default recommend always sending with the `latestVersion`, but what if the latest version has a vulnerability? We wouldn't want to send anymore either. 

This might only be an extra consideration for if `latestVersion` (recommended sending version) is vulnerable. For example, lets say there's version 1,2,3, we are sending on version 3, but still receiving all versions. Version 2 found a bug, we could pause receives, or block version 2, but keep sending with 3. 

However if version 3 had a bug, we'd pause receives, or block version 3, and need to no longer send with version 3. I think we can add a check in the helper to check whether sending version is blocked, if it is, no sending until new vesrion is added.
```
function _getTeleporterMessenger() internal virtual returns(ITeleporterMessenger) {
    uint256 latestVersion = teleporterRegistry.latestVersion();
    // require latsetVersion is not blocked
   return teleporterRegistry.getTeleporterFromVersion(latestVersion);
}
```

minghinmatthewlam:
## Why this should be merged
Adds pause feature in `TeleporterUpgradeable`, fixes #172 .

## How this works
Dapps that integrate with Teleporter through `TeleporterUpgradeable` will have a function `pauseTeleporterAddress`, that is safeguarded by `_checkTeleporterUpgradeAccess`. Pausing a Teleporter address means that Teleporter address will no longer be able to deliver Teleporter messages to the dapp, and will also not be used for sending in `_getTeleporterMessenger()`. 

Pausing is allowed on addresses that are not registered in the registry, so dapps can pause a pre determined Teleporter address before it is deployed. 

## How this was tested
Unit tests
CI

## How is this documented
README


contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -85,13 +103,47 @@ abstract contract TeleporterUpgradeable is ITeleporterReceiver {
...
+    function pauseTeleporterAddress(address teleporterAddress) public virtual {
+        _checkTeleporterUpgradeAccess();
+        require(
+            teleporterAddress != address(0),
+            "TeleporterUpgradeable: zero Teleporter address"
+        );
+        require(
+            !isTeleporterAddressPaused(teleporterAddress),
+            "TeleporterUpgradeable: address already paused"
+        );
```
cam-schultz, line 126: Is there any harm in allowing this function to be called for an already paused address?

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -135,4 +187,27 @@ abstract contract TeleporterUpgradeable is ITeleporterReceiver {
      * This function should be overridden by contracts that inherit from this contract.
      */
     function _checkTeleporterUpgradeAccess() internal virtual;
+
+    /**
+     * @dev Returns the Teleporter messenger used to send Teleporter messages,
+     * and checks that the Teleporter messenger is not paused.
+     *
+     * By default returns the latest Teleporter messenger, but can be overriden to
+     * return a Teleporter messenger of specific version.
```
cam-schultz, line 196:
```suggestion
     * return a Teleporter messenger of a specific version.
```

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -79,18 +79,34 @@ To prevent anyone from calling the dapp's `updateMinTeleporterVersion`, which wo
     }
 ```
 
-For sending messages with the Teleporter registry, dapps should generally use `TeleporterRegistry.getLatestTeleporter` for the latest version, but if the dapp wants to send a message to a specific version, it can use `TeleporterRegistry.getTeleporterFromVersion` to get the specific Teleporter version.
+For sending messages with the Teleporter registry, dapps should generally use `TeleporterUpgradeable._getTeleporterMessenger`, which by default will use the latest version. If the dapp wants to send a message through a specific Teleporter version, it can override `_getTeleporterMessenger()` to use the specific Teleporter version with  `TeleporterRegistry.getTeleporterFromVersion`.
```
cam-schultz, line 82: Can you add here how TeleporterUpgradeable._getTeleporterMessenger extends TeleporterRegistry.getLatestTeleporter?

contracts/src/Teleporter/upgrades/README.md
```diff
@@ -79,18 +79,34 @@ To prevent anyone from calling the dapp's `updateMinTeleporterVersion`, which wo
...
+
+            return teleporter;
+        }
+
+        ITeleporterMessenger teleporterMessenger = _getTeleporterMessenger();
 ```
+
+## Pausing Teleporter interactions
+
+Dapps that inherit from `TeleporterUpgradeable` can pause Teleporter interactions by calling `TeleporterUpgradeable.pauseTeleporterAddress`. This pause function stops receiving Teleporter message from the paused Teleporter address, and also makes sure to not send messages through the paused Teleporter address in `_getTeleporterMessenger()`.
```
cam-schultz, line 110:
```suggestion
Dapps that inherit from `TeleporterUpgradeable` can pause Teleporter interactions by calling `TeleporterUpgradeable.pauseTeleporterAddress`. This function prevents the contract from interacting with the paused Teleporter  address via `_getTeleporterMessenger()`. 
```

This section is worded a bit awkwardly. I suggested a change to clarify things.

cam-schultz (commented):
Generally LGTM! I just have a few minor nits and questions.

contracts/src/Teleporter/upgrades/README.md
geoff-vball: The suggestion seems more clear, but there's an extra space between `Teleporter  address`

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -85,13 +103,47 @@ abstract contract TeleporterUpgradeable is ITeleporterReceiver {
...
+     * to this contract, and this contract will not send messages through that Teleporter address.
+     * The address does not need to be registered with the Teleporter registry.
+     * Emits a {TeleporterAddressPaused} event if successfully paused.
+     * Requirements:
+     *
+     * - `msg.sender` must have Teleporter upgrade access.
+     * - `teleporterAddress` is not already paused.
+     */
+    function pauseTeleporterAddress(address teleporterAddress) public virtual {
+        _checkTeleporterUpgradeAccess();
```
geoff-vball, line 118: Nit: We should probably rename this to `_checkTeleporterAdminAccess` or something similar now that it's used for more than just upgrading.

geoff-vball (commented):
One nit otherwise LGTM

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -135,4 +187,27 @@ abstract contract TeleporterUpgradeable is ITeleporterReceiver {
...
+     */
+    function _getTeleporterMessenger()
+        internal
+        view
+        virtual
+        returns (ITeleporterMessenger)
+    {
+        ITeleporterMessenger teleporter = teleporterRegistry
+            .getLatestTeleporter();
+        require(
```
michaelkaplan13, line 206: Should we allow for sending via non-latest unpaused versions if available?

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: I prefer to leave as is, because some access controls might not have an "admin", for example in Warp off-chain message. Also, I think pausing still has relation to integration with the upgradeability.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: nope there's no harm, I was just thinking as a caller it'd be nice to know if the address was already paused, vs just calling pause again for a noop. Plus the user can save gas from duplicate calls.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: Are you thinking if the `latestVersion` is paused, keep going down the versions to find one that is not paused to send with? That was my initial thought, but I want to stay away from it for the default implementation. Dapps might only be receiving from the latest version, and if they send with a older version they wouldn't be able to receive these messages. Also it becomes less clear which version the dapp sends each Teleporter version with.

geoff-vball: I think the issue with not going down the list is: If we upgrade from say, v2 to v3, but v3 breaks, we have no way to use v2 again until an out-of-band warp-message is sent to add v2 as v4.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
cam-schultz: Cool, just wanted to make sure there wasn't some behavior I was misunderstanding. I'm cool with keeping it as is.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: Yup a new Teleporter version would needed to registered, but I think that is preferable to default to sending on an older Teleporter version, and leading to messages that can't be received.

geoff-vball: I don't think it's too much to ask to have dapp owners be aware of what versions are and are not available on other chains. In fact, I think it's a requirement if you're running a cross-chain app, because new teleporter versions can arrive on different chains at different times. If we wanted to be explicit about it, we could have a separate function call for `enableFallback` or something, which would have to be called while the latest version is paused.

I'm very wary against locking people out of using a teleporter version they've said their contract can use. They can then only use their contract again once the validators submit a new version with an out-of-band warp message, which they are not guaranteed to do, let alone on any particular timeline.

michaelkaplan13: In this case, the contract owners should be able to unpause the version though right? As long as that's the case, I'm fine with pausing the latest version having the effect of being unable to send messages. The origin transaction would revert, so it wouldn't take any effect until it's unpaused.

michaelkaplan13 (approved):

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
geoff-vball: Discussed offline that anything extra here is probably over-engineering.

geoff-vball (approved):

Issue #150, opened 2023-11-20
Warp `bind.WaitMined` with a dealine context in E2E tests


cam-schultz:
We should put a wrapper around `bind.WaitMined` and set a deadline on the context and pass that in. Right now we could possibly hang here forever.

_Originally posted by @geoff-vball in https://github.com/ava-labs/teleporter/pull/93#discussion_r1397543409_

PR #196, opened 2023-12-13
Refactor transaction checking helpers and add timeout

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/150

Cleans up a large portion of the testing utility code.

## How this works
Adds a 10-second timeout when making any call to `bind.waitMined`, otherwise a misconfigured test could hang.

Refactors all tests/utilities to use the same group of functions to send/wait for a transaction.

When we expect a transaction to succeed, but it fails, we will get a trace of the transaction and print it out for debugging purposes.

Normalizes `txn` and `tx` to `tx`.

Normalizes the order of function params to `context.Context, interfaces.SubnetTestInfo, *types.Transaction`.

## How this was tested
E2E Tests

## How is this documented
N/A


tests/local/network.go
```diff
@@ -351,7 +350,8 @@ func (n *localNetwork) RelayMessage(ctx context.Context,
 	sourceReceipt *types.Receipt,
 	source interfaces.SubnetTestInfo,
 	destination interfaces.SubnetTestInfo,
-	expectSuccess bool) *types.Receipt {
+	expectSuccess bool,
+) *types.Receipt {
```
cam-schultz, line 354: Might just be the way Github is displaying this, but can you gofmt this file?

cam-schultz (approved):
LGTM! One tiny comment about a potential formatting issue, but it might also just be the Github UI. Thanks for cleaning this up.

tests/local/network.go
geoff-vball: It's not github, I moved the return type to its own line. All of our other methods do this, but I don't think gofmt/goimports has an opinion on this for whatever reason.

cam-schultz (approved):
LGTM!

michaelkaplan13 (approved):

Issue #198, opened 2023-12-13
Standardized solidity formatting


minghinmatthewlam:
**Context and scope**
For our solidity contracts we don't have a standardized formatter, so will depend on the different formatting rules the developers is using. 

**Discussion and alternatives**
Can take a look at [forge fmt](https://book.getfoundry.sh/reference/config/formatter), and possibly apply a check like the `abi_binding.sh` or `lint.sh` to format the contracts and check that format is correct in CI.

PR #200, opened 2023-12-14
Standardize Solidity Formatting

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/198

## How this works
Uses `forge` to format solidity files. 

Adds a script to automatically format all solidity files in `contracts/src`.

Adds a ci job to check the formatting of of solidity files.

Feel free to suggest any changes to the formatting rules :)

## How this was tested
N/A

## How is this documented


scripts/format_solidity.sh
```diff
@@ -0,0 +1,43 @@
...
+while [ $# -gt 0 ]; do
+    case "$1" in
+        -c | --check) CHECK=true ;;
+        -h | --help) HELP=true ;;
+    esac
+    shift
+done
+
+if [ "$HELP" = true ]; then
+    echo "Usage: ./scripts/lint.sh [OPTIONS]"
```
michaelkaplan13, line 18:
```suggestion
    echo "Usage: ./scripts/format_solidity.sh [OPTIONS]"
```

Actually, any reason to not combine this with existing `lint.sh` script? I don't feel strongly either way, but generally prefer to have fewer scripts to reduce the amount of steps that need to be taken in typical flows.

contracts/foundry.toml
```diff
@@ -5,4 +5,22 @@ libs = ['lib']
...
-# See more config options https://github.com/foundry-rs/foundry/tree/master/config
\ No newline at end of file
+[fmt]
+line_length = 100
+tab_width = 4
+bracket_spacing = false
+int_types = 'preserve'
+multiline_func_header = 'params_first'
+quote_style = 'double'
+number_underscores = 'preserve'
```
michaelkaplan13, line 15: I'd be down to change this to `'thousands'` based on the documentation here: https://book.getfoundry.sh/reference/config/formatter.

contracts/foundry.toml
```diff
@@ -5,4 +5,22 @@ libs = ['lib']
...
+override_spacing = true
+wrap_comments = false
```
michaelkaplan13, line 17: Not 100% what the effects would be, but I think it may make sense to set this to `true` also.

michaelkaplan13 (commented):
I think this is 👌 . Definitely a few spots that look a little awkward with the new formatter, but on the whole I think it's actually a distinct improvement in the formatting, particularly around reducing the number of multi-line statements and breaking up function definitions first by putting parameters only on newlines.

Two suggestions options we may want to consider changing, but LGTM otherwise.

contracts/foundry.toml
geoff-vball: Yeah, I don't mind it in theory, but what it would do right now is just take any comment over `line_length` and put the extra on a newline, so in practice if we have formatted a bunch of comments ourselves, it may end up looking like 

`
// This comment used to be super long 
// but
// the formatter made it kind of ugly.
`

So someone would have to go through and manually clean then up.

michaelkaplan13: Ah yep, don't like that.

michaelkaplan13 (dismissed):
🙏

.github/workflows/test.yml
```diff
@@ -22,10 +22,17 @@ jobs:
...
         run: |
           export PATH=$PATH:$HOME/.foundry/bin
           cd contracts/
           forge test -vvv
+      
+      - name: Check Solidity Formatting
+        if: success() || failure()
+        run: |
+          export PATH=$PATH:$HOME/.foundry/bin
+          ./scripts/lint.sh --sol-format-check
```
cam-schultz, line 35: This seems out of place in the test job. Can we move to the linter job, or better yet a standalone format check job?

.github/workflows/test.yml
```diff
@@ -22,10 +22,17 @@ jobs:
         run: ./scripts/install_foundry.sh
 
       - name: Run unit tests
+        id: sol_unit_tests
```
cam-schultz, line 25: Can you explain how this id is used?

geoff-vball: Was going to use this some for something, but didn't end up needing it. Removed now

.github/workflows/linter.yml
```diff
@@ -43,3 +43,21 @@ jobs:
...
+      - name: Checkout repository and submodules
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install Foundry
+        run: ./scripts/install_foundry.sh
+      
+      - name: Check Solidity Formatting
+        if: success() || failure()
```
cam-schultz, line 60: What's the purpose of this conditional? If it's pertaining to the previous lint jobs, I'd expect it to be directly under `format-solidity`, not in the step.

geoff-vball: Whoops, didn't need this either.

cam-schultz (approved):

michaelkaplan13 (approved):

PR #203, opened 2023-12-15
Convert message IDs to unique hashes

michaelkaplan13:
**Note:** The diff of this PR is relatively large due to the number of tests that needed to be updated to account for a type change from `uint256` to `bytes32`. When reviewing, I recommend starting with `ITeleporterMessenger.sol` and `TeleporterMessenger.sol` as those files contain the only core changes.

## Why this should be merged and how it works
Currently the `TeleporterMessenger` contract assigns a unique ID to each message sent by track how many messages have been sent to a given destination chain, and incrementing the "message ID" to be used messages sent to that chain by each time a message is sent to it. The message ID is used when `TeleporterMessenger` receives a message to provide replay protection by ensuring that the given message has never been received in the past. 

While this is sufficient for the `TeleporterMessenger` contract itself, these message IDs are not unique across chains or `TeleporterMessenger` versions. The first message from one `TeleporterMessenger` instance to any other chain will alway have a message ID of 1, etc. On a global Avalanche level, in order to identify a specific Teleporter message, you must specify:
- The origin blockchain ID
- The destination blockchain ID
- The Teleporter contract address
- The message ID assigned to the message.

This makes indexing and querying for Teleporter messages more cumbersome.

In this PR, the unique `messageID` assigned to each message by the `TeleporterMessenger` contract is changed from a `uint256` value to a `bytes32` value. Instead of being a sequence number of the number of message that have been sent by the `TeleporterMessenger` instance to another blockchain ID, it is computed as:
```
sha256(<teleporter_contract_address>, <current_blockchain_id>, <message_nonce>)
```

The `message_nonce` above is tracked as the total number of message sent from the given `TeleporterMessenger` instance to any other blockchain ID. The resulting `bytes32` message ID is guaranteed to be unique for each message sent by the contract given that the `message_nonce` value will be different for every message. It can be used in the same exact way as the previous message ID `uint256` was to provide replay protection.

The new message ID hash is not included in the `TeleporterMessage` that is sent to the destination. Only the message nonce used to create the message ID hash is included in the message. When `receiveCrossChainMessage` is called to receive a message on the destination, the message ID hash is recomputed by the receiving chain using the source blockchain ID of the Warp message, and the message nonce included in the message. This ensures that message IDs received from different chains will be unique, even if a subnet is corrupted.

Since the message ID values no longer depend on the destination blockchain ID, the `getNextMessageID` function of the `ITeleporterMessenger` interface is also simplified to have no parameters.

**The high level result of this change is that the message IDs assigned to messages by different `TeleporterMessenger` contract instances across different blockchains or at different contract addresses are unique.** This allows for simplified data indexing and querying of messages offchain. 

## How this was tested
Unit tests, E2E test

## How is this documented

TODO 😄


contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -755,9 +763,11 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
     /**
-     * @dev Returns the next message ID to be used to send a message to the given `originBlockchainID`.
+     * @dev Calculates the message ID given this contract's address and required input parameters.
      */
-    function _getNextMessageID(bytes32 originBlockchainID) private view returns (uint256) {
-        return latestMessageIDs[originBlockchainID] + 1;
+    function _calculateNextMessageID() private view returns (bytes32) {
+        bytes32 blockchainID_ = blockchainID;
+        require(blockchainID_ != bytes32(0), "TeleporterMessenger: blockchainID not set");
+        return sha256(abi.encode(address(this), blockchainID_, messageNonce));
```
geoff-vball, line 771: What's the point of a `private view` function, is it some sort of gas optimization?

tests/utils/utils.go
```diff
@@ -152,11 +152,12 @@ func SendSpecifiedReceiptsAndWaitForAcceptance(
 	ctx context.Context,
 	originChainID ids.ID,
 	source interfaces.SubnetTestInfo,
-	messageIDs []*big.Int,
+	messageIDs [][32]byte,
 	feeInfo teleportermessenger.TeleporterFeeInfo,
 	allowedRelayerAddresses []common.Address,
 	senderKey *ecdsa.PrivateKey,
-) (*types.Receipt, *big.Int) {
+	// transactor *teleportermessenger.TeleporterMessenger,
```
geoff-vball, line 159: I think you added this line back in funky merge

contracts/src/CrossChainApplications/NativeTokenBridge/tests/ERC20TokenSourceTests.t.sol
```diff
@@ -17,28 +17,14 @@ import {
...
-    bytes32 private constant _DEFAULT_OTHER_CHAIN_ID =
-        bytes32(hex"abcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcd");
-    address private constant _DEFAULT_OTHER_BRIDGE_ADDRESS =
-        0xd54e3E251b9b0EEd3ed70A858e927bbC2659587d;
-    uint256 private constant _DEFAULT_INITIAL_RESERVE_IMBALANCE = 1000000000;
-    address private constant _DEFAULT_RECIPIENT = 0xa4CEE7d1aF6aDdDD33E3b1cC680AB84fdf1b6d1d;
-    uint256 private constant _DEFAULT_TRANSFER_AMOUNT = 1e18;
-    uint256 private constant _DEFAULT_FEE_AMOUNT = 123456;
-
+contract ERC20TokenSourceTest is NativeTokenBridgeTest {
```
geoff-vball, line 20: Ohhh I was wondering if there was a way to share code between these tests. Thank you 🙏 . I think there's some setup/mocking code they share as well, but I can clean that up after.

geoff-vball (commented):
This looks great! I like it❗

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -62,22 +61,22 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-        bytes32 destinationBlockchainID => mapping(uint256 messageID => SentMessageInfo messageInfo)
+        bytes32 destinationBlockchainID => mapping(bytes32 messageID => SentMessageInfo messageInfo)
     ) public sentMessageInfo;
 
     // Tracks the hash of messages that have been received but whose execution has never succeeded.
     // Enables retrying of failed messages with higher gas limits. Message execution is guaranteed to
     // succeed at most once. The first key is the blockchain ID of the sending chain, the second key is the message ID, and
     // the value is the hash of the uniquely identified message whose execution failed.
-    mapping(bytes32 sourceBlockchainID => mapping(uint256 messageID => bytes32 messageHash)) public
+    mapping(bytes32 sourceBlockchainID => mapping(bytes32 messageID => bytes32 messageHash)) public
```
cam-schultz, line 71: Since the messageID encodes the blockchainID, I believe we can simplify this by removing `sourceBlockchainID` as a key, and just keeping the inner map. Same applies to `_relayerRewardAddresses`. However, we need to keep `sentMessageInfo` as is since the messageID does _not_ encode the `destinationBlockchainID`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -566,9 +566,20 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         TeleporterMessageInput memory messageInput,
         TeleporterMessageReceipt[] memory receipts
-    ) private returns (uint256) {
+    ) private returns (bytes32) {
+        // If the blockchain ID has yet to be initialized, do so now.
+        bytes32 blockchainID_ = blockchainID;
+        if (blockchainID_ == bytes32(0)) {
+            blockchainID_ = WARP_MESSENGER.getBlockchainID();
+            blockchainID = blockchainID_;
+        }
```
cam-schultz, line 575: An internal or private `initializeBlockchainID()` would be good to include since we do this in multiple places now.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -566,9 +566,20 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        require(blockchainID_ != bytes32(0), "TeleporterMessenger: blockchainID not set");
+
         // Get the message ID to use for this message.
-        uint256 messageID = _getNextMessageID(messageInput.destinationBlockchainID);
+        bytes32 messageID = _calculateNextMessageID();
+
+        // Increment the message nonce so the next message will have a different ID
+        ++messageNonce;
```
cam-schultz, line 582: It shouldn't matter functionally, but can we move this to after `bytes memory teleporterMessageBytes = abi.encode(teleporterMessage);`? 

This will be consistent with the pattern before, where we update the state variable pertaining to messageID after this line.

Additionally, I think it makes sense to update the `messageNonce` value only after it's used to contsruct the `TeleporterMessage` (even though now it's indirect)

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -227,14 +227,14 @@ interface ITeleporterMessenger {
...
         bytes32 destinationBlockchainID,
-        uint256 messageID
+        bytes32 messageID
     ) external view returns (address, uint256);
 
     /**
-     * @dev Gets the next message ID to be used for a given chain ID.
-     * @return The next message ID to be used to send a message to the given chain ID.
+     * @dev Gets the message ID to be used for the next message sent from the contract instance.
+     * @return The message ID to be used for the message sent from the contract instance.
```
cam-schultz, line 235:
```suggestion
     * @return The message ID to be used for the next message sent from the contract instance.
```

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -227,14 +227,14 @@ interface ITeleporterMessenger {
...
     function getFeeInfo(
         bytes32 destinationBlockchainID,
-        uint256 messageID
+        bytes32 messageID
     ) external view returns (address, uint256);
 
     /**
-     * @dev Gets the next message ID to be used for a given chain ID.
-     * @return The next message ID to be used to send a message to the given chain ID.
+     * @dev Gets the message ID to be used for the next message sent from the contract instance.
```
cam-schultz, line 234: Can you add a comment here explaining that the return value is not only guaranteed to match the message ID of the next message sent to the contract, regardless of the sender address, or recipient chain or address?

cam-schultz (commented):
Left some comments on `TeleporterMessenger.sol`, and will take another pass to review the downstream changes.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: We have to be careful here for a couple of reasons:
- It is technically not correct to say that the message ID "encodes" the source blockchain ID because there is no way to decode what the source blockchain ID value is given a message ID because it's a one way hash.
- While message IDs _should_ be unique from every source blockchain ID now that the source blockchain ID is used as input to that hash that generates the message ID, a single corrupted source subnet can send arbitrary `messageID` values that conflict with other chains. 

An additional step we could take here, is to verify the correctness of the message ID hash in `receiveCrossChainMessage`. We have the source blockchain ID from the warp message itself, and verify that the sender address is always the same as `address(this)` since Teleporter messages are always sent from/to the same address on other chains. The only additional piece of information needed to be able to re-compute the expected message ID hash is the `messageNonce` used by the sender. 

If we have the message include the nonce instead of the message ID, and then re-compute the message ID on the receiving side, then we would be able to rely on its uniqueness across all source chains and could likely leverage that further.

contracts/src/Teleporter/ITeleporterMessenger.sol
michaelkaplan13: I don't think that's accurate. The return value is guaranteed to match the message ID of the next message sent from the contract. I'm guessing what you're referring to is that a caller of this method may not be guaranteed to be the next sender of a message and thus may get a different message ID assigned to their message, but I think that's assuming a specific use case that doesn't need to be documented in the contract itself.

For instance, a contract could call `getNextMessageID`, and then `sendCrossChainMessage` in the same function such that it does know it is the next caller of `sendCrossChainMessage`. I'm not sure what the use case would be, but something along the lines of:

```
contract MyContract {
    function calculateReportSend() public {
        // Calculate the message ID before sending.
        bytes32 expectedMessageID = teleporterMessenger.getNextMessageID();
        
        // Logic here to do something with expectedMessageID.
        
        // Send the message
        bytes32 messageID = teleporterMessenger.sendCrossChainMessage(...);
        require(expectedMessageID == messageID);
    }
}

Comment (cam-schultz) [reply contracts/src/Teleporter/ITeleporterMessenger.sol:234]: Yup, that's what I meant, thanks. In my original comment, an extraneous "not" snuck its way in. What I meant to say was: 

Can you add a comment here explaining that the return value is ~~not~~ only guaranteed to match the message ID of the next message sent to the contract, regardless of the sender address, or recipient chain or address?

File: contracts/src/Teleporter/TeleporterMessenger.sol (hunk 1 above)
Comment (michaelkaplan13) [reply contracts/src/Teleporter/TeleporterMessenger.sol:777]: It took me a minute to remember why we have it structured this way 😅 

The reason is that interface functions (such as `getNextMessageID`) are required to be `external`. In order to call those methods from within the contract itself, you have to use `this.functionName()`, which uses significantly more gas, and is generally not best practice. 

So instead, we have a `private view` method for internal use and add on `getNextMessageID` which provides the external access point to it.

**Comment by michaelkaplan13** [comment]
I've updated this PR such that the new unique message ID is not included in the messages themselves.

Instead the `messageNonce` used to compute the unique message ID is included in the message, and the receiver side recomputes the unique message ID based on that nonce. This makes it such that the message IDs received by a given Teleporter instance are ensured to be unique across all subnets even if a subnet gets corrupted.

This allows for flattening of multiple data structures that were previously keyed on `[sourceBlockchainID, messageID]`, because it is now guaranteed that `messageID` is unique across all `sourceBlockchainID`s.

File: contracts/src/Teleporter/TeleporterMessenger.sol [hunk 9]
```diff
@@ -46,10 +46,9 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     // The blockchain ID of the chain the contract is deployed on. Initialized lazily on the first call of `receiveCrossChainMessage`
     bytes32 public blockchainID;
 
-    // Tracks the latest message ID used for a given destination chain.
-    // Key is the destination blockchain ID, and the value is the last message ID used for that chain.
-    // Note that the first message ID used for each chain will be 1 (not 0).
-    mapping(bytes32 destinationBlockchainID => uint256 messageID) public latestMessageIDs;
+    // A monotonically incremented integer tracking the total number of message sent by this TeleporterMessenger contract.
```
geoff-vball, line 49:
```suggestion
    // A monotonically incremented integer tracking the total number of messages sent by this TeleporterMessenger contract.
```

geoff-vball (approved):
🙏

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -566,13 +570,17 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     function _sendTeleporterMessage(
         TeleporterMessageInput memory messageInput,
         TeleporterMessageReceipt[] memory receipts
-    ) private returns (uint256) {
+    ) private returns (bytes32) {
+        // If the blockchain ID has yet to be initialized, do so now.
+        bytes32 blockchainID_ = _initializeBlockchainID();
+        require(blockchainID_ != bytes32(0), "TeleporterMessenger: blockchainID not set");
```
cam-schultz, line 576: This check seems unnecessary. We already check against the `bytes32(0)` case inside the function, so this should never occur.

If we want to keep this as a redundant check, can we also add this `require` in the same place in `receiveCrossChainMessage`?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -745,19 +754,32 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+        bytes32 sourceBlockchainID,
+        uint256 nonce
+    ) private view returns (bytes32) {
+        return sha256(abi.encode(address(this), sourceBlockchainID, nonce));
+    }
+
+    /**
+     * @dev Calculates the message ID given this contract's address and required input parameters.
+     */
+    function _calculateNextMessageNonceAndID() private view returns (uint256, bytes32) {
```
cam-schultz, line 779: The wording of this function suggests that it calculates the message ID based on `messageNonce + 1`. Can we rename to `_calculateMessageNonceAndID`?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -123,9 +120,12 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         bytes32 destinationBlockchainID,
```
minghinmatthewlam, line 120: re-looking at this, do we even need to pass in `destinationBlockchainID`? It's already inside `TeleproterMessage`

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: I think we should remove the redundant check

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -566,13 +570,17 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         TeleporterMessageReceipt[] memory receipts
-    ) private returns (uint256) {
+    ) private returns (bytes32) {
+        // If the blockchain ID has yet to be initialized, do so now.
+        bytes32 blockchainID_ = _initializeBlockchainID();
+        require(blockchainID_ != bytes32(0), "TeleporterMessenger: blockchainID not set");
+
         // Get the message ID to use for this message.
-        uint256 messageID = _getNextMessageID(messageInput.destinationBlockchainID);
+        (uint256 messageNonce_, bytes32 messageID_) = _calculateNextMessageNonceAndID();
```
minghinmatthewlam, line 579: can we just use `_calculateMessageID` here? We already have the check that blockchainID != zero bytes

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: imo we can remove this function in favor of `_calculateMessageID`, then for `getNextMessageID` we can add a check for whether blockchainID is set

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -62,23 +61,21 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     // The first key is the blockchain ID of the destination chain, the second key is the message ID, and the value is the info
     // for the uniquely identified message.
     mapping(
-        bytes32 destinationBlockchainID => mapping(uint256 messageID => SentMessageInfo messageInfo)
+        bytes32 destinationBlockchainID => mapping(bytes32 messageID => SentMessageInfo messageInfo)
```
minghinmatthewlam, line 64: I think we can actually get rid of the first key, since `messageID` is now unique regardless of which `destinationBlockchainID` it's sending to, and is unique to `messageNonce`

minghinmatthewlam (commented):
Overall like the changes so far, but I think we'd be able to remove the first key of `destinationBlockchainID` for `sentMessageInfo`, since each messageID is unique regardless of the destinationBlockchainID, and will match to a message nonce. Then in some functions that take the extra destination blockchain ID, it can also be removed as a parameter, i.e. `getFeeinfo`, `getMessageHash`, etc.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -745,19 +754,32 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-     * @dev Returns the next message ID to be used to send a message to the given `originBlockchainID`.
+     * @dev Calculates the message ID for the given source blockchain ID and message nonce.
      */
-    function _getNextMessageID(bytes32 originBlockchainID) private view returns (uint256) {
-        return latestMessageIDs[originBlockchainID] + 1;
+    function _calculateMessageID(
+        bytes32 sourceBlockchainID,
+        uint256 nonce
+    ) private view returns (bytes32) {
+        return sha256(abi.encode(address(this), sourceBlockchainID, nonce));
```
minghinmatthewlam, line 773: why do we opt for sha256 vs keccak256

geoff-vball: A tiny bit of research seems to say keccak is both cheaper (marginally) and more secure (not so important here, but nice to get for free).

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: These changes are made in #209

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Agreed, was already use keccak256 else in the contract also, so I switched to it as a part of #209 to be consistent.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Removed in #209 👍

michaelkaplan13:
Closing in favor of #209

Issue #193, opened 2023-12-12
Author C-Chain <> subnet integration tests


cam-schultz:
**Context and scope**
As of Avalanchego 1.10.17, Warp is activated by default on local networks. There is therefore no special setup required to author integration tests that send Teleporter messages between the C-Chain and subnets.

**Discussion and alternatives**
The test cases we should include are:
- Send basic Teleporter messages between the C-Chain and a subnet
- Replicate/rewrite the native token bridge test cases to source assets from the C-Chain
- Validator churn with the C-Chain as the source chain

geoff-vball:
Would it make sense instead of duplicating tests, to modify the testing interface to provide a couple different choices for `localNetworkInstance`, which a variation that has the c-chain as subnetA or subnetB, then we could pass that instance into the already existing tests.

PR #204, opened 2023-12-18
C chain integ tests

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/teleporter/issues/193

TODO: 
- [x] Modify `run.sh` to use Warp enabled C-Chain once https://github.com/ava-labs/avalanche-cli/pull/1330 is merged and a release issued
- [x] Add back in example bash script, and expand README with usage
- [x] ~~Add ethclient interface that both subnet-evm and coreth ethclient implement~~
  - Tracking in https://github.com/ava-labs/teleporter/issues/208

## How this works
Replaces one of the chains in each of the E2E test cases with the C-Chain.

## How this was tested
E2E test modifications, CI

## How is this documented
N/A


docker/relayerConfigTemplate.json
```diff
@@ -5,7 +5,7 @@
     "source-subnets": [
         {
             "subnet-id": "11111111111111111111111111111111LpoYY",
-            "chain-id": "<C_CHAIN_CHAIN_ID>",
+            "blockchain-id": "<C_CHAIN_CHAIN_ID>",
```
cam-schultz, line 8: TODO: Update the AWM Relayer version in versions.sh once https://github.com/ava-labs/awm-relayer/pull/115 is merged and a new release created.

scripts/local/e2e_test.sh
```diff
@@ -16,6 +16,7 @@ BASEDIR=${BASEDIR:-"$HOME/.teleporter-deps"}
 
 cwd=$(pwd)
 # Install the avalanchego and subnet-evm binaries
+rm -rf $BASEDIR/avalanchego
```
cam-schultz, line 19: A recent change is causing the network artifacts to not fully be cleaned on exit, causing errors on subsequent runs. This works around that.

tests/local/e2e_test.go
```diff
@@ -34,83 +34,83 @@ func TestE2E(t *testing.T) {
...
 var _ = ginkgo.Describe("[Teleporter integration tests]", func() {
 	// Cross-chain application tests
 	ginkgo.It("Send native tokens from subnet A to B and back", func() {
-		flows.NativeTokenBridge(localNetworkInstance)
+		flows.NativeTokenBridge(LocalNetworkInstance)
 	})
 	ginkgo.It("Send ERC20 tokens from subnet A to Native tokens on subnet B and back", func() {
-		flows.ERC20ToNativeTokenBridge(localNetworkInstance)
+		flows.ERC20ToNativeTokenBridge(LocalNetworkInstance)
 	})
```
cam-schultz, line 65: These two test cases currently fail when using the C-Chain on one end. I need to investigate this further.

tests/local/network.go
```diff
@@ -54,22 +59,19 @@ const (
 	fundedKeyStr = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
 )
 
-func newLocalNetwork(warpGenesisFile string) *localNetwork {
+func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
```
cam-schultz, line 62: We export these functions so that they can be used in `awm-relayer`

docker/relayerConfigTemplate.json
michaelkaplan13: Nit, should we renaming the template variables `<BLOCKCHAIN_ID>` rather than `<CHAIN_ID>`?

tests/local/network.go
```diff
@@ -33,14 +37,15 @@ import (
...
 
 // Implements Network, pointing to the network setup in local_network_setup.go
-type localNetwork struct {
-	teleporterContractAddress       common.Address
-	subnetAID, subnetBID, subnetCID ids.ID
-	subnetsInfo                     map[ids.ID]*interfaces.SubnetTestInfo
-	subnetNodeNames                 map[ids.ID][]string
+type LocalNetwork struct {
+	teleporterContractAddress common.Address
+	primaryNetworkInfo        *interfaces.SubnetTestInfo
```
michaelkaplan13, line 45: I think `cchainInfo`, etc is probably more accurate for the naming through out these tests, because the `SubnetTestInfo` doesn't have any value regarding the X or P chains.

tests/local/network.go
```diff
@@ -174,11 +163,61 @@ func newLocalNetwork(warpGenesisFile string) *localNetwork {
...
+	res.setPrimaryNetworkValues()
 	return res
 }
 
-func (n *localNetwork) setSubnetValues(subnetID ids.ID) {
+// Should be called after setSubnetValues for all subnets
+func (n *LocalNetwork) setPrimaryNetworkValues() {
+	var nodeURIs []string
+	nodeURIs = append(nodeURIs, n.subnetsInfo[n.subnetAID].NodeURIs...)
+	nodeURIs = append(nodeURIs, n.subnetsInfo[n.subnetBID].NodeURIs...)
```
michaelkaplan13, line 174: Do you know if there are any other nodes in the local primary network other than the subnet validator nodes? 

If not, could you add a comment explaining that all nodes should be accounted for here?

tests/local/network.go
```diff
@@ -226,24 +265,25 @@ func (n *localNetwork) setSubnetValues(subnetID ids.ID) {
...
-	subnetsInfoList := n.GetSubnetsInfo()
-
 	// Set the package level teleporterContractAddress
 	n.teleporterContractAddress = contractAddress
 
 	ctx := context.Background()
 
-	for _, subnetInfo := range subnetsInfoList {
+	subnets := n.GetSubnetsInfo()
+	subnets = append(subnets, *n.primaryNetworkInfo)
```
michaelkaplan13, line 284: Maybe worth considering something along the lines of a `network.GetAllBlockchainInfo` helper for cases like this.

michaelkaplan13 (commented):
Few small suggestions, but other looking good 👍

tests/local/e2e_test.go
cam-schultz: Fixed in https://github.com/ava-labs/teleporter/pull/207

tests/local/network.go
cam-schultz: While true, I'd prefer to keep the naming consistent with respect to subnet vs blockchain for all subnets, including the primary network.

tests/local/network.go
cam-schultz: Nope, I believe these should be all the nodes. I added a comment noting such.

For context, we only ever use the 1st element of this list, so the actual length and content is pretty arbitrary.

tests/local/network.go
cam-schultz: Good call, added

tests/flows/block_hash_publish_receive.go
```diff
@@ -31,15 +32,6 @@ func BlockHashPublishReceive(network interfaces.Network) {
...
 
-	expectedBlockNumberU64, err := subnetAInfo.RPCClient.BlockNumber(ctx)
-	Expect(err).Should(BeNil())
-	expectedBlockNumber := big.NewInt(0).SetUint64(expectedBlockNumberU64)
-
-	block, err := subnetAInfo.RPCClient.BlockByNumber(
-		ctx, expectedBlockNumber)
-	Expect(err).Should(BeNil())
-	expectedBlockHash := block.Hash()
-
```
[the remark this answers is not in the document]
cam-schultz: After a bit more digging, if a `coreth` `ethclient` is used to grab the block and block height, then everything matches as expected. However, if we dial the same path but with a `subnet-evm` `ethclient`, then we get the mismatch.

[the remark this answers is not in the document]
cam-schultz: Added a workaround to use the proper client in the test. Created an issue to implement a more robust fix.

docker/run_setup.sh
```diff
@@ -43,50 +43,50 @@ if [ ! -e $dir_prefix/NETWORK_RUNNING ]; then
...
     subnet_c_subnet_id=$(cat $HOME/.avalanche-cli/subnets/subnetc/sidecar.json |  getJsonVal "['Networks']['Local Network']['SubnetID']")
     c_chain_subnet_id=11111111111111111111111111111111LpoYY # hardcoded primary subnet ID
-    c_chain_chain_id=$(curl -X POST --data '{"jsonrpc": "2.0","method": "platform.getBlockchains","params": {},"id": 1}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/bc/P | getJsonVal "['result']['blockchains']" | python3 docker/getBlockChainId.py C-Chain)
+    c_chain_blockchain_id=$(curl -X POST --data '{"jsonrpc": "2.0","method": "platform.getBlockchains","params": {},"id": 1}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/bc/P | getJsonVal "['result']['blockchains']" | python3 docker/getBlockChainId.py C-Chain)
 
-    echo "Subnet A chain ID: $subnet_a_chain_id"
-    echo "Subnet B chain ID: $subnet_b_chain_id"
-    echo "Subnet C chain ID: $subnet_c_chain_id"
-    echo "C-Chain chain ID: $c_chain_chain_id"
+    echo "Subnet A chain ID: $subnet_a_blockchain_id"
```
minghinmatthewlam, line 73: nit: can you update the echo to be blockchain ID

scripts/local/examples/basic_send_receive.sh
```diff
@@ -0,0 +1,148 @@
...
+#   subnet_b_blockchain_id_hex
+#   subnet_a_subnet_id_hex
+#   subnet_b_subnet_id_hex
+#   teleporter_contract_address
+#   warp_messenger_precompile_addr
+#   subnet_a_teleporter_registry_address
+#   subnet_b_teleporter_registry_address
+#   subnet_c_teleporter_registry_address
+
+# Test covers:
```
minghinmatthewlam, line 29: what is the purpose for adding just the `basic_send_receive.sh` back vs the other previous test scripts?

tests/flows/block_hash_publish_receive.go
```diff
@@ -29,19 +31,23 @@ func BlockHashPublishReceive(network interfaces.Network) {
 		subnetAInfo.BlockchainID,
 	)
 
-	// gather expectations
-
-	expectedBlockNumberU64, err := subnetAInfo.RPCClient.BlockNumber(ctx)
+	// coreth and subnet-evm have different Block implementations,
+	// which means that the block hashes will be different when queried from using different clients
```
minghinmatthewlam, line 35:
```suggestion
	// which means that the block hashes will be different when queried from different clients
```

tests/flows/block_hash_publish_receive.go
```diff
@@ -29,19 +31,23 @@ func BlockHashPublishReceive(network interfaces.Network) {
...
+	// To workaround this, we need to query the block hash from the coreth client
+	// TODO: Design a unified interface that accounts for this different
+	rpcUri := utils.HttpToRPCURI(subnetAInfo.NodeURIs[1], "C")
```
minghinmatthewlam, line 38: can we set a constant for c chain blockchain ID

tests/flows/block_hash_publish_receive.go
```diff
@@ -4,14 +4,16 @@ import (
...
+	coreEthClient "github.com/ava-labs/coreth/ethclient"
 	"github.com/ava-labs/subnet-evm/accounts/abi/bind"
 	"github.com/ava-labs/teleporter/tests/interfaces"
 	"github.com/ava-labs/teleporter/tests/utils"
 	. "github.com/onsi/gomega"
 )
 
 func BlockHashPublishReceive(network interfaces.Network) {
-	subnetAInfo, subnetBInfo, _ := utils.GetThreeSubnets(network)
+	subnetAInfo := network.GetPrimaryNetworkInfo()
```
minghinmatthewlam, line 15: nit: can you rename these to primarySubnetInfo or similar

minghinmatthewlam: same for other test cases

tests/local/network.go
```diff
@@ -174,11 +162,56 @@ func newLocalNetwork(warpGenesisFile string) *localNetwork {
...
-	res.setSubnetValues(subnetCID)
+	res.setPrimaryNetworkValues()
 	return res
 }
 
-func (n *localNetwork) setSubnetValues(subnetID ids.ID) {
+// Should be called after setSubnetValues for all subnets
+func (n *LocalNetwork) setPrimaryNetworkValues() {
+	// Get the C-Chain node URIs.
+	// All subnet nodes are also C-Chain nodes, so we can include them all here
```
minghinmatthewlam, line 172: nit: can you elaborate on the comment to make it more clear that all subnet nodes validate the primary network, therefore are also c-chain nodes

tests/local/network.go
```diff
@@ -174,11 +162,56 @@ func newLocalNetwork(warpGenesisFile string) *localNetwork {
...
+
+	var cChainBlockchainID ids.ID
+	for _, chain := range blockChains {
+		if chain.Name == "C-Chain" {
+			cChainBlockchainID = chain.ID
+		}
+	}
+	Expect(cChainBlockchainID).ShouldNot(Equal(ids.Empty))
+
+	chainWSURI := utils.HttpToWebsocketURI(nodeURIs[0], "C")
```
minghinmatthewlam, line 188: same as above, lets use a contsant for "C" blockchain id

tests/local/network.go
```diff
@@ -300,53 +336,66 @@ func (n *localNetwork) deployTeleporterRegistryContracts(
...
-		*n.subnetsInfo[n.subnetCID],
 	}
 }
 
-func (n *localNetwork) GetTeleporterContractAddress() common.Address {
+func (n *LocalNetwork) GetPrimaryNetworkInfo() interfaces.SubnetTestInfo {
+	return *n.primaryNetworkInfo
+}
+
+func (n *LocalNetwork) getAllSubnetsInfo() []interfaces.SubnetTestInfo {
```
minghinmatthewlam, line 374: can you add a comment that this includes primary network that is appended at the end

tests/flows/block_hash_publish_receive.go
minghinmatthewlam: any particular reason we use index 1 instead of 0?

tests/local/network.go
michaelkaplan13: Hm, fair point. The previous naming assumption was that a given subnet will only have 1 blockchain, which is still true for the subnets, but not true for primary network.

I'm fine with it for now, but would prefer we clean it up at some point...maybe `BlockchainTestInfo` for the struct name, and `subnetBlockchainInfo` rather than `subnetsInfo` (along with `cChainInfo`) to make everything fully accurate/consistent?

I'm also curious now if we need `subnetAID` and `subnetBID`, and if so if we should also add `primaryNetworkID` here, but can also be deferred to later clean up IMO.

michaelkaplan13:
Other than the small suggestions from Geoff and Matt, LGTM.

scripts/local/examples/basic_send_receive.sh
cam-schultz: After working through a live demo recently, I realized that some reference commands for how to interact with Teleporter via bash would be extremely helpful. This script isn't intended to be run anywhere, nor is it intended to demonstrate anything more that very basic usage. Using this script as a starting point, a user could easily expand to deploying their own contracts, calling other Teleporter methods, etc.

tests/flows/block_hash_publish_receive.go
cam-schultz: It shouldn't really matter, but we use 1 here since `subnetAInfo.RPCClient` already uses the URI at index 0.

tests/flows/block_hash_publish_receive.go
cam-schultz: "Primary Network" is actually the correct terminology: https://docs.avax.network/learn/avalanche/avalanche-platform

geoff-vball (dismissed):

michaelkaplan13 (dismissed):

geoff-vball (approved):
👍

michaelkaplan13 (approved):

PR #207, opened 2023-12-19
Fix Native Token Bridge Tests for c-chain

geoff-vball:
## Why this should be merged
Fixes native token bridge contracts and tests for compatibility with c-chain

## How this works
Fund the deployerAddress in the native token bridge tests from the funded address.

Add `SendNativeTransfer` util function.

`subnet-evm` disallows and transactions being sent to `0x0100000000000000000000000000000000000000` on the c-chain including strict value transfers. This means we need to have a different address to burn tokens with in `NativeTokenSource`. I chose `0x0100000000000000000000000000000000010203` to [fall outside of the reserved range of addresses.](https://github.com/ava-labs/subnet-evm/blob/master/precompile/modules/registerer.go#L20) 

Adds all API endpoints including tracer for the c-chain and subnet configs in our tests.

Added a check to the log for publish block hash.

## How this was tested
E2E tests.

## How is this documented


cam-schultz:
> subnet-evm disallows and transactions being sent to 0x0100000000000000000000000000000000000000 on the c-chain including strict value transfers.

Was this a recent change? I'm wondering why we didn't bump into this before.

tests/local/network.go
```diff
@@ -77,7 +77,13 @@ func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
...
 	f, err := os.CreateTemp(os.TempDir(), "config.json")
 	Expect(err).Should(BeNil())
-	_, err = f.Write([]byte(`{"warp-api-enabled": true}`))
+	_, err = f.Write([]byte(`{
+		"warp-api-enabled": true, 
+		"eth-apis":["eth","eth-filter","net","admin","web3",
+					"internal-eth","internal-blockchain","internal-transaction",
+					"internal-debug","internal-account","internal-personal",
+					"debug","debug-tracer","debug-file-tracer","debug-handler"]
+	}`))
```
cam-schultz, line 86: Can you put this in a string const and reference that here and on line 95?

tests/testnet/network.go
```diff
@@ -56,6 +56,12 @@ type testNetwork struct {
 	fundedKey                 *ecdsa.PrivateKey
 }
 
+// GetPrimaryNetworkInfo implements interfaces.Network.
+func (*testNetwork) GetPrimaryNetworkInfo() interfaces.SubnetTestInfo {
+	// TODO
+	panic("unimplemented")
+}
```
cam-schultz, line 63: Fixed this in the base branch.

tests/local/network.go
```diff
@@ -57,6 +57,13 @@ type LocalNetwork struct {
 
 const (
 	fundedKeyStr = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
+	cChainConfig = `{
```
cam-schultz, line 60: nit: This is used for both the C-Chain and subnet chain config. Can we rename to `warpEnabledChainConfig` or something similar?

cam-schultz (dismissed):

cam-schultz (dismissed):

contracts/src/CrossChainApplications/examples/NativeTokenBridge/ERC20TokenSource.sol
```diff
@@ -30,10 +30,9 @@ contract ERC20TokenSource is
     ITokenSource,
     ReentrancyGuard
 {
-    // The address where the burned transaction fees are credited.
-    // Defined as BLACKHOLE_ADDRESS at
-    // https://github.com/ava-labs/subnet-evm/blob/e23ab058d039ff9c8469c89b139d21d52c4bd283/constants/constants.go
-    address public constant BURNED_TX_FEES_ADDRESS = 0x0100000000000000000000000000000000000000;
+    // Designated Blackhole Address for this contract. Tokens are sent here to be "burned" when
+    // a SourceAction.Burn message is received from the destination chain.
+    address public constant BURN_ADDRESS = 0x0100000000000000000000000000000000010203;
```
michaelkaplan13, line 35: Not for this PR, but are there lower hanging clean up options here to create `TokenSource` parent contract that both `ERC20TokenSource` `NativeTokenSource` could inherit from?

That way we wouldn't have to define this address in multiple places, and have an abstract `_burnTokens` method, etc.

tests/flows/erc20_to_native_token_bridge.go
```diff
@@ -59,45 +61,66 @@ func ERC20ToNativeTokenBridge(network interfaces.LocalNetwork) {
...
-		exampleERC20ContractAddress,
-	)
+	{
+		// Fund the deployer addresses on each chain
+		utils.SendNativeTransfer(
+			ctx,
+			sourceSubnet,
+			fundedKey,
+			deployerAddress,
+			utils.BigIntMul(initialReserveImbalance, big.NewInt(2)),
```
michaelkaplan13, line 71: Is there a reason why we're using `initialReserveImbalance*2` here and below? IIUC, we should be providing the amount needed to deploy the contract, which is likely unrelated to the reserve imbalance. If so, can we update to a new variable/name?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/ERC20TokenSource.sol
geoff-vball: Yeah, we could share about half the code, so I think it would be worthwhile. I'll make a ticket.

tests/flows/erc20_to_native_token_bridge.go
geoff-vball: We also use the deployer address to send tokens from one chain to another. I don't want to use the `funded` address to keep this test independent of the others, and using a different address would add more code. The amount transferred in the test is 1.25*`initialReserveImbalance`, so 2* accounts for contract deployment and gas fees/headroom.

michaelkaplan13: > The amount transferred in the test is 1.25*initialReserveImbalance, so 2* accounts for contract deployment and gas fees/headroom.

Gotcha, thanks for the explanation. For the ERC20->native case it's an independent asset though right? So we only need the amount to deploy the contract?

Also could you add that explanation as a comment in the native->native cases? I think it's very helpful context for who/whenever is looking at these down the line 👍

tests/flows/erc20_to_native_token_bridge.go
```diff
@@ -59,45 +61,69 @@ func ERC20ToNativeTokenBridge(network interfaces.LocalNetwork) {
...
-		deployerPK,
-		sourceSubnet,
-		erc20TokenSourceAbi,
-		sourceSubnet.TeleporterRegistryAddress,
-		destSubnet.BlockchainID,
-		bridgeContractAddress,
-		exampleERC20ContractAddress,
-	)
+	{
+		// deployerAddress needs 10 eth (1e19 wei) to deploy the contract, plus some for gas for other transactions
```
michaelkaplan13, line 65:
```suggestion
		// Fund the deployer address with sufficient native tokens (10 eth = 1e20 wei) on the source chain to deploy the contract and send a number of transfer transactions.
```

I think the `10 eth` you're referring to here is the amount needed to deploy the `TeleporterMessenger` contract using Nick's method. This should be much lower because it uses an accurate gas estimation and dynamic fee txs.

tests/flows/erc20_to_native_token_bridge.go
```diff
@@ -59,45 +61,69 @@ func ERC20ToNativeTokenBridge(network interfaces.LocalNetwork) {
...
-		deployerPK,
-		destSubnet,
-		nativeTokenDestinationAbi,
-		destSubnet.TeleporterRegistryAddress,
-		sourceSubnet.BlockchainID,
-		bridgeContractAddress,
-		initialReserveImbalance,
-	)
+		// deployerAddress needs valueToReturn to attempt (and fail) to send tokens before it is sent any from the source.
+		// It also needs some extra for gas costs, so we send valueToReturn*2
```
michaelkaplan13, line 76:
```suggestion
		// On the destination chain, the deployer address needs valueToReturn native tokens to attempt (and fail) to send tokens before the bridge contracts are collateralized.
		// It also needs some extra for gas costs, so we send valueToReturn*2
```

tests/flows/native_token_bridge.go
```diff
@@ -56,6 +58,27 @@ func NativeTokenBridge(network interfaces.LocalNetwork) {
 	Expect(err).Should(BeNil())
 	log.Info("Native Token Bridge Contract Address: " + bridgeContractAddress.Hex())
 
+	{
+		// The deployer is used to also send native transfers so we don't have to create an additional address.
+		// The deployer will send 1.25*initialReserveImbalance over the course of the test, plus 10 eth to
+		// deploy the contract, plus a little bit of gas, so transfer 2*initialReserveImbalance to be safe.
```
michaelkaplan13, line 64:
```suggestion
		// The deployer is used to also send native transfers so we don't have to create an additional address.
		// The deployer will send 1.25*initialReserveImbalance over the course of the test. Send 2*initialReserveImbalance native tokens
		// so that it also has enough to cover transaction fees, including deploying the contract. 
```

michaelkaplan13 (commented):
LGTM, just a few nits on the wording of comments. I think you may have been confusing the 10eth required to send the Nick's method transaction to deploy `TeleporterMessenger` with the transaction fee of deploying other contracts, and think it's worth avoiding that possible confusion in the comments.

michaelkaplan13 (dismissed):

michaelkaplan13 (approved):

michaelkaplan13 (dismissed):

tests/flows/erc20_to_native_token_bridge.go
```diff
@@ -63,21 +62,26 @@ func ERC20ToNativeTokenBridge(network interfaces.LocalNetwork) {
 	log.Info("Example ERC20 Contract Address: " + exampleERC20ContractAddress.Hex())
 
 	{
-		// Fund the deployer addresses on each chain
+		// Fund the deployer address with sufficient native tokens (100 eth = 1e20 wei) on the source chain to deploy the
+		// contract and send a number of transfer transactions.
+		sourceFundingAmount := utils.BigIntMul(big.NewInt(1e15), big.NewInt(1e5))
```
cam-schultz, line 67: Can we just use `big.NewInt(1e20)`?

tests/flows/native_token_bridge.go
```diff
@@ -56,6 +58,27 @@ func NativeTokenBridge(network interfaces.LocalNetwork) {
...
+	{
+		// The deployer is used to also send native transfers so we don't have to create an additional address.
+		// The deployer will send 1.25*initialReserveImbalance over the course of the test. Send 2*initialReserveImbalance
+		// native tokens so that it also has enough to cover transaction fees, including deploying the contract.
+		utils.SendNativeTransfer(
+			ctx,
+			sourceSubnet,
+			fundedKey,
+			deployerAddress,
+			utils.BigIntMul(initialReserveImbalance, big.NewInt(2)),
```
cam-schultz, line 70: Can we use the same `sourceFundingAmount` as in `erc20_to_native_token_bridge.go`?

tests/flows/erc20_to_native_token_bridge.go
geoff-vball: Nope, `1e19` is the largest that works I believe.  The scientific notation passes in an int64

tests/flows/native_token_bridge.go
geoff-vball: No, this test needs more because we need to send native tokens to the dest subnet. The erc20 test doesn't need these because it's sending erc20 tokens

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenSource.sol
```diff
@@ -32,10 +32,9 @@ contract NativeTokenSource is
     ITokenSource,
     ReentrancyGuard
 {
-    // The address where the burned transaction fees are credited.
-    // Defined as BLACKHOLE_ADDRESS at
-    // https://github.com/ava-labs/subnet-evm/blob/e23ab058d039ff9c8469c89b139d21d52c4bd283/constants/constants.go
-    address public constant BURNED_TX_FEES_ADDRESS = 0x0100000000000000000000000000000000000000;
+    // Designated Blackhole Address for this contract. Tokens are sent here to be "burned" when
+    // a SourceAction.Burn message is received from the destination chain.
+    address public constant BURN_ADDRESS = 0x0100000000000000000000000000000000010203;
```
minghinmatthewlam, line 37: how is this new burn address chosen? If it's not by random can you add in comment.

tests/local/network.go
```diff
@@ -55,7 +55,14 @@ type LocalNetwork struct {
 }
 
 const (
-	fundedKeyStr = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
+	fundedKeyStr           = "56289e99c94b6912bfc12adc093c9b51124f0dc54ac7a766b2bc5ccf558d8027"
+	warpEnabledChainConfig = `{
+		"warp-api-enabled": true, 
+		"eth-apis":["eth","eth-filter","net","admin","web3",
```
minghinmatthewlam, line 61: what is the `eth-apis` config field used for, is it required?

tests/utils/utils.go
```diff
@@ -444,15 +455,31 @@ func waitForTransaction(
...
+func GetEventFromLogsOrTrace[T any](
+	ctx context.Context,
+	receipt *types.Receipt,
+	subnetInfo interfaces.SubnetTestInfo,
+	parser func(log types.Log) (T, error),
+) T {
+	log, err := GetEventFromLogs(receipt.Logs, parser)
+	if err != nil {
+		// Gomega will print the transaction trace
+		Expect(TraceTransaction(ctx, subnetInfo, receipt.TxHash)).Should(Equal(""))
```
minghinmatthewlam, line 478: i feel we can add this check inside the function `TraceTransaction`, and possibly indicate in the naming of the function that it will exit/fail

tests/flows/native_token_bridge.go
cam-schultz: Ah yeah, makes sense.

cam-schultz (dismissed):

tests/local/network.go
geoff-vball: They're different collections of API endpoints that the node can chose to run or not. There's a default set, and then some more advanced ones that not everyone might want to expose. `TraceTransaction` uses one of the non-default APIs, and I've just enabled all the APIs I could find here to make it easier to use others in the future.

tests/utils/utils.go
geoff-vball: Yeah, good suggestion

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenSource.sol
geoff-vball: It's just arbitrary

geoff-vball: Addresses of the form `0x010000000000000000000000000000000000XXXX` are reserved for precompiles.

michaelkaplan13 (dismissed):

cam-schultz (approved):

minghinmatthewlam (approved):

Issue #189, opened 2023-12-08
Consolidate origin vs source blockchainID and senderAddress


minghinmatthewlam:
**Context and scope**
Throughout the repo we use different variations of origin vs source for variables like blockchainID and senderAddress. We should consolidate them to one.

**Discussion and alternatives**
`subnet-evm` uses `sourceChainID` and `originSenderAddress` for Warp https://github.com/ava-labs/subnet-evm/blob/master/contracts/contracts/interfaces/IWarpMessenger.sol

michaelkaplan13:
I'd vote for keeping consistent with `subnet-evm` and using `sourceBlockchainID` and `originSenderAddress` in all instances.

PR #209, opened 2023-12-20
Globally Unique Message IDs

michaelkaplan13:
Note: The diff of this PR is relatively large due to the number of tests that needed to be updated to account for a type change from uint256 to bytes32. When reviewing, I recommend starting with ITeleporterMessenger.sol and TeleporterMessenger.sol as those files contain the only core changes.

## Why this should be merged

Globally unique Teleporter message IDs will provide a significant UX improvement for querying and debugging Teleporter messages.

## How it works
Currently the TeleporterMessenger contract assigns a unique ID to each message sent by track how many messages have been sent to a given destination chain, and incrementing the "message ID" to be used messages sent to that chain by each time a message is sent to it. The message ID is used when TeleporterMessenger receives a message to provide replay protection by ensuring that the given message has never been received in the past.

While this is sufficient for the TeleporterMessenger contract itself, these message IDs are not unique across chains or TeleporterMessenger versions. The first message from one TeleporterMessenger instance to any other chain will alway have a message ID of 1, etc. On a global Avalanche level, in order to identify a specific Teleporter message, you must specify:

- The origin blockchain ID
- The destination blockchain ID
- The Teleporter contract address
- The message ID assigned to the message.

This makes indexing and querying for Teleporter messages more cumbersome.

In this PR, the unique messageID assigned to each message by the TeleporterMessenger contract is changed from a uint256 value to a bytes32 value. Instead of being a sequence number of the number of message that have been sent by the TeleporterMessenger instance to another blockchain ID, it is computed as:
```
keccak256(<teleporter_contract_address>, <current_blockchain_id>, <destination_blockchain_id>, <message_nonce>)
```
The message_nonce above is tracked as the total number of message sent from the given TeleporterMessenger instance to any other blockchain ID. The resulting `bytes32` message ID is guaranteed to be unique for each message sent by the contract given that the `message_nonce` value will be different for every message. It can be used in the same exact way as the previous message ID `uint256` was to provide replay protection.

The new message ID hash is not included in the TeleporterMessage that is sent to the destination. Only the message nonce used to create the message ID hash is included in the message. When `receiveCrossChainMessage` is called to receive a message on the destination, the message ID hash is recomputed by the receiving chain using the source blockchain ID of the Warp message, and the message nonce included in the message. This ensures that message IDs received from different chains will be unique, even if a sending subnet is corrupted.

Similarly, message receipts now only track message nonces, and the ID for the corresponding messages are recomputed when the receipt is received back on the origin chain. Destination chains are not able to forge receipts for messages that were sent to different chains because of the fact that the message ID depends on the `destinationBlockchainID`. If a corrupted subnet were to send a receipt with a message nonce for a message that was sent to a different chain, when that receipt is received back on the origin chain, the resulting message ID would not match any message that had been sent, and the "forged" receipt is effectively a no-op. 

In order to account for `sendSpecificedReceipts` still taking message IDs as a parameter, which is the only value easily available to users, `TeleporterMessenger` now tracks `_receivedMessageNonces` to map received message IDs to the their nonce, and allow for looking up the nonce value to send for specific receipts. This mapping is know used to provide replay protection rather than `relayerRewardAddresses`, since message nonces will never be 0. This allows for removing the requirement that the `relayerRewardAddress` provided to `receiveCrossChainMessage` be non-zero.

These unique message IDs allow for simplifying the `ITeleporterMessenger` interface to no longer need the `destinationBlockchainID` as a parameter for many functions in order to look up a sent message.

## How this was tested
Unit tests, E2E test

## How is this documented
TODO


contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -66,18 +66,14 @@ interface ITeleporterMessenger {
...
-    );
+    event AddFeeAmount(bytes32 indexed messageID, TeleporterFeeInfo updatedFeeInfo);
 
     /**
      * @dev Emitted when a Teleporter message is being delivered on the destination chain to an address,
      * but message execution fails. Failed messages can then be retried with `retryMessageExecution`
      */
     event MessageExecutionFailed(
-        bytes32 indexed originBlockchainID, uint256 indexed messageID, TeleporterMessage message
+        bytes32 indexed messageID, bytes32 indexed originBlockchainID, TeleporterMessage message
```
minghinmatthewlam, line 76: not sure if we'd care about `originBlockchainID` in this event and `MessageExecuted` if message ids are unique

michaelkaplan13: Yeah I thought about removing it entirely, but think it could still be helpful in tracking down the source chain message since that is not known for a given `messageID` even though the message ID is unique.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -177,7 +175,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         // will be cleared from state. At this point, you can not add to its fee. This is also the
         // case if the given message never existed.
         require(
-            sentMessageInfo[destinationBlockchainID][messageID].messageHash != bytes32(0),
+            sentMessageInfo[messageID].messageHash != bytes32(0),
```
minghinmatthewlam, line 178: i think we looked at this before for gas savings, and not part of PR, but surprised we do so many look up of a state variable, instead of saving in local variable.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -245,42 +232,47 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         // Check that the caller is allowed to deliver this message.
         require(
             _checkIsAllowedRelayer(msg.sender, teleporterMessage.allowedRelayerAddresses),
             "TeleporterMessenger: unauthorized relayer"
         );
 
-        // Store the relayer reward address provided, effectively marking the message as received.
-        _relayerRewardAddresses[warpMessage.sourceChainID][teleporterMessage.messageID] =
-            relayerRewardAddress;
+        // Store the message nonce, effectively marking the message as received.
```
minghinmatthewlam, line 265: think we can use a private function like `_markReceived` or `setReceived` to make it more clear

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -493,20 +496,19 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
 
     /**
-     * @dev Returns the next message ID to be used to send a message to the given blockchain ID.
+     * @dev Gets the next message ID to be used for a message sent from the contract instance.
+     * @return The next message ID to be used for a message sent from the contract instance.
      */
-    function getNextMessageID(bytes32 destinationBlockchainID) external view returns (uint256) {
-        return _getNextMessageID(destinationBlockchainID);
+    function getNextMessageID(bytes32 destinationBlockchainID) external view returns (bytes32) {
+        bytes32 blockchainID_ = blockchainID;
```
minghinmatthewlam, line 509: what's the decision between this and calling `initializeBlockChainID`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -558,6 +569,20 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return false;
     }
 
+    /**
+     * @dev If not already set, initialize blockchainID by getting the current blockchain ID
+     * value from the Warp precompile.
+     * @return The current blockchain ID.
+     */
+    function _initializeBlockchainID() private returns (bytes32) {
```
minghinmatthewlam, line 577: we initialized move the `blockchainID` initialization out of constructor for deployment reasons, but now that we call initialize on a bunch of places, what is the difference with setting it in the state variable declaration again?. i.e. 
`bytes32 public immutable blockchainID = WARP_MESSENGER.getBlockchainID();`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -568,13 +593,18 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         TeleporterMessageReceipt[] memory receipts
-    ) private returns (uint256) {
+    ) private returns (bytes32) {
+        // If the blockchain ID has yet to be initialized, do so now.
+        bytes32 blockchainID_ = _initializeBlockchainID();
+
         // Get the message ID to use for this message.
-        uint256 messageID = _getNextMessageID(messageInput.destinationBlockchainID);
+        uint256 messageNonce_ = messageNonce;
+        bytes32 messageID_ =
```
minghinmatthewlam, line 602: my understanding is we add `_` to suffix when it's repeated with a state variable, but we don't have a state variable `messageID` so might not be needed, but no strong pref here

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -747,19 +782,12 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
      */
     function _storeFailedMessageExecution(
         bytes32 originBlockchainID,
+        bytes32 messageID,
```
minghinmatthewlam, line 785: nit: we might want to have `messageID` as first parameter in functions, same for other functions

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -393,23 +393,35 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         TeleporterMessageReceipt[] memory receiptsToSend = new TeleporterMessageReceipt[](
                 messageIDs.length
             );
         for (uint256 i; i < messageIDs.length; ++i) {
-            uint256 receivedMessageID = messageIDs[i];
+            bytes32 receivedMessageID = messageIDs[i];
+
+            // Get the nonce for this message ID.
+            uint256 receivedMessageNonce = _receivedMessageNonces[receivedMessageID];
+            require(receivedMessageNonce != 0, "TeleporterMessenger: receipt not found");
```
cam-schultz, line 411: It would be helpful to emit the index of the missing receipt in the input array.

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: If we call `initializeBlockChainID`, this function could not be `view`

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -526,15 +528,24 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return receiptQueues[originBlockchainID].getReceiptAtIndex(index);
     }
 
+    /**
+     * @dev Calculates the message ID for the given source blockchain ID and message nonce.
+     */
+    function calculateMessageID(
+        bytes32 sourceBlockchainID,
+        bytes32 destinationBlockchainID,
```
cam-schultz, line 536: I wonder if it would be worth trying to distinguish these two using different type aliases.

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: The state variable would still be initialized during deployment, so the same consideration around `WARP_MESSENGER.getBlockchainID` reverting would still apply.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -568,13 +593,18 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
         TeleporterMessageInput memory messageInput,
         TeleporterMessageReceipt[] memory receipts
-    ) private returns (uint256) {
+    ) private returns (bytes32) {
+        // If the blockchain ID has yet to be initialized, do so now.
+        bytes32 blockchainID_ = _initializeBlockchainID();
+
         // Get the message ID to use for this message.
-        uint256 messageID = _getNextMessageID(messageInput.destinationBlockchainID);
+        uint256 messageNonce_ = messageNonce;
```
cam-schultz, line 601: Is the reason for this local variable just to make the function less error prone for developers? From a correctness standpoint, I think we can use `messageNonce` directly, as long as we finish using it before we increment.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -633,12 +663,16 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
      * the message information from `sentMessageInfo` and increments the reward redeemable by the specified relayer reward address.
      */
     function _markReceipt(
-        bytes32 destinationBlockchainID,
-        uint256 messageID,
+        bytes32 sourceBlockchainID_,
```
cam-schultz, line 666: nit: we should be consistent between `sourceBlockchainID` and `originBlockchainID` in our variable naming.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: It's actually a gas optimization that Louis called out. `messageNonce` is kept in storage, so reading it once into a local variable reduces the total number of storage reads since we use the value more than once.

cam-schultz: Nice, makes sense

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Yeah good call. I went with `_markMessageReceived` and also moved the check that the message nonce is non-zero to be within it to make it less error prone.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Removed, agree to keep that convention consistent as only where required.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I agree, and I think we should go with `sourceBlockchainID`, because because `subnet-evm` uses the "source" terminology for chain ID's. Interestingly, `subnet-evm` also uses `originSenderAddress`.

Deferring it to this issue though: https://github.com/ava-labs/teleporter/issues/189

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: That's an interesting idea. I think it requires it's own investigation outside the scope of this issue though, so I'm going to leave it as is for now.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I think this is actually more difficult that it's worth because it requires converting the `uint256` or `bytes32` to a string, and then doing string manipulation to format the error message.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -245,42 +232,47 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-            "TeleporterMessenger: message already delivered"
+        // Require that the message nonce is non-zero because the value is used to provide replay protection.
+        require(teleporterMessage.messageNonce != 0, "TeleporterMessenger: zero message nonce");
+
+        // Calculate the message ID of the message given the source blockchain ID and message nonce.
+        bytes32 messageID = calculateMessageID(
+            warpMessage.sourceChainID, blockchainID_, teleporterMessage.messageNonce
         );
 
+        // Require that the message has not been delivered previously.
```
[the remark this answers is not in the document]
michaelkaplan13: Updating this in all of the other places where we used "delivered" incorrectly also. The only correct use of "delivered" I think is when referencing the relayer that delivered a message to the destination.

contracts/src/CrossChainApplications/examples/ERC20Bridge/tests/ERC20BridgeTests.t.sol
```diff
@@ -645,6 +645,11 @@ contract ERC20BridgeTest is Test {
         vm.expectCall(tokenContract, abi.encodeCall(token.decimals, ()));
     }
 
+    function _createMessageID(uint256 messageNonce) private pure returns (bytes32) {
```
minghinmatthewlam, line 648: we should update to match new design

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenBridgeTests.t.sol
```diff
@@ -0,0 +1,31 @@
...
+    bytes32 internal constant _DEFAULT_OTHER_CHAIN_ID =
+        bytes32(hex"abcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcdefabcd");
+    address internal constant _DEFAULT_OTHER_BRIDGE_ADDRESS =
+        0xd54e3E251b9b0EEd3ed70A858e927bbC2659587d;
+    uint256 internal constant _DEFAULT_INITIAL_RESERVE_IMBALANCE = 1000000000;
+    address internal constant _DEFAULT_RECIPIENT = 0xa4CEE7d1aF6aDdDD33E3b1cC680AB84fdf1b6d1d;
+    uint256 internal constant _DEFAULT_TRANSFER_AMOUNT = 1e18;
+    uint256 internal constant _DEFAULT_FEE_AMOUNT = 123456;
+
+    function _createMessageID(uint256 messageNonce) internal pure returns (bytes32) {
```
minghinmatthewlam, line 27: same here

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -46,10 +46,10 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     // The blockchain ID of the chain the contract is deployed on. Initialized lazily on the first call of `receiveCrossChainMessage`
```
minghinmatthewlam, line 46: can you update this comment

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -59,26 +59,22 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
 
     // Tracks the message hash and fee information for each message sent that has yet to be acknowledged
     // with a receipt. The messages are tracked per chain and keyed by message ID.
```
minghinmatthewlam, line 61:
```suggestion
    // with a receipt. 
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -59,26 +59,22 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // The first key is the blockchain ID of the destination chain, the second key is the message ID, and the value is the info
-    // for the uniquely identified message.
-    mapping(
-        bytes32 destinationBlockchainID => mapping(uint256 messageID => SentMessageInfo messageInfo)
-    ) public sentMessageInfo;
+    // The key is the message ID, and the value is the info for the uniquely identified message.
+    mapping(bytes32 messageID => SentMessageInfo messageInfo) public sentMessageInfo;
 
     // Tracks the hash of messages that have been received but whose execution has never succeeded.
```
minghinmatthewlam, line 65:
```suggestion
    // Tracks the hash of messages that have been received but have never succeeded in execution.
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -59,26 +59,22 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // provided by the deliverer of the uniquely identified message.
-    mapping(
-        bytes32 sourceBlockchainID => mapping(uint256 messageID => address relayerRewardAddress)
-    ) internal _relayerRewardAddresses;
+    // succeed at most once. The key is the message ID, and the value is the hash of the uniquely
+    // identified message whose execution failed.
+    mapping(bytes32 messageID => bytes32 messageHash) public receivedFailedMessageHashes;
+
+    // Tracks the message nonce for each message that has been received.
+    // Note that these values are also used to determine if a given message has been received or not.
```
minghinmatthewlam, line 72:
```suggestion
    // Note: the `messageNonce` values are also used to determine if a given message has been received or not.
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -289,7 +278,10 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         for (uint256 i; i < length; ++i) {
             TeleporterMessageReceipt memory receipt = teleporterMessage.receipts[i];
             _markReceipt(
```
minghinmatthewlam, line 280: might be gas savings but not necessary, we could pass in the receipt directly to `_markReceipt`, instead of individual fields then we don't need the extra variable declaration above

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -332,9 +324,14 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         bytes32 originBlockchainID,
         TeleporterMessage calldata message
     ) external receiverNonReentrant {
+        bytes32 blockchainID_ = _initializeBlockchainID();
```
minghinmatthewlam, line 327: don'pt think we need to intiailzie here, since we'll have to pass the check that a message was found, and can only pass if we've recieved a Teleporter message before, meaning blockchain id was initialized.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -393,23 +390,35 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
      */
     function sendSpecifiedReceipts(
         bytes32 originBlockchainID,
-        uint256[] calldata messageIDs,
+        bytes32[] calldata messageIDs,
         TeleporterFeeInfo calldata feeInfo,
         address[] calldata allowedRelayerAddresses
-    ) external senderNonReentrant returns (uint256) {
+    ) external senderNonReentrant returns (bytes32) {
+        bytes32 blockchainID_ = _initializeBlockchainID();
```
minghinmatthewlam, line 397: same here, think `blockchainID` must be initialized to pass the receipt found check, since we need to receive before

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -453,31 +462,22 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
     /**
      * @dev See {ITeleporterMessenger-getRelayerRewardAddress}
      */
-    function getRelayerRewardAddress(
-        bytes32 originBlockchainID,
-        uint256 messageID
-    ) external view returns (address) {
-        return _relayerRewardAddresses[originBlockchainID][messageID];
+    function getRelayerRewardAddress(bytes32 messageID) external view returns (address) {
+        return _relayerRewardAddresses[messageID];
```
minghinmatthewlam, line 480: should we require that the message was received before?

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -225,16 +208,13 @@ interface ITeleporterMessenger {
      * @dev Gets the fee token address and amount for a given message.
      * @return The fee token address and fee amount for a the given message.
      */
-    function getFeeInfo(
-        bytes32 destinationBlockchainID,
-        uint256 messageID
-    ) external view returns (address, uint256);
+    function getFeeInfo(bytes32 messageID) external view returns (address, uint256);
```
minghinmatthewlam, line 211: can you add a comment about what happens if a messageID isn't found.

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: I think I'm leaning towards `initializeBlockChainID`, since we'll want to intiailize it eventually, don't think having `view` is a must, and right now you can't `getNextMessageID` until it is initialized, so user's can't call it for `messageNonce=1`.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -526,15 +525,24 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return receiptQueues[originBlockchainID].getReceiptAtIndex(index);
     }
 
+    /**
+     * @dev Calculates the message ID for the given source blockchain ID and message nonce.
```
minghinmatthewlam, line 529: can you update comemnt to include destination blockchainID and addres of this instance

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -626,19 +653,33 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return messageID;
     }
 
+    /**
+     * @dev Marks a message as being received by storing the message nonce associated with the
+     * given message ID. The message must not be zero in order to be able to distinguish between
+     * received and unreceived message based on their ID.
```
minghinmatthewlam, line 659:
```suggestion
     * received and unreceived messages based on their ID.
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -672,10 +713,11 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
      */
     function _handleInitialMessageExecution(
         bytes32 originBlockchainID,
+        bytes32 messageID,
```
minghinmatthewlam, line 716: nit: can we make `messageID` first parameter

tests/utils/utils.go
```diff
@@ -697,3 +696,21 @@ func GetThreeSubnets(network interfaces.Network) (
...
 	return subnets[0], subnets[1], subnets[2]
 }
+
+func CalculateMessageID(
+	teleporterAddress common.Address,
+	sourceBlockchainID ids.ID,
+	destinationBlockchainID ids.ID,
+	messageNonce *big.Int,
+) ids.ID {
+	// TODO: Clean up ABI encoding of parameters.
```
minghinmatthewlam, line 706: double checking on this TODO, wasn't too sure what it was about and if it still applies

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I feel pretty strongly about keeping any `get` method as `view` actually. I think it'd be an anti-pattern to have a get method alter the state of the contract. I agree it's odd to not be able to call `getNextMessageID` before a message has been sent or received though. I think the best option here is to just make `initializeBlockchainID` public, so that it can be explicitly initialized before sending/receiving messages, if desired.

contracts/src/CrossChainApplications/examples/ERC20Bridge/tests/ERC20BridgeTests.t.sol
michaelkaplan13: Good call. I opted to remove `_createMessageID` all together in application-level tests in favor of just using a mock value. For for Teleporter tests, we need to calculate the expected values the contract will create, but for application tests, we mock out the value Teleporter returns either way. 👍

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenBridgeTests.t.sol
michaelkaplan13: Removed now

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Nice clean up catch 🙏

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I was thinking it would be a more descriptive error message in the case it hadn't been initialized yet, but you're right, "message not found" is just as descriptive because receiving the message will initialize the blockchain ID if necessary.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Yeah I think that makes sense now that we don't require relayer reward addresses to be non-zero

contracts/src/CrossChainApplications/examples/ERC20Bridge/tests/ERC20BridgeTests.t.sol
```diff
@@ -645,6 +645,11 @@ contract ERC20BridgeTest is Test {
...
+        return
+            sha256(abi.encode(MOCK_TELEPORTER_MESSENGER_ADDRESS, _MOCK_BLOCKCHAIN_ID, messageNonce));
```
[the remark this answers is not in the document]
michaelkaplan13: I now have removed this all together, explained [here](https://github.com/ava-labs/teleporter/pull/209#discussion_r1435145261). 👍

contracts/src/Teleporter/ITeleporterMessenger.sol
michaelkaplan13: Added

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -225,16 +208,13 @@ interface ITeleporterMessenger {
...
 
     /**
-     * @dev Gets the next message ID to be used for a given chain ID.
-     * @return The next message ID to be used to send a message to the given chain ID.
+     * @dev Gets the message ID to be used for the next message sent from the contract instance.
+     * @return The message ID to be used for the next message sent from the contract instance.
```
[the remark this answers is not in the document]
michaelkaplan13: Updated 👍 Also included the note that @cam-schultz originally suggested regarding the return value not be ensured to be the next message ID assigned to the caller's message, given that other messages may be sent still.

contracts/src/Teleporter/tests/ReceiveCrossChainMessageTests.t.sol
```diff
@@ -163,17 +173,35 @@ contract ReceiveCrossChainMessagedTest is TeleporterMessengerTest {
         teleporterMessenger.receiveCrossChainMessage(0, DEFAULT_RELAYER_REWARD_ADDRESS);
     }
 
-    function testInvalidRelayerAddress() public {
```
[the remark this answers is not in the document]
michaelkaplan13: Added [here](https://github.com/ava-labs/teleporter/blob/use-nonce-for-receipts/contracts/src/Teleporter/tests/GetRelayerRewardAddressTests.t.sol#L36)

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: Good call

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: good call, I rethought about it and it is good to have `get` methods as read only on state. Agree that we can just have `initializeBlockchainID` public.

contracts/src/Teleporter/TeleporterMessenger.sol
minghinmatthewlam: I looked at this originally incorrect actually. I thought we could check whether the message was found first without the `blockchainID`, but blockchainID needs to be intiialized to calculate `messageID`. It still works since if `blockchainID` is not initialized, the messageID will be computed incorrectly. Can you add the same comment you added to `sendSpecifiedReceipts` to here?

tests/utils/utils.go
```diff
@@ -700,3 +699,15 @@ func GetTwoSubnets(network interfaces.Network) (
 	Expect(len(subnets)).Should(BeNumerically(">=", 2))
 	return subnets[0], subnets[1]
 }
+
+func CalculateMessageID(
```
minghinmatthewlam, line 703: I don't think we need this extra utils function and can call the bidnings functions directly in test

contracts/src/Teleporter/ITeleporterMessenger.sol
```diff
@@ -223,18 +206,20 @@ interface ITeleporterMessenger {
 
     /**
      * @dev Gets the fee token address and amount for a given message.
```
minghinmatthewlam, line 208:
```suggestion
     * @dev Gets the fee token address and amount for a given sent message.
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenBridgeTests.t.sol
cam-schultz, line 1: Can we rename this file and contract to reflect that there are no tests run, just constants imported by other tests?

contracts/src/Teleporter/tests/GetRelayerRewardAddressTests.t.sol
```diff
@@ -16,26 +16,43 @@ contract GetRelayerRewardAddressTest is TeleporterMessengerTest {
...
         // Now it has the relayer reward address.
-        assertEq(
-            teleporterMessenger.getRelayerRewardAddress(DEFAULT_DESTINATION_CHAIN_ID, mockMessageID),
-            relayerRewardAddress
+        assertEq(teleporterMessenger.getRelayerRewardAddress(mockMessageID), relayerRewardAddress);
+    }
+
+    function testZeroRelayerRewardAddress() public {
+        // Receive a message with a zero relayer reward address
+        uint256 mockNonce = 4343;
```
cam-schultz, line 38: I'm not sure how the VM state is handled under the hood, or if it persists from test to test. But if these tests rely on the mockNonces in the individual tests to be unique, then it seems pretty error prone. If that's the case, we should consider adding a utility global to all the tests to ensure uniqueness.

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: OTOH, requiring that all elements of the list are valid in order to send any receipts, especially when the list of pending receipts is always changing, is a less than ideal UI that may be difficult to integrate in an automated fashion. 

Could we instead silently drop invalid receipts, but still send the valid ones?

cam-schultz: (I recognize that this discussion isn't strictly in scope for this PR, so feel free to defer)

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -58,27 +59,23 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // provided by the deliverer of the uniquely identified message.
-    mapping(
-        bytes32 sourceBlockchainID => mapping(uint256 messageID => address relayerRewardAddress)
-    ) internal _relayerRewardAddresses;
+    // succeed at most once. The key is the message ID, and the value is the hash of the uniquely
+    // identified message whose execution failed.
+    mapping(bytes32 messageID => bytes32 messageHash) public receivedFailedMessageHashes;
+
+    // Tracks the message nonce for each message that has been received.
+    // Note: the `messageNonce` values are also used to determine if a given message has been received or not.
```
cam-schultz, line 73: Can you explicitly note the key and value, as is the case with the other mappings?

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -156,12 +156,11 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
      * Requirements:
      *
      * - `additionalFeeAmount` must be non-zero.
-     * - `message` must exist and not have been delivered yet.
+     * - `message` must exist and not have been acknowledge with a receipt yet.
```
cam-schultz, line 159:
```suggestion
     * - `message` must exist and not have been acknowledged with a receipt yet.
```

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -393,23 +383,38 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-        uint256[] calldata messageIDs,
+        bytes32[] calldata messageIDs,
         TeleporterFeeInfo calldata feeInfo,
         address[] calldata allowedRelayerAddresses
-    ) external senderNonReentrant returns (uint256) {
+    ) external senderNonReentrant returns (bytes32) {
+        // If the blockchain ID has yet to be initialized, no messages have ever been received by
+        // this contract, meaning that the receipt(s) will not be found in any event.
+        // Thus, don't need to initialize the blockchain ID here.
+        bytes32 blockchainID_ = blockchainID;
```
cam-schultz, line 393: Even so, I'm not a fan of selective treatment of a lazily initialized variable. Ideally, initialization would be behind an interface that completely abstracts the underlying variable state, and that variable would not be able to be directly accessed. We could probably implement something like that here, though that's probably more heavyweight than its worth. In the absence of enforcing that at the compiler level, I think we should fetch `blockchainID_` in a consistent way _everywhere_ it's used, even if we can guarantee it's initialized. That makes the contract much easier to reason about, even at a marginally higher gas cost for some methods. My 2c.

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -626,19 +651,33 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
         return messageID;
     }
 
+    /**
+     * @dev Marks a message as being received by storing the message nonce associated with the
+     * given message ID. The message must not be zero in order to be able to distinguish between
```
cam-schultz, line 656:
```suggestion
     * given message ID. The message nonce must not be zero in order to be able to distinguish between
```

tests/utils/utils.go
michaelkaplan13: Unless anyone feels strongly, I'm in favor of keeping it. I think it is general enough that it is worth having as a utility, and that way it's less error prone because `Expect(err).Should(BeNil())` doesn't need to be duplicated everywhere.

contracts/src/Teleporter/tests/GetRelayerRewardAddressTests.t.sol
michaelkaplan13: Each individual test gets run against a fresh contract (created by `setUp()`), so no state  is ever shared between multiple forge unit tests. Because of that, I don't think this is an issue.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I don't think this will be an issue either way really. 

The current state of the pending receipt queue doesn't affect `sendSpecifiedReceipts` here, because once a given message ID is received, its receipt will always be valid it to be sent via `sendSpecifiedReceipts`, whether or not the receipt has been sent in a message in the past. So automation shouldn't be too difficult. You can always check with `messageReceived` or `getRelayerRewardAddress` if a receipt is able to be sent for a message, and if it is, it will always be on a go forwards basis.

I think reverting in the event an invalid message ID is provided is more intuitive than silently dropping invalid ones, which would be likely to go unnoticed.

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I don't feel strongly either way here. 

The one observation I'd call out is that it's impossible for the blockchain ID to be initialize as a part of `sendSpecifiedReceipts`, `retryMessageExecution`, or `retrySendCrossChainMessage`, because if it has not previously been initialized, any calls to those functions will revert in any event, leaving the value uninitialized afterwards even if `initializeBlockchainID` is called.

I agree having a uniform access pattern everyone is less complicated and still correct OTOH though. Ultimately doesn't make a big difference to me as long as the comments make it clear.

geoff-vball (approved):
👍  LGTM!

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -43,42 +43,39 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
-    // provided by the deliverer of the uniquely identified message.
-    mapping(
-        bytes32 sourceBlockchainID => mapping(uint256 messageID => address relayerRewardAddress)
-    ) internal _relayerRewardAddresses;
+    // succeed at most once. The key is the message ID, and the value is the hash of the uniquely
+    // identified message whose execution failed.
+    mapping(bytes32 messageID => bytes32 messageHash) public receivedFailedMessageHashes;
+
+    // Tracks the message nonce for each message that has been received. The key is the message ID,
+    // and the value is the nonce value that was received as a part for that message.
```
bernard-avalabs, line 72: typo: "as a part for" -> "as a part of"

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -43,42 +43,39 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
...
+    // Note: the `messageNonce` values are also used to determine if a given message has been received or not.
```
bernard-avalabs, line 73: Should a slightly more extended explanation of the messageNonce be provided here? This might be the first point that someone comes across the term messageNonce.

contracts/src/Teleporter/TeleporterMessenger.sol
bernard-avalabs: I agree in general that we should avoid code-level optimizations for small gas savings if they hurt readability. It really should be the job of the compiler to perform these optimizations. With that said, because of the reverts that Michael pointed out, the functions will need comments about initialized/uninitialized variables no matter which way we go. So perhaps just having the comments is enough.

bernard-avalabs (commented):
Generally makes sense to me. I've left a few minor comments.

contracts/src/Teleporter/tests/GetRelayerRewardAddressTests.t.sol
cam-schultz: Cool, thanks for the explanation. I wasn't sure if the nonce values being different from test to test was pertinent or not.

minghinmatthewlam (dismissed):
LGTM 🚢

contracts/src/Teleporter/TeleporterMessenger.sol
michaelkaplan13: I think the comment above for `messageNonce` on lines 50-53 explain it in more detail, but happy to expand on it there if you think there's a piece that's missing.

minghinmatthewlam (dismissed):

contracts/src/Teleporter/TeleporterMessenger.sol
cam-schultz: > it's impossible for the blockchain ID to be initialize as a part of sendSpecifiedReceipts, retryMessageExecution, or retrySendCrossChainMessage

That's a good point. I'm ok with leaving as is and denoting this behavior with comments.

cam-schultz (dismissed):
LGTM!

cam-schultz (dismissed):

cam-schultz (approved):

minghinmatthewlam (approved):

Issue #199, opened 2023-12-13
Create Release Workflow


michaelkaplan13:
**Context and scope**
A Teleporter "release" primarily consists of a single raw EVM transaction that can be sent on any EVM chain to deploy the `TeleporterMessenger` contract with the same byte code to the same address. This is constructed using [Nick's method](https://yamenmerhi.medium.com/nicks-method-ethereum-keyless-execution-168a6659479c#:~), and can be built using the Golang scripts as described [here](https://github.com/ava-labs/teleporter/tree/main/utils/contract-deployment). 

When a Teleporter release is cut, there should be a github workflow that automatically constructs this transaction, and uploads it to the release artifacts. Along with the transaction, it should also include the deployer address (sender) of the transaction, and the resulting contract address.

There should also be a script within the repository source that can take a release version, RPC URL, and funded account key, and deploy `TeleporterMessenger` contract on that chain. It will first need to send the required amount of native tokens to the deployer address, then broadcast the raw transaction and confirm the result.

**Discussion and alternatives**
The repository on the whole also includes Golang components and packages that will likely be updated far more frequently than the `TeleporterMessenger` contract. These can likely be published just with tagged versions, and not full releases. 

Tag versions can use minor and patch version increments, and Teleporter contract releases should always use a major version increment. Ideally, the major version will align with the version numbers used in the primary `TeleporterRegistry` contracts.

PR #210, opened 2023-12-21
Teleporter release flow

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/199

Adds a release flow to the CI which will create a new release and attach the relevant artifacts when a new version tag is pushed.

Adds `deploy_teleporter.sh` script to deploy any teleporter release to any chain.

## How this works
The CI release flow is triggered on any pushed tag matching `v*.*.0` (major and minor version bumps). https://github.com/ava-labs/teleporter/issues/199 Discusses only triggering on major bumps, but with how golang treats major version bumps I think we may want to stick to minor version bumps for new versions of the TeleporterMessenger contract.

The release will contain a changelog of all the PRs merged since the last release, and will contain `.txt` files for the `deployer address`, `contract address`, and `deployer transaction`.

`deploy_teleporter.sh` requires a teleporter release version and an rpc url. It optionally takes a private key if they user wants to fund the deployer address directly from the script.

## How this was tested
Local network using a slightly modified `run.sh` that does not deploy the teleporter contract for one subnet.

## How is this documented
`--help` option in `deploy_teleporter.sh`


.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: softprops/action-gh-release@v1
+        with:
+          name: ${{ github.ref_name }}
+          tag_name: ${{ github.ref_name }}
+          prerelease: false
+          body: ${{ steps.build_changelog.outputs.changelog }}
+          files: |
+                ${{ steps.artifacts.outputs.tx }}
+                ${{ steps.artifacts.outputs.deployer }}
+                ${{ steps.artifacts.outputs.address }}
```
michaelkaplan13, line 72: My understanding here is that there are 3 files that will be uploaded, named:
- `TeleporterMessenger_DeployerTransaction_${{ github.ref_name }}.txt`
- `TeleporterMessenger_DeployerAddress_${{ github.ref_name }}.txt`
- `TeleporterMessenger_ContractAddress_${{ github.ref_name }}.txt`

Could we simplify the workflow by setting those variables up top somehow, rather than renaming the files to them above, outputing the file names as other variable names, and then referencing that output here?

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,92 @@
...
+function printHelp() {
+    echo "Usage: ./scripts/deploy_teleporter.sh [OPTIONS]"
+    echo "Deploys a selected TeleporterMessenger contract to the specified chain"
+    echo "For a list of releases, go to https://github.com/ava-labs/teleporter/releases"
+    printUsage
+}
+
+function printUsage() {
+    echo "Options:"
+    echo "  --fund-deployer <Private_key>    Optional. Funds the deployer address with the account held by <Private_key>"
```
michaelkaplan13, line 26:
```suggestion
    echo "  --fund-deployer <private_key>    Optional. Funds the deployer address with the account held by <private_key>"
```

.github/workflows/release.yml
geoff-vball: I could get rid of one step here, by replacing lines 70-72 here with just the expected filenames, and then we could get rid of the 3 `echo` lines above. 

We could also modify the `ConstructKeylessTransaction` function to take an optional `version` parameter, and just make it write out the filenames we want in the first place, eliminating the 3 `mv` lines.

michaelkaplan13: > I could get rid of one step here, by replacing lines 70-72 here with just the expected filenames, and then we could get rid of the 3 echo lines above.

Yeah I think that'd be great 🙏 Bonus points if we can set the 3 file names as variables at the workflow level rather than having them in 2 place also, but I'm not sure how that can be configured myself

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Create Artifacts
+        id: artifacts
+        run: |
+          go run utils/contract-deployment/contractDeploymentTools.go constructKeylessTx contracts/out/TeleporterMessenger.sol/TeleporterMessenger.json
+          
+          mv UniversalTeleporterDeployerTransaction.txt ${{ env.deployment_tx_fn }}
+          
+          mv UniversalTeleporterDeployerAddress.txt ${{ env.deployer_addr_fn }}
+          
+          mv UniversalTeleporterMessengerContractAddress.txt ${{ env.contract_addr_fn }}
```
michaelkaplan13, line 61:
```suggestion
          go run utils/contract-deployment/contractDeploymentTools.go constructKeylessTx contracts/out/TeleporterMessenger.sol/TeleporterMessenger.json
          mv UniversalTeleporterDeployerTransaction.txt ${{ env.deployment_tx_fn }}
          mv UniversalTeleporterDeployerAddress.txt ${{ env.deployer_addr_fn }}
          mv UniversalTeleporterMessengerContractAddress.txt ${{ env.contract_addr_fn }}
```

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,92 @@
...
+        *) 
+          echo "Invalid option: -$1" && printHelp && exit 1;;
+    esac
+    shift
+done
+
+# Tokens required to deploy the contract.
+# Equal to contractCreationGasLimit * contractCreationGasPrice
+# from utils/deployment-utils/deployment_utils.go
+gas_tokens_required=10000000000000000000 # 10^19 gwei
```
michaelkaplan13, line 70: Can this be shortened to `10ether`?

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,92 @@
...
+
+teleporter_contract_address=$(curl -sL https://github.com/ava-labs/teleporter/releases/download/$teleporter_version/TeleporterMessenger_ContractAddress_$teleporter_version.txt)
+echo "TeleporterMessenger $teleporter_version contract address: $teleporter_contract_address"
+teleporter_deployer_address=$(curl -sL https://github.com/ava-labs/teleporter/releases/download/$teleporter_version/TeleporterMessenger_DeployerAddress_$teleporter_version.txt)
+echo "TeleporterMessenger $teleporter_version deployer address: $teleporter_deployer_address"
+teleporter_deploy_tx=$(curl -sL https://github.com/ava-labs/teleporter/releases/download/$teleporter_version/TeleporterMessenger_DeployerTransaction_$teleporter_version.txt)
+
+
```
michaelkaplan13, line 78:
```suggestion
```

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,92 @@
...
+if [[ $(cast code $teleporter_contract_address --rpc-url $rpc_url) != "0x" ]]; then
+  echo "TeleporterMessenger $teleporter_version has already been deployed on this chain." && exit 1
+fi
+
+if [[ $user_private_key != "" ]]; then
```
michaelkaplan13, line 83: Thinking through how to make this the most robust so we don't have folks accidentally sending tokens where they don't need to. Could we make the steps:
1. Check if there is code at the expected Teleporter contract address. If there is, it's already deployed to this chain and we can exit early.
2. Check if there are sufficient funds at the Teleporter deployer address. If there are, we can skip the funding step, but if not, we should require a private key.
3. Publish the raw transaction.

geoff-vball: Yeah, we can. It's just really annoying to check the balance because I think bash treats arithmetic with 64-bit ints which only gets you up to ~9e18. I think there's a separate utility I can use to figure it out though. I'll do that when I get back this afternoon.

scripts/deploy_teleporter.sh
geoff-vball: It can, but `cast` returns balances in wei, so in order to compare them to check if we should fund or not, we need to keep it in wei

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,99 @@
...
+gas_tokens_required=10000000000000000000 # 10^19 wei = 10 eth
+
+teleporter_contract_address=$(curl -sL https://github.com/ava-labs/teleporter/releases/download/$teleporter_version/TeleporterMessenger_ContractAddress_$teleporter_version.txt)
+echo "TeleporterMessenger $teleporter_version contract address: $teleporter_contract_address"
+teleporter_deployer_address=$(curl -sL https://github.com/ava-labs/teleporter/releases/download/$teleporter_version/TeleporterMessenger_DeployerAddress_$teleporter_version.txt)
+echo "TeleporterMessenger $teleporter_version deployer address: $teleporter_deployer_address"
+teleporter_deploy_tx=$(curl -sL https://github.com/ava-labs/teleporter/releases/download/$teleporter_version/TeleporterMessenger_DeployerTransaction_$teleporter_version.txt)
+
+if [[ $(cast code --rpc-url $rpc_url $teleporter_contract_address) != "0x" ]]; then
+    echo "TeleporterMessenger $teleporter_version has already been deployed on this chain." && exit 1
```
michaelkaplan13, line 79: I think exit 0 would make more sense here, since it technically is a success case.

Would you also mind adding comments explaining each of the steps in this script here?

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,99 @@
...
+fi
+
+deployer_balance=$(cast balance --rpc-url $rpc_url $teleporter_deployer_address)
+
+if [[ $(echo "$deployer_balance>=$gas_tokens_required" | bc) == 1 ]]; then
+    echo "Deployer Address already funded"
```
michaelkaplan13, line 85:
```suggestion
    echo "Deployer address already funded"
```

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,99 @@
...
+else 
+    transfer_amount=$(echo "$gas_tokens_required-$deployer_balance" | bc)
+    if [[ $user_private_key == "" ]]; then
+        echo "No private key provided. Deployer address must be funded with $transfer_amount wei to deploy contract" && exit 1
+    fi
+    echo "Funding Deployer Address with $transfer_amount wei"
```
michaelkaplan13, line 91:
```suggestion
    echo "Funding deployer address with $transfer_amount wei"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+  push:
+    tags:
+      - "v*.*.0"
+
+jobs:
+  build_and_upload_artifacts:
+    name: Build and Upload Teleporter Artifacts
+    runs-on: ubuntu-20.04
+    env:
+      deployment_tx_fn: TeleporterMessenger_DeployerTransaction_${{ github.ref_name }}.txt
```
michaelkaplan13, line 16: Nit: Sorry I know this was a pre-existing file name, but could we change it to `DeploymentTransaction` rather than `DeployerTransaction`? 

I think that is more clear/descriptive.

michaelkaplan13 (dismissed):
LGTM. I don't know of a good way to test the release workflow without creating a release. Once merged, I recommend testing with a `v0.0.0` release that can be later removed once tested.

michaelkaplan13 (dismissed):

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+name: Release Teleporter
+
+on:
+  push:
+    tags:
+      - "v*.*.0"
```
cam-schultz, line 9: Why do we require the bugfix version to be `0` in order to issue a release? It's possible we'd release a fix that did not change the interface at all, in which case we'd want to tag it as a bugfix. 

I don't think we should use semver tags for non-contract changes. Those can either be rc's or left untagged.

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Create Artifacts
+        id: artifacts
+        run: |
+          go run utils/contract-deployment/contractDeploymentTools.go constructKeylessTx contracts/out/TeleporterMessenger.sol/TeleporterMessenger.json
+          mv UniversalTeleporterDeployerTransaction.txt ${{ env.deployment_tx_fn }}
+          mv UniversalTeleporterDeployerAddress.txt ${{ env.deployer_addr_fn }}
+          mv UniversalTeleporterMessengerContractAddress.txt ${{ env.contract_addr_fn }}
+
+      - name: Create release
```
cam-schultz, line 60: Is the zipped source code automatically included, even though it's not specified explicitly? That's the default behavior when creating a release through Github, so it would be good to include that in the automated flow as well.

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,103 @@
...
+set -e
+
+TELEPORTER_PATH=$(
+    cd "$(dirname "${BASH_SOURCE[0]}")"
+    cd .. && pwd
+)
+
+if ! command -v forge &> /dev/null; then
+    echo "forge not found, installing"
+    $TELEPORTER_PATH/scripts/install_foundry.sh
```
cam-schultz, line 14: I'd prefer to have the script exit here rather than installing foundry automatically. That way the deployer can control what foundry installation they wish to use (the main advantage of our fork is stable CI, but it's by no means a requirement to using Teleporter ourside of CI). We should print that `install_foundry.sh` can be used to install our fork before exiting.

.github/workflows/release.yml
geoff-vball: Yes it's already included automatically

scripts/deploy_teleporter.sh
geoff-vball: Yep, good call.

.github/workflows/release.yml
geoff-vball: We can still manually make releases for the whole repo for, let's say `v.0.1.2`, but it won't trigger this flow. This flow just adds in the teleporter contract artifacts, which we only want to do for a new version of the teleporter contract.

To treat semver properly, we would probably only generate the teleporter artifacts on major version changes, but with the way golang treats major version changes, I think we should stick with minor versions.

cam-schultz: > we would probably only generate the teleporter artifacts on major version changes

I think it's possible we'd generate Teleporter artifacts for major, minor, or bug fix versions. We should treat this as a Solidity repository primarily, and have the tagged releases map to Teleporter and the supporting contracts. Other changes, including the Golang tests and the example contracts, should not constitute tagged releases on their own IMO.

michaelkaplan13: With how Nick's method is used to deploy the contract to the same address on every chain, any change to the contract (even one that doesn't change the interface) would be a "breaking" change in the sense that the new contract will be deployed to a new address and not be able to send/receive message from older versions. 

I think it makes sense to leave bugfix versions for changes to the test and/or golang utilities that other repos may consume, and require that any change to the Teleporter contract itself be at least a new minor version. While the "teleporter version" numbers in the registry contracts are completely independent, I think we should only create new Teleporter contract releases when we plan on updating the commonly used registries via an off-chain warp message.

michaelkaplan13 (commented):
Can you add a section to the top level README documenting the usage of `deploy_teleporter.sh`? I think we should also include explanations regarding the major/minor/bugfix version usage for the repo, and make it clear that each version can only send messages to/from the same version on other chains.

geoff-vball:
@michaelkaplan13 We currently have a doc at `utils/contract-deployment/README.md`, linked from the toplevel README that describes how to do all the steps manually. Do you think I should replace that doc entirely? Or leave it and just add details about the deployment script?

michaelkaplan13:
> @michaelkaplan13 We currently have a doc at `utils/contract-deployment/README.md`, linked from the toplevel README that describes how to do all the steps manually. Do you think I should replace that doc entirely? Or leave it and just add details about the deployment script?

I'd be in favor of adding a new "Deploying `TeleporterMessenger` to a Subnet" section to the top level README, including how the need for the contract to be deployed to the same address on each chain and the intended usage of the new script.

.github/workflows/release.yml
cam-schultz: Sounds good to me. Can we document somewhere how we're planning on treating release versions since we're diverging from semver as a it relates to the Teleporter contract?

geoff-vball: @cam-schultz Yep I'll add it.

scripts/deploy_teleporter.sh
```diff
@@ -0,0 +1,102 @@
...
+
+set -e
+
+TELEPORTER_PATH=$(
+    cd "$(dirname "${BASH_SOURCE[0]}")"
+    cd .. && pwd
+)
+
+if ! command -v forge &> /dev/null; then
+    echo "forge not found. You can install by calling $TELEPORTER_PATH/scripts/install_foundry.sh" && exit 1
```
minghinmatthewlam, line 13: nit : in `abi_bindings.sh` we call the script to install if not installed already, do we want to do the same? 

If not, can we change wording to just "forge not found, install by calling ..."

geoff-vball: Yep, changed.

README.md
```diff
@@ -118,6 +118,21 @@ cp .env.example .env # Set proper values after copying.
...
+- `--rpc-url <url>` Required. Specify the rpc url of the node to use.
+- `--fund-deployer <private_key>`  Optional. Funds the deployer address with the account held by `<private_key>`
+
+To ensure that Teleporter can be deployed to the same address on every EVM based chain, it uses [Nick's Method](https://yamenmerhi.medium.com/nicks-method-ethereum-keyless-execution-168a6659479c) to deploy from a pre-determined `deployer address`. Teleporter costs exactly `10eth` in the subnet's native gas token to deploy, which must be sent to the `deployer address`. 
+
+`deploy_teleporter.sh` will send the necessary funds to the deployer address if it is provided with a private key to send the funds with. Alternatively, the `deployer address` can be funded externally. The `deployer address` for each version can be found by looking up the appropriate version at https://github.com/ava-labs/teleporter/releases and downloading `TeleporterMessenger_Deployer_Address_<VERSION>.txt`.
```
michaelkaplan13, line 134:
```suggestion
To ensure that Teleporter can be deployed to the same address on every EVM based chain, it uses [Nick's Method](https://yamenmerhi.medium.com/nicks-method-ethereum-keyless-execution-168a6659479c) to deploy from a static deployer address. Teleporter costs exactly `10eth` in the subnet's native gas token to deploy, which must be sent to the deployer address. 

`deploy_teleporter.sh` will send the necessary native tokens to the deployer address if it is provided with a private key for an account with sufficient funds. Alternatively, the deployer address can be funded externally. The deployer address for each version can be found by looking up the appropriate version at https://github.com/ava-labs/teleporter/releases and downloading `TeleporterMessenger_Deployer_Address_<VERSION>.txt`.
```

michaelkaplan13 (commented):
One super small nit on wording, but LGTM! 

I think it makes sense to keep the `fixed-width` text formatting for variable names or commands, so didn't think it made sense for `deployer address`.

michaelkaplan13 (approved):
🚀 🚢

cam-schultz (approved):

PR #211, opened 2023-12-21
OpenZeppelin defender fixes

### LINKED ISSUE #186: OpenZeppelin Defender initial fixes
**Context and scope**
The repo just integrated the OpenZeppelin Defender that adds additional security checks, and the first report [here](https://github.com/ava-labs/teleporter/pull/176#issuecomment-1846095512) has generated some low severity findings.

Certain alerts we purposely decided against, for example "Use custom errors". Also lots of alerts are from test files that we don't scrutinize as strictly as tests, but can also be addressed if wanted.

minghinmatthewlam:
## Why this should be merged
Fixes #186 

## How this works
Handles fixes from the most recent [report](https://defender.openzeppelin.com/v2/#/code/report-details/e20ca4ce-58e7-4d3d-93d8-01b91e6d190a) that are applicable:

- missing docstrings
- consistency of `msg.sender` vs `_msgSender` to support EIP2771 meta transactions
- additional indexed parameters i.e. `event UnlockTokens`
- lack of security contact
- adding `ReceiptReceived` event 
- literal number with many digits
- unused events
- named return variables
- write array length to stack for forloops to save gas
- variables are initialized to default values and uses more gas  i.e. `uint256 val = 0;`
- `_nativeMinter` can be made constant

## How this was tested
Unit + e2e test updates
CI

## How is this documented
TODO


michaelkaplan13 (commented):
👍 These are all great.

Left one super small nit, and one comment about how we'll need to update the E2E test that is failing.

tests/flows/basic_send_receive.go
```diff
@@ -94,7 +95,17 @@ func BasicSendReceive(network interfaces.Network) {
...
 	//
-	network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, true)
+	deliveryReceipt = network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, true)
+
+	// Check that the receipt was received for expected Teleporter message id
+	receiptEvent, err := utils.GetEventFromLogs(
+		deliveryReceipt.Logs,
+		subnetAInfo.TeleporterMessenger.ParseReceiptReceived)
+	Expect(err).Should(BeNil())
+	Expect(receiptEvent.MessageID).Should(Equal(expectedReceiptID))
```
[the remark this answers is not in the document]
michaelkaplan13: We have a semi-hacky workaround for this in the send specific receipts test [here](https://github.com/ava-labs/teleporter/blob/9e3880d32c9c4082811363a0efa68f28e4b64fed/tests/flows/send_specific_receipts.go#L28), which involves sending messages to clear the receipt queues prior to proceeding with a test. This is only done in the case of a local network, which is ensure to have a reasonable receipt queue size.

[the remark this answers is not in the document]
minghinmatthewlam: added a check for the `ReceiptReceived` event emission

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -39,7 +39,7 @@ contract NativeTokenDestination is ITeleporterReceiver, INativeTokenDestination,
     // they can be tracked separately.
     address public constant BURN_FOR_TRANSFER_ADDRESS = 0x0100000000000000000000000000000000000001;
 
-    INativeMinter private immutable _nativeMinter =
+    INativeMinter public constant NATIVE_MINTER =
```
geoff-vball, line 42: Definitely agree with the change to constant vs immutable here. Is there a gas-usage benefit the making it public?

geoff-vball (approved):

bernard-avalabs (dismissed):
LGTM

contracts/src/Teleporter/ReceiptQueue.sol
```diff
@@ -49,32 +49,34 @@ library ReceiptQueue {
...
+        return receipt;
     }
 
     /**
      * @dev Returns the outstanding receipts for the given chain ID that should be included in the next message sent.
      */
     function getOutstandingReceiptsToSend(TeleporterMessageReceiptQueue storage queue)
         internal
-        returns (TeleporterMessageReceipt[] memory result)
+        returns (TeleporterMessageReceipt[] memory)
```
cam-schultz, line 67: I agree that we should remove named return parameters for the sake of readability, but I'm curious if there are any gas benefits one way or the other? The OZ report does not specify.

cam-schultz (dismissed):
LGTM, just one question about named return parameters.

tests/flows/basic_send_receive.go
```diff
@@ -94,7 +95,17 @@ func BasicSendReceive(network interfaces.Network) {
...
 	// Relay the message to the destination
 	//
-	network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, true)
+	deliveryReceipt = network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, true)
+
+	// Check that the receipt was received for expected Teleporter message ID
+	// This check is not performed for external networks because unrelated messages may have already changed
+	// the state of the receipt queues.
+	if !network.IsExternalNetwork() {
+		Expect(utils.CheckReceiptReceived(
```
michaelkaplan13, line 104: In order to not depend on the order that tests run, I think we should clear the receipt queue at the start of this flow. Otherwise, other tests may send messages that cause the receipt queue length to be >5 at the start of this flow, and the first message sent back may not have this specific receipt.

michaelkaplan13:
Is the OpenZeppelin submodule version used easy to update here to close this item?

image

michaelkaplan13:
I think this item is small enough to address in this PR also:
 
image

geoff-vball:
@michaelkaplan13 our openzepplin submode is at version `v4.8.1` and `subnet-evm` is using `v4.9.3`

michaelkaplan13:
> @michaelkaplan13 our openzepplin submode is at version `v4.8.1` and `subnet-evm` is using `v4.9.3`

Ah, thanks. I should have checked myself. I was assuming it was a more recent fix published for that issue. 

Now I'm confused why that OpenZeppelin report item is listed at all then. Do we think it's a pure false positive?

geoff-vball:
@michaelkaplan13 Actually, `https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable` is a separate repo. I don't see anywhere that a specific version is being included or even used anywhere though.

contracts/src/Teleporter/ReceiptQueue.sol
minghinmatthewlam: from the gas reports didn't notice gas changes, think that makes sense since under the hood the named return parameters are likely the same as declaring a new local variable.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: no gas benefits, I'm going to change it back to private.

tests/flows/basic_send_receive.go
minghinmatthewlam: moved the `ClearReceiptQueue` into a utils function and call in flow if not an external network

minghinmatthewlam:
> I think this item is small enough to address in this PR also:
> 
> image

only the first two apply, updated.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: nvm updated back to public, to keep in convention with our other Warp declarations

minghinmatthewlam:
> Member

This looks to be a false positive. If you click into the link attached in the alert https://github.com/advisories/GHSA-9c22-pwxw-p6hx, it specifies `@openzeppelin/contracts` and `@openzeppelin/contracts-upgradeable`. We only use the first package, and as @geoff-vball pointed out the versions we use are >v4.4.1

michaelkaplan13 (approved):

cam-schultz (dismissed):

geoff-vball (approved):

michaelkaplan13 (approved):

PR #215, opened 2023-12-22
Fix example scripts

cam-schultz:
## Why this should be merged
Fixes two bugs in the example scripts:
- `cast send` results return numerical values in the form `1000 [1e3]`. We need to strip the raw number before using it.
- Some C-Chain and Subnet A variables were overloaded.

This PR also expands the contract deployment README to include complete steps for deploying Teleporter, not just creating the keyless transaction params

## How this works
See above

## How this was tested
Ran run.sh

## How is this documented
Self explanatory

### DIFF
--- a/utils/contract-deployment/README.md
+++ b/utils/contract-deployment/README.md
@@ -24,7 +24,7 @@ Now that the keyless transaction is constructed, fund the deployer address. For 
 
 ```bash
 teleporter_deploy_address=$(cat UniversalTeleporterDeployerAddress.txt)
-cast send --private-key $my_private_key --value 50ether $teleporter_deploy_address --rpc-url $my_rpc_url
+cast send --private-key $my_private_key --value 10ether $teleporter_deploy_address --rpc-url $my_rpc_url
 ```
 
 Then, deploy Teleporter by sending the keyless transaction:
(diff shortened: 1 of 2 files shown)


geoff-vball (dismissed):

utils/contract-deployment/README.md
```diff
@@ -17,4 +17,21 @@ OR
...
 ## Results
-The resulting raw transaction, `TeleporterMessenger` contract address, and universal deployer address are written to standard output, as well as to `UniversalTeleporterDeployerTransaction.txt`, `UniversalTeleporterMessengerContractAddress.txt`, and `UniversalTeleporterDeployerAddress.txt` respectively.
\ No newline at end of file
+The resulting raw transaction, `TeleporterMessenger` contract address, and universal deployer address are written to standard output, as well as to `UniversalTeleporterDeployerTransaction.txt`, `UniversalTeleporterMessengerContractAddress.txt`, and `UniversalTeleporterDeployerAddress.txt` respectively.
+
+## Deploy the contract
+Now that the keyless transaction is constructed, fund the deployer address. For example, using `cast`:
+
+```bash
+teleporter_deploy_tx=$(cat UniversalTeleporterDeployerTransaction.txt)
```
michaelkaplan13, line 26: I think this should be reading in `$teleporter_deploy_address`, not the transaction

utils/contract-deployment/README.md
```diff
@@ -17,4 +17,21 @@ OR
...
+cast send --private-key $my_private_key --value 50ether $teleporter_deploy_address --rpc-url $my_rpc_url
```
michaelkaplan13, line 27: The current keyless transaction uses 10 native tokens for gas in all cases, so I'd recommend updating this to `10ether` so it's not larger than necessary.

michaelkaplan13 (approved):
:+1:

geoff-vball (approved):

PR #218, opened 2024-01-02
Reorder execution on receive

minghinmatthewlam:
## Why this should be merged
We want to keep a standard of having external calls at the end of functions when possible, and after state changes. This can help narrow down the scope of possible reentrancy attacks. No known vulnerability with the previous approach, but the current change makes it not possible to call an extra `addFeeAmount` when receiving a message in `receiveCrossChainMessage`.

## How this works
Moves external call in message execution to end of `receiveCrossChainMessage`

## How this was tested
Unit test updates
CI

## How is this documented
N/A


cam-schultz (approved):

geoff-vball (approved):

michaelkaplan13:
> but the current change makes it not possible to call an extra `addFeeAmount` when receiving a message in `receiveCrossChainMessage`.

Want to clarify that it's still possible to call `addFeeAmount` when receiving a message in `receiveCrossChainMessage` generally, but it is now not possible to call it to add a fee amount for a message whose receipt is contained in the current message itself (because it is known to have already been delivered prior to the message execution now).

michaelkaplan13 (approved):

bernard-avalabs (approved):

Issue #213, opened 2023-12-22
Add reentrancy guard to TeleporterUpgrade.receiveTeleporterMessage


minghinmatthewlam:
**Context and scope**
Currently in `TeleporterMessenger` we have a `receiverNonReentrant` guard, that prevents the same Teleporter instance from receiving and executing multiple messages in one call. But inside `TeleporterUpgradeable.receiveTeleporterMessage` the dapp can allow different Teleporter versions to deliver messages, and there's no reentrancy guard between the two versions.

**Discussion and alternatives**
Add a reentrancy guard to default implementation of `TeleporterUpgradeable.receiveTeleporterMessage`. Add unit tests for same Teleporter instance, and also different Teleporter instances delivering messages.

**Open questions**
Should we mark the external `receiveTeleporterMessage` implementation as virtual as well?

PR #224, opened 2024-01-05
Make TeleporterUpgradeable receiveTeleporterMessage nonreentrant

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/213

Implements a reentrancy guard for `receiveTeleporterMessage` in `TeleporterUpgradeable`. `TeleporterMessenger.receiveCrossChainMessage` already has a reentrancy guard, but this will prevent `receiveTeleporterMessage` from being called from two different `TeleporterMessenger` contracts in the same message execution. 

## How this works
Note: forge's `mockCall` and `expectRevert` don't work for low-level calls, so we can't check the revert of the internal teleporter message execution in the tests.

## How this was tested
Unit tests

## How is this documented


contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -52,6 +56,17 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver {
      */
     event TeleporterAddressUnpaused(address indexed teleporterAddress);
 
+    // upgradeableNonReentrant modifier makes sure we cannot reenter between receiver calls.
+    // This modifier should be used for TeleporterUpgradeable contracts so that the execution of a
+    // call from one TeleporterMessenger contract cannot call the ReceiveCrossChainMessage function
+    // of a different TeleporterMessenger contract version that then calls into the same contract.
+    modifier upgradeableNonReentrant() {
```
minghinmatthewlam, line 63: can you clarify why we need a separate reentrancy guard here and not OpenZeppelin's?

geoff-vball: My understanding is that any non-reentrant implementation applies to every function of that contract, so if we used OZ's reentrant guard here and someone inherits from `TeleporterUpgradeable` and puts the OZ `nonReentrant` on, say, `_receiveTeleporterMessage`, it will always revert because the `_ENTERED` flag will be set when you enter `_receiveTeleporterMessage`.

bernard-avalabs (commented):
Generally makes sense to me. However, if I understand the following correctly:

"... so if we used OZ's reentrant guard here and someone inherits from TeleporterUpgradeable and puts the OZ nonReentrant on, say, _receiveTeleporterMessage, it will always revert because the _ENTERED flag will be set when you enter _receiveTeleporterMessage."

it seems odd that we are implementing our own reentrancy guard in order to prevent a conflict for a inheriting contract that also uses OZ's reentrancy guard. I'm surprised there isn't a standard pattern to handle this case that doesn't require an alternate reentrancy guard implementation.

geoff-vball:
So after doing some research, it seems like everyone I can find uses reentrancy guard solely as a protection against malicious contracts, NOT against coder error.

The prevailing wisdom seems to be to use the OZ reentrancy guard on any external function that you wouldn't want to be reentered externally. If you want to call a function internally that you have a reentrancy guard on, you should make an internal version that contains the functionality, and create an external version with the reentrancy guard that just calls the internal version.

Moral of the story: Just don't put any bugs in your code :)

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,154 @@
...
+    ITeleporterMessenger,
+    TeleporterMessage,
+    TeleporterMessageReceipt
+} from "../../ITeleporterMessenger.sol";
+import {WarpMessage} from "@subnet-evm-contracts/interfaces/IWarpMessenger.sol";
+import {TeleporterMessenger} from "../../TeleporterMessenger.sol";
+
+uint32 constant warpMessageIndex = 2;
+
+contract NonreentrantUpgradeableApp is TeleporterUpgradeable {
```
michaelkaplan13, line 20:
```suggestion
contract NonReentrantUpgradeableApp is TeleporterUpgradeable {
```

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,154 @@
...
+            sourceChainID: DEFAULT_ORIGIN_BLOCKCHAIN_ID,
+            originSenderAddress: teleporterAddress,
+            payload: abi.encode(messageToReceive)
+        });
+
+        // Same index as in NonreentrantUpgradeableApp._receiveTeleporterMessage()
+        _mockGetVerifiedWarpMessage(warpMessageIndex, warpMessage, true);
+
+        bytes32 messageId = TeleporterMessenger(teleporterAddress).calculateMessageID(
+            DEFAULT_ORIGIN_BLOCKCHAIN_ID, MOCK_BLOCK_CHAIN_ID, 987
```
michaelkaplan13, line 99:
```suggestion
            DEFAULT_ORIGIN_BLOCKCHAIN_ID, MOCK_BLOCK_CHAIN_ID, messageToReceive.messageNonce
```

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,154 @@
...
+        WarpMessage memory warpMessage = WarpMessage({
+            sourceChainID: DEFAULT_ORIGIN_BLOCKCHAIN_ID,
+            originSenderAddress: teleporterAddress,
+            payload: abi.encode(messageToReceive)
+        });
+
+        // Same index as in NonreentrantUpgradeableApp._receiveTeleporterMessage()
+        _mockGetVerifiedWarpMessage(warpMessageIndex, warpMessage, true);
+
+        bytes32 messageId = TeleporterMessenger(teleporterAddress).calculateMessageID(
```
michaelkaplan13, line 98:
```suggestion
        bytes32 messageID = TeleporterMessenger(teleporterAddress).calculateMessageID(
```

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,154 @@
...
+
+        vm.expectCall(
+            address(teleporterV2),
+            abi.encodeCall(
+                ITeleporterMessenger.receiveCrossChainMessage, (warpMessageIndex, address(app))
+            )
+        );
+
+        bytes32 messageId = TeleporterMessenger(teleporterV2).calculateMessageID(
+            DEFAULT_ORIGIN_BLOCKCHAIN_ID, MOCK_BLOCK_CHAIN_ID, 987
```
michaelkaplan13, line 143:
```suggestion
            DEFAULT_ORIGIN_BLOCKCHAIN_ID, MOCK_BLOCK_CHAIN_ID, messageToReceive.messageNonce
```

michaelkaplan13 (commented):
Just capitalization nits, but LGTM

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,154 @@
...
+import {TeleporterRegistryTest} from "./TeleporterRegistryTests.t.sol";
+import {
+    ITeleporterMessenger,
+    TeleporterMessage,
+    TeleporterMessageReceipt
+} from "../../ITeleporterMessenger.sol";
+import {WarpMessage} from "@subnet-evm-contracts/interfaces/IWarpMessenger.sol";
+import {TeleporterMessenger} from "../../TeleporterMessenger.sol";
+
+uint32 constant warpMessageIndex = 2;
```
minghinmatthewlam, line 18: any specific reason to declare outside of contract? Can we move this inside

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// The flow for the tests below is as follows:
+// NonreentrantUpgradeableApp::receiveTeleporterMessage ->
+// NonreentrantUpgradeableApp::_receiveTeleporterMessage ->
+// TeleporterMessenger::receiveCrossChainMessage ->
+// NonreentrantUpgradeableApp::receiveTeleporterMessage
+// The last step should revert because receiveTeleporterMessage (contained in
+// TeleporterUpgradeable) is non-reentrant.
+contract TeleporterNonreentrantTest is TeleporterUpgradeableTest {
```
minghinmatthewlam, line 53:
```suggestion
contract NonReentrantTest is TeleporterUpgradeableTest {
```

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,150 @@
...
+    address public constant DEFAULT_DESTINATION_ADDRESS = 0xd54e3E251b9b0EEd3ed70A858e927bbC2659587d;
```
minghinmatthewlam, line 54: is this used anywhere?

contracts/src/Teleporter/upgrades/tests/NonReentrantTests.t.sol
```diff
@@ -0,0 +1,149 @@
...
+        vm.expectEmit(true, true, true, true, address(teleporterAddress));
+        emit MessageExecutionFailed(messageID, DEFAULT_SOURCE_BLOCKCHAIN_ID, messageToReceive);
+
+        vm.prank(teleporterAddress);
+        nonReentrantApp.receiveTeleporterMessage(
+            DEFAULT_SOURCE_BLOCKCHAIN_ID, DEFAULT_ORIGIN_ADDRESS, DEFAULT_MESSAGE
+        );
+    }
+
+    function testNonreentrantDifferentTeleporter() public {
```
michaelkaplan13, line 105:
```suggestion
    function testNonReentrantDifferentTeleporter() public {
```

michaelkaplan13 (dismissed):

minghinmatthewlam (approved):

michaelkaplan13 (approved):

PR #226, opened 2024-01-07
Pre-increment message nonce

michaelkaplan13:
## Why this should be merged
Per Louis' idea and suggestion, instead of initializing `messageNonce` to 1, have it be default initialized to 0, and increment it before assigning the message nonce to use for a new message to be sent. While not a correctness issue, this makes `messageNonce` accurately reflect the total number of message sent at any point in time. As previously implemented, a more accurate name would have been `nextMessageNonce`.

## How this works
Uses pre-increment (i.e. `messageNonce_ = ++messageNonce`) instead of using the current `messageNonce` value for a new message to be sent, and then incrementing it afterwards.

## How this was tested
Unit tests, CI

## How is this documented
N/A


geoff-vball (dismissed):

contracts/src/Teleporter/TeleporterMessenger.sol
```diff
@@ -51,7 +51,7 @@ contract TeleporterMessenger is ITeleporterMessenger, ReentrancyGuards {
     // A monotonically incremented integer tracking the total number of messages sent by this TeleporterMessenger contract.
     // Used to provide uniqueness when generating message IDs for new messages. Initially starts at 1 such that the
```
cam-schultz, line 52: We should update this comment to destinguish between initialization and first usage of this variable. Perhaps something like: "The first message sent will have a `messageNonce` of 1 such that the nonce value can be used to provide replay protection"

cam-schultz (commented):
Makes sense to me, just one suggestion to update a comment.

minghinmatthewlam (commented):
LGTM, just Cam's comment about updating the doc

cam-schultz (approved):

PR #235, opened 2024-01-17
fix: display right contract addr in script

Nuttymoon:
## Why this should be merged

This PR brings minor fixes to the `deploy_teleporter.sh` script:
- Separating the `cast code` from the `if` statement allows for better error handling. For example, a 404 error will end up printing `TeleporterMessenger $teleporter_version has already been deployed on this chain.` (what is false)
- Display the right address in the final print

## How this works

N/A

## How this was tested

By deploying Teleporter to an EVM chain on a local network.

## How is this documented

N/A


geoff-vball (approved):
Thanks for this!

michaelkaplan13 (approved):
🙏 nice catch and thanks for fixing it

PR #236, opened 2024-01-18
typo

meaghanfitzgerald:
## Why this should be merged

fixed typo

## How is this documented


geoff-vball (approved):
Thanks for the fix Meaghan! :)

PR #240, opened 2024-01-18
Teleporter registry e2e test

### LINKED ISSUE #221: Integration Tests for Registry Updates
**Context and scope**
Now that off-chain Warp message support has been added to `coreth` and `subnet-evm` ([PR here](https://github.com/ava-labs/subnet-evm/pull/1028)), we should include Ginkgo integration tests of the relevant registry update flows.

Possible test cases include:
- Deploying new `TeleporterMessenger` version by modifying the Nick's method transaction such that the same contract byte code will be deployed to a different address.
- Adding the new contract address to the registry on one chain but not another. Check that messages are sent using the new version, but can't be received. 
- Check that messages sent from the other chain still use the old version and are able to be received.
- Add the contract address on the second chain, check that previously sent messages are now able to be received.
- Update the minimum required version for a dApp, ensure messages sent using the older version are no longer able to be received.

minghinmatthewlam:
## Why this should be merged
Adds an e2e test for testing the `TeleporterRegistry`, fixes #221.

## How this works
Tests:
- Setting Warp off-chain messages in the chain config and restarting nodes
- Adding new protocol version to `TeleporterRegistry` with signed Warp off-chain message
- Updating `minTeleporterVersion` of a dapp that uses the registry
- Makes sure receiving fails if Teleporter version is not `>= minTeleporterVersion`
- When both chains have the new deployed Teleporter, but only subnet B has it registered in registry. If we send from chain with registered new version, message will deliver to destination Teleporter, but message execution fails and destination dapp does not receive.
  - After subnet A registers new version, subnet B can deliver to dapp on subnet A again with newer Teleporter version

## How this was tested
Steps
```
	// Deploy dapp on both chains that use Teleporter Registry
	// Deploy version 2 of Teleporter to both chains
	// Construct AddProtocolVersion txs for both chains
	// Send tx for one of the two chains, verify events emitted
	// Send a message from chain with old version to chain with new version, verify message is received
	// Update minTeleporterVersion on chain with new version
	// Send same message from chain with old version, verify message is not received
	// Send a message from chain with new version to chain with old version, verify message is not received
	// Send tx with new protocol version to other chain, verify events emitted
	// Retry the previously failed message execution, verify message is now able to be delivered to dapp
```
CI

## How is this documented
test comments


minghinmatthewlam:
The latest `subnet-evm` release does not contain off-chain message changes, so will wait for a new release before merging. Currently running with a separate branch off `master`

utils/deployment-utils/deployment_utils.go
```diff
@@ -188,3 +188,10 @@ func ConstructKeylessTransaction(
 	}
 	return contractCreationTxBytes, senderAddress, contractAddress, nil
 }
+
+// AlterNicksTransaction increments the gas price of the transaction used to deploy the contract.
```
cam-schultz, line 192: I think this would be better in the testing utils, since we wouldn't ever use this function in a real deployment.

abi-bindings/go/Teleporter/upgrades/TeleporterRegistry/packing.go
```diff
@@ -0,0 +1,81 @@
...
+
+var teleporterRegistryEntryType abi.Type
+
+type TeleporterRegistryEntry struct {
+	Entry              ProtocolRegistryEntry `json:"protocolRegistryEntry"`
+	DestinationAddress common.Address        `json:"destinationAddress"`
+}
+
+func init() {
+	// Create an ABI binding for TeleporterRegistryEntry, defined in TeleporterRegistry.sol
```
cam-schultz, line 22: Is `TeleporterRegistryEntry` an actual type defined in the contract? I think what we're packing here is the packed `ProtocolRegistryEntry` and `DestinationAddress` [here](https://github.com/ava-labs/teleporter/blob/main/contracts/src/Teleporter/upgrades/TeleporterRegistry.sol#L101).

I think we should define the ABI spec for `protocolRegistryEntryType` to be consistent with the actual contract type, then add helpers to extend to `teleporterRegistryEntryType`

tests/flows/teleporter_registry.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/ava-labs/teleporter/tests/utils"
+	deploymentUtils "github.com/ava-labs/teleporter/utils/deployment-utils"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/log"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	networkID              uint32 = 1337
```
cam-schultz, line 25: We can use the default local network ID defined elsewhere in this PR

tests/flows/teleporter_registry.go
```diff
@@ -0,0 +1,221 @@
...
+		network,
+		subnetBInfo,
+		exampleMessengerB,
+		subnetAInfo,
+		exampleMessengerContractA,
+		exampleMessengerA,
+		fundedKey,
+		false)
+
+	utils.AddProtocolVersionAndWaitForAcceptance(
```
cam-schultz, line 134: Can you add comments to each of these statements mapping them to the test steps described at the beginning of the test function?

tests/flows/teleporter_registry.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/ethereum/go-ethereum/log"
+	. "github.com/onsi/gomega"
+)
+
+const (
+	networkID              uint32 = 1337
+	teleporterByteCodeFile        = "./contracts/out/TeleporterMessenger.sol/TeleporterMessenger.json"
+)
+
+func TeleporterRegistry(network interfaces.LocalNetwork) {
```
cam-schultz, line 29: IIUC, this test only updates the Teleporter version for subnets A and B. Won't that leave the subnets incompatible with subnet C and the C-Chain for any tests that run after this? 

In any case, can you add a comment to the function description explaining the state changes made by this test? It seems possible that a future test may assume a certain Teleporter address, which would change after this test.

cam-schultz (commented):
Overall makes sense to me. Thanks for expanding the utils library as well. Left a handful of relatively minor comments and some questions.

tests/testnet/network.go
```diff
@@ -52,8 +56,9 @@ var _ interfaces.Network = &testNetwork{}
 
 type testNetwork struct {
 	teleporterContractAddress common.Address
-	primaryNetwork            interfaces.SubnetTestInfo
-	subnets                   []interfaces.SubnetTestInfo
+	primaryNetwork            *interfaces.SubnetTestInfo
+	subnetIDs                 []ids.ID
```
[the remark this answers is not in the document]
minghinmatthewlam: removed

tests/flows/teleporter_registry.go
minghinmatthewlam: i'm leaving a constant `defaultNetworkID` field in this test, and left a TODO when we update `avalanche-network-runner` we can directly use the constant from there

tests/local/network.go
```diff
@@ -274,19 +276,17 @@ func (n *LocalNetwork) DeployTeleporterContracts(
...
 
 	ctx := context.Background()
 
-	subnets := n.getAllSubnetsInfo()
+	subnets := n.GetAllSubnetsInfo()
 	for _, subnetInfo := range subnets {
 		// Fund the deployer address
 		{
-			fundAmount := big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(10)) // 10eth
+			fundAmount := big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(11)) // 10eth
```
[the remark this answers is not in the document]
minghinmatthewlam: this is an AVAX amount. Needed an increase because with the new Teleporter Nick's tx we increase the gas price

scripts/install_avalanchego_release.sh
```diff
@@ -109,10 +119,8 @@ else
     # if the build-directory doesn't exist, build avalanchego
     if [[ ! -d ${BUILD_DIR} ]]; then
       echo "building avalanchego ${COMMIT} to ${BUILD_DIR}"
-      ./scripts/build.sh
       mkdir -p ${BUILD_DIR}
-
-      mv ${GIT_CLONE_PATH}/build/* ${BUILD_DIR}/
+      ./scripts/build.sh ${BUILD_DIR}/subnet-evm
```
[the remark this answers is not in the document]
minghinmatthewlam: removed

scripts/install_avalanchego_release.sh
```diff
@@ -47,6 +47,13 @@ extract_archive() {
   fi
 }
 
+extract_commit() {
+  if [[ $AVALANCHEGO_VERSION == *-* ]]; then
+      # Extract the substring after the last '-'
+      AVALANCHEGO_VERSION=${AVALANCHEGO_VERSION##*-}
```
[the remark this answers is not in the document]
minghinmatthewlam: changed it so that it first tries to checkout the tag/version, then if it fails it'll do another attempt by removing hyphen

scripts/install_subnetevm_release.sh
```diff
@@ -45,6 +45,13 @@ extract_archive() {
   fi
 }
 
+extract_commit() {
+  if [[ $SUBNET_EVM_VERSION == *-* ]]; then
+      # Extract the substring after the last '-'
+      SUBNET_EVM_VERSION=${SUBNET_EVM_VERSION##*-}
```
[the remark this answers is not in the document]
minghinmatthewlam: same as above

tests/flows/teleporter_registry.go
minghinmatthewlam: added

abi-bindings/go/Teleporter/upgrades/TeleporterRegistry/packing.go
minghinmatthewlam: updated to match `protocolRegistryEntryType` and chagned function as a helper `PackTeleporterRegistryWarpPayload`

tests/flows/teleporter_registry.go
minghinmatthewlam: yup that's a good point. Right now after this test runs the C-Chain's `TeleporterRegistry` will be behind on a Teleporter version. The reason tests are still okay for now is because the tests use C-Chain as source mostly, so the destination dapp would still receive from an older version. 

Currently the `subnet-evm` release is not compatible with a `coreth` that has Warp off-chain, will need to update this after so that all subnets match in registered Teleporter versions.

tests/flows/teleporter_registry.go
```diff
@@ -0,0 +1,233 @@
...
+
+	return unsignedMessage
+}
+
+func deployNewTeleporterVersion(
+	ctx context.Context,
+	network interfaces.LocalNetwork,
+	fundedKey *ecdsa.PrivateKey,
+) common.Address {
+	deploymentUtils.AlterNicksTransaction()
```
geoff-vball, line 213: We shouldn't have a single test modifying a global variable like this. We should modify `ConstructKeylessTransaction` to take a gas price and gas limit as params. The current global vars should be set as defaults, and then we can pass different values in from this function.

If we want to be able to call this function multiple times, we can use a variable local to this file to increment as a counter to know what gas price to use.

tests/testnet/network.go
```diff
@@ -52,8 +56,8 @@ var _ interfaces.Network = &testNetwork{}
 
 type testNetwork struct {
 	teleporterContractAddress common.Address
-	primaryNetwork            interfaces.SubnetTestInfo
-	subnets                   []interfaces.SubnetTestInfo
+	primaryNetwork            *interfaces.SubnetTestInfo
+	subnetsInfo               map[ids.ID]*interfaces.SubnetTestInfo
```
geoff-vball, line 60: I'm fairly ambivalent, but I think keeping this is a slice and just implementing a getter method on the `testNetwork` struct that takes an `ids.ID` and returns a `SubnetTestInfo` or nil would be slightly cleaner.

tests/utils/utils.go
```diff
@@ -323,6 +322,79 @@ func CreateReceiveCrossChainMessageTransaction(
...
+	signedWarpMsg := network.GetSignedMessage(ctx, source, destination, unsignedMessage.ID())
+	log.Info("Got signed warp message", "messageID", signedWarpMsg.ID())
+
+	// Construct tx to add protocol version and send to destination chain
+	signedTx := CreateAddProtocolVersionTransaction(
+		ctx,
+		signedWarpMsg,
+		source.TeleporterRegistryAddress,
+		senderKey,
+		source)
```
geoff-vball, line 379: nit: close parentheses on next line

tests/utils/utils.go
```diff
@@ -323,6 +322,79 @@ func CreateReceiveCrossChainMessageTransaction(
...
+		gasFeeCap,
+		gasTipCap,
+		big.NewInt(0),
+		callData,
+		types.AccessList{},
+		warp.ContractAddress,
+		signedMessage.Bytes(),
+	)
+
+	return SignTransaction(destinationTx, senderKey, subnetInfo.EVMChainID)
```
geoff-vball, line 358: I feel like this function could also call SendTransactionAndWaitForSuccess and return the receipt.

tests/utils/utils.go
```diff
@@ -800,3 +872,81 @@ func CalculateMessageID(
...
+	sourceExampleMessenger *examplecrosschainmessenger.ExampleCrossChainMessenger,
+	destination interfaces.SubnetTestInfo,
+	destExampleMessengerAddress common.Address,
+	destExampleMessenger *examplecrosschainmessenger.ExampleCrossChainMessenger,
+	senderKey *ecdsa.PrivateKey,
+	expectSuccess bool,
+) {
+	// Generate a random message
+	randomNumber := rand.Intn(1000)
+	randomMessage := fmt.Sprintf("%d", randomNumber)
```
geoff-vball, line 889: nit: I think it's good practice that anytime you use a random number in a test, to print that random number to the screen, in case some random inputs are causing the test to fail, otherwise it might be very hard to track down.

cam-schultz: I'd say we should avoid random numbers altogether. AFAIK there's no need for these messages to be unique, so it might be acceptable to simply send the same value every time. If not, we should derive the value deterministically.

tests/utils/utils.go
minghinmatthewlam: going to leave this, similar to `CreateReceiveCrossChainMessageTransaction`, `CreateSendCrossChainMessageTransaction`, etc.

scripts/install_avalanchego_release.sh
```diff
@@ -87,17 +94,25 @@ else
...
 
-    # if it's not a branch, try to checkout the commit
+    # if it's not a branch, try to checkout the commit 
     if [[ $CHECKOUT_STATUS -ne 0 ]]; then
       set +e
       git checkout ${AVALANCHEGO_VERSION} > /dev/null 2>&1
       CHECKOUT_STATUS=$?
       set -e
 
       if [[ $CHECKOUT_STATUS -ne 0 ]]; then
```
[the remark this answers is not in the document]
minghinmatthewlam: updated

tests/utils/utils.go
minghinmatthewlam: removed using random number, and instead pass in a message

tests/flows/teleporter_registry.go
minghinmatthewlam: I like that we use the same registry and have random order of e2e tests, so we can that the after updating the registry across all chains, the other e2e tests all still work. Will just need to add the new protocol version to C-chain after a new compatible release is made.

tests/flows/teleporter_registry.go
minghinmatthewlam: updated to take in a paramter

scripts/install_avalanchego_release.sh
```diff
@@ -16,8 +16,7 @@ source "$TELEPORTER_PATH"/scripts/versions.sh
 source "$TELEPORTER_PATH"/scripts/constants.sh
 
 ############################
-# download avalanchego
-# https://github.com/ava-labs/avalanchego/releases
+# download avalanchego# https://github.com/ava-labs/avalanchego/releases
```
geoff-vball, line 19: nit: you've got a second `#` on this line

utils/deployment-utils/deployment_utils.go
```diff
@@ -189,9 +190,6 @@ func ConstructKeylessTransaction(
 	return contractCreationTxBytes, senderAddress, contractAddress, nil
 }
 
-// AlterNicksTransaction increments the gas price of the transaction used to deploy the contract.
-// This changes the derived contract and deployer address, which is useful for deploying the same
-// Teleporter contract to different addresses.
-func AlterNicksTransaction() {
-	contractCreationGasPrice.Add(contractCreationGasPrice, big.NewInt(1))
+func GetDefaultContractCreationGasPrice() *big.Int {
+	return defaultContractCreationGasPrice
```
geoff-vball, line 194: Can we copy the big.Int here? Returning the pointer to the global var means it can be altered

geoff-vball (commented):
Two nits, otherwise lgtm

geoff-vball (approved):

tests/utils/utils.go
```diff
@@ -323,6 +321,80 @@ func CreateReceiveCrossChainMessageTransaction(
...
+	callData, err := teleporterregistry.PackAddProtocolVersion(0)
+	Expect(err).Should(BeNil())
+
+	gasFeeCap, gasTipCap, nonce := CalculateTxParams(ctx, subnetInfo, PrivateKeyToAddress(senderKey))
+
+	destinationTx := predicateutils.NewPredicateTx(
+		subnetInfo.EVMChainID,
+		nonce,
+		&teleporterRegistryAddress,
+		500_000,
```
cam-schultz, line 347: How'd you arrive at 500_000? Curious because I'm implementing similar logic in the relayer to handle off chain messages.

minghinmatthewlam: Initially had the e2e test with too little gas, and in the error message it told me the expected gas

cam-schultz (approved):
LGTM

geoff-vball (commented):
🚀

geoff-vball (approved):

PR #247, opened 2024-01-22
refactored basic_send_receive.sh

hsk81:
The original code was not readable due to long variable names. Hence, renamed and slightly reformatted the code to increase readability.

### DIFF
--- a/scripts/local/examples/basic_send_receive.sh
+++ b/scripts/local/examples/basic_send_receive.sh
@@ -2,140 +2,180 @@
 # Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
 # See the file LICENSE for licensing terms.
 
-set -e # Stop on first error
+set -e # stop on first error
 
-# Variables provided by run_setup.sh:
-#   user_private_key
-#   user_address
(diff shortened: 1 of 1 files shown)


hsk81:
Further, the `send_bytes32` variable seems to be not used.

michaelkaplan13:
@hsk81 thanks for opening the PR. As a general standard, we have preferred descriptive variable names (even if longer) rather than abbreviations that people may not be familiar with. It makes it easier for people looking at the code for the first time to understand, and lightens the cognitive load of having to remember what each abbreviation means.

For that reason, I think we'll keep the variable names as is. If you'd like to update the PR though, I'm definitely on board with removing the unused var you mentioned.

hsk81:
No, after reading the `README.md`, as a dev you're directed towards this `bash_send_receive.sh` shell script, which to be very frank is an unreadable abomination. In such cases, my preferred way of understanding such code is to keep renaming, reformating and refactoring stuff until it starts making sense to me. Hence, it was more of a personal exercise, and I thought you guys may want to use it. But if not, due to code convention issues, just feel free to close this PR.

PR #253, opened 2024-01-24
Add version to openzeppelin imports and remove subnet-evm-contracts dependency for Teleporter upgrade contracts

martineckardt:
## Why this should be merged

The previous contracts worked only in the provided development repository. This may be a problem for teams with existing development setups and devs just playing around with Teleporter using Remix. This is due to two main issues:
- @openzeppelin/contracts resolves to the latest version with breaking changes
- @subet-evm-contracts is not published

## How this works
- Add version to foundry remapping to match publishable URL
```
import {IERC20} from "@openzeppelin/contracts@4.8.1/token/ERC20/IERC20.sol";
```
- Copy IWarpMessenger.sol to Teleporter directory to avoid dependency on file in subnet-evm-contracts git submodule

## How this was tested
- local/examples/basic_send_receive.sh
- compile sample contract with Remix:

```
pragma solidity 0.8.18;

import "https://github.com/ava-labs/teleporter/blob/fix-registry-ownable-explicit-owner/contracts/src/Teleporter/upgrades/TeleporterOwnerUpgradeable.sol";

contract ReceiverOnBulletin is TeleporterOwnerUpgradeable {

	// Constructor passes in the Teleporter registry address
    // to the TeleporterUpgradeable contract.
    constructor(address teleporterRegistryAddress)
        TeleporterOwnerUpgradeable(teleporterRegistryAddress)
    {}
  
    string public lastMessage;

    /**
     * @dev See {ITeleporterReceiver-receiveTeleporterMessage}.
     *
     * Receives a message from another chain.
     */
    function _receiveTeleporterMessage(
        bytes32 originChainID,
        address originSenderAddress,
        bytes memory message
    ) internal override {
        // Store the message.
        lastMessage = abi.decode(message, (string));
    }

}
```

## How is this documented


martineckardt:
Not sure why the test are failing with an error about the submodules setup, but will look into this, if the general idea looks good to you guys

geoff-vball:
Sorry, could you break down the change for OZ just a little bit more for me? I'm assuming remix resolves `@openzeppelin/contracts@4.8.1` to `https://github.com/OpenZeppelin/openzeppelin-contracts/tree/v4.8.1`? Does remix have no way of remapping like forge does?

I'd like to avoid copy-pasting `IWarpMessenger` from subnet-evm if possible. Will it work for you if we just use the absolute path to the file within the subnet-evm submodule when we import it?

minghinmatthewlam:
We have updated the import dependencies on `openzeppelin-contracts` and `subnet-evm` contracts to be compatible with both Foundry and Remix. Foundry will use the `remappings.txt` to compile, and when using with Remix in the academy courses it'll resolve to the existing npm packages. The versions in the imports help us set the package version we use in Remix, and make it easier to tell which version we have set to use in the repo.

martineckardt:
Works like a charm! Thanks a lot for your support @minghinmatthewlam 🙌🏻

cam-schultz (approved):
This is great! We'll have to remember to update the versions in `remappings.txt` whenever we update the OZ or subnet-evm submodules (or at least whenever the relevant contracts in those submodules change).

I'm wondering if we could add a CI check of some kind to make sure they stay up to date, but my hunch is that it would be tricky. Especially for subnet-evm since the npm and git versions are distinct.

minghinmatthewlam:
> This is great! We'll have to remember to update the versions in `remappings.txt` whenever we update the OZ or subnet-evm submodules (or at least whenever the relevant contracts in those submodules change).
> 
> I'm wondering if we could add a CI check of some kind to make sure they stay up to date, but my hunch is that it would be tricky. Especially for subnet-evm since the npm and git versions are distinct.

I think right now we won't be changing the OZ/Subnetevm contract versions often/much at all, so should be good to update manually for remappings when needed.

PR #256, opened 2024-01-25
Move off chain helpers to utils package

cam-schultz:
## Why this should be merged
Exports and moves the helpers introduced in https://github.com/ava-labs/teleporter/pull/240 to `package utils` so that they can be used in similar `awm-relayer` e2e tests.

### DIFF
--- a/tests/utils/utils.go
+++ b/tests/utils/utils.go
@@ -25,7 +25,9 @@ import (
 	gasUtils "github.com/ava-labs/teleporter/utils/gas-utils"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/constants"
 	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
+	"github.com/ava-labs/avalanchego/vms/platformvm/warp/payload"
 	"github.com/ava-labs/subnet-evm/accounts/abi"
 	"github.com/ava-labs/subnet-evm/accounts/abi/bind"
 	"github.com/ava-labs/subnet-evm/core/types"
@@ -36,6 +38,7 @@ import (
 	"github.com/ava-labs/subnet-evm/rpc"
 	"github.com/ava-labs/teleporter/tests/interfaces"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


geoff-vball (dismissed):

tests/utils/utils.go
```diff
@@ -944,3 +947,81 @@ func SendExampleCrossChainMessageAndVerify(
...
 	}
 }
+
+func InitChainConfig(
+	networkID uint32,
+	subnet interfaces.SubnetTestInfo,
+	teleporterAddress common.Address,
+) (*avalancheWarp.UnsignedMessage, string) {
+	unsignedMessage := CreateOffChainRegistryMessage(networkID, subnet, teleporterregistry.ProtocolRegistryEntry{
+		Version:         big.NewInt(2),
```
minghinmatthewlam, line 957: can this be a parameter for utils function

tests/utils/utils.go
```diff
@@ -944,3 +947,81 @@ func SendExampleCrossChainMessageAndVerify(
...
+	return unsignedMessage
+}
+
+func DeployNewTeleporterVersion(
+	ctx context.Context,
+	network interfaces.LocalNetwork,
+	fundedKey *ecdsa.PrivateKey,
+	teleporterByteCodeFile string,
+) common.Address {
+	contractCreationGasPrice := (&big.Int{}).Add(deploymentUtils.GetDefaultContractCreationGasPrice(), big.NewInt(1))
```
minghinmatthewlam, line 1003: note that when this gas price for the tx increases, the funds transferred to the deployer to `DeployTeleporterContracts` will also need to increase

cam-schultz: The deployer address is funded in `network.DeployTeleporterContracts`

tests/utils/utils.go
cam-schultz: For sure

tests/utils/utils.go
```diff
@@ -944,3 +947,82 @@ func SendExampleCrossChainMessageAndVerify(
 		Expect(currMessage).ShouldNot(Equal(message))
 	}
 }
+
+func InitChainConfig(
```
michaelkaplan13, line 951: Could you add a comment documenting what these functions do and/or what they are intended to be used for?

michaelkaplan13 (commented):
Just a request for comments, but otherwise LGTM

tests/utils/utils.go
minghinmatthewlam: As in right now the `DeployTeleporterContracts` hardcodes to 11avax, but if multiple `DeployNewTeleporterVersion`s are called, each time it'd require more funds, but it's still hardcoded at 11avax so would end up failing

geoff-vball (approved):

tests/utils/utils.go
cam-schultz: I don't think we'll ever realistically hit that case in our tests. The gas price in the keyless tx is set to 2500 nAVAX, and the gas limit to 4000000, which are intentionally high to prevent network congestion from causing the tx to fail.

minghinmatthewlam (dismissed):

tests/utils/utils.go
cam-schultz: Added

michaelkaplan13 (approved):

minghinmatthewlam (approved):

PR #265, opened 2024-01-30
Adding initalOwner to TeleporterOwnerUpgradeable

### LINKED ISSUE #262: Add initialOwner to TeleporterOwnerUpgradeable
**Context and scope**
We currently use openzeppelin contracts `v4.8.1` because the newer releases require solc version `^0.8.20`, but Teleporter had its audits on version `0.8.18` and don't think it's worth the changes to update audited code right before launch.

In comparing the differences between OZ's latest release compared to our set release version, there weren't too many notable changes for code we import from OZ. This issue does address a [change](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2639) to add `initialOwner` as an argument to `Ownable` so that at deployment we can set an owner different from deployer without the additional need to call `transferOwnership`.

minghinmatthewlam:
## Why this should be merged
Adds `initialOwner` as a constructor to `TeleporterOwnerUpgradeable`, fixes #262 

## How this works
`TeleporterOwnerUpgradeable` takes in an extra parameter `initialOwner` so that the owner is not by default `msg.sender`. In the constructor we will call `transferOwnership` to the `initialOwner`

## How this was tested
Unit tests
CI

## How is this documented
updated README.md


tests/flows/example_messenger.go
```diff
@@ -10,16 +10,22 @@ import (
...
-		ctx, fundedKey, subnetBInfo,
+	_, exampleMessengerA := utils.DeployExampleCrossChainMessenger(ctx,
+		fundedKey,
+		fundedAddress,
+		subnetAInfo,
+	)
+	exampleMessengerAddressB, exampleMessengerB := utils.DeployExampleCrossChainMessenger(ctx,
+		fundedKey,
+		fundedAddress,
+		subnetBInfo,
```
geoff-vball, line 28: nit: can you put `ctx` on a new line?

tests/flows/insufficient_gas.go
```diff
@@ -17,11 +17,17 @@ func InsufficientGas(network interfaces.Network) {
...
+		subnetAInfo,
+	)
 	// Deploy ExampleMessenger to Subnets B
-	exampleMessengerContractB, subnetBExampleMessenger :=
-		utils.DeployExampleCrossChainMessenger(ctx, fundedKey, subnetBInfo)
+	exampleMessengerContractB, subnetBExampleMessenger := utils.DeployExampleCrossChainMessenger(ctx,
+		fundedKey,
+		fundedAddress,
+		subnetBInfo,
+	)
```
geoff-vball, line 30: Same here for ctx

tests/flows/pause_teleporter.go
```diff
@@ -12,16 +12,22 @@ import (
...
-		ctx, fundedKey, subnetBInfo,
+	_, exampleMessengerA := utils.DeployExampleCrossChainMessenger(ctx,
+		fundedKey,
+		fundedAddress,
+		subnetAInfo,
+	)
+	exampleMessengerAddressB, exampleMessengerB := utils.DeployExampleCrossChainMessenger(ctx,
+		fundedKey,
+		fundedAddress,
+		subnetBInfo,
```
geoff-vball, line 30: and here

geoff-vball (commented):
Just a few nits, otherwise lgtm

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
```diff
@@ -233,6 +238,7 @@ contract NativeTokenDestinationTest is NativeTokenBridgeTest {
 
         new NativeTokenDestination(
             MOCK_TELEPORTER_REGISTRY_ADDRESS,
+            address(this),
```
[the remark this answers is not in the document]
michaelkaplan13: I actually have the reverse question, why `msg.sender` in the cases above instead of `address(this`) 😅 

My understanding is that `address(this)` is the unit test contract itself, meaning later on in the test we could directly make calls as the owner. I'm not sure what the `msg.sender` value is within the unit test functions (presumably some default address that coordinates the test runs, or is mocked out). I'd recommend not using `msg.sender` here either way since it's not clear what the address will be, and it may change if forge changes the behavior. Instead, I'd recommend using a default address that we define as a constant, and can mock if needed.

contracts/src/CrossChainApplications/examples/ExampleMessenger/ExampleCrossChainMessenger.sol
```diff
@@ -54,9 +54,10 @@ contract ExampleCrossChainMessenger is ReentrancyGuard, TeleporterOwnerUpgradeab
         bytes32 indexed sourceBlockchainID, address indexed originSenderAddress, string message
     );
 
-    constructor(address teleporterRegistryAddress)
-        TeleporterOwnerUpgradeable(teleporterRegistryAddress)
-    {}
+    constructor(
+        address teleporterRegistryAddress,
+        address initialOwner
```
michaelkaplan13, line 59: To reduce the diff and given that these are all only example applications, I have a slight preference for using `msg.sender` rather than adding an extra constructor parameter to all of the examples. I think it raises more questions than benefits in the examples.

"Owner" is also a scary term in smart contracts since it applies they have significant admin-type access. Is there a better term we could use to make clear the owner only has the ability to set the minimum Teleporter version allowed to be used?

geoff-vball: Maybe something like `teleporterUpgradeableAdmin`?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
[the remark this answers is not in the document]
minghinmatthewlam: I meant to use `address(this)`, instead now I added a `DEFAULT_OWNER_ADDRESS` constant. A bit of a pain in the owner upgradeable tests since the tests need to `vm.prank` the owner address to test calls, but outside of that contract don't think it'll be used in too many unit tests

contracts/src/CrossChainApplications/examples/ExampleMessenger/ExampleCrossChainMessenger.sol
minghinmatthewlam: My thinking is that the example cross chain apps should illustrate a standard for how apps should integrate with Teleporter, and that it'd be a good standard for apps to have an `initialOwner` argument. 

No strong pref for naming, good with `teleporterUpgradeableAdmin`, `teleporterManager`, `teleporterUpdater`, etc.

michaelkaplan13: `teleporterManager` SGTM 👍

geoff-vball (approved):

contracts/src/CrossChainApplications/GETTING_STARTED.md
```diff
@@ -218,7 +225,7 @@ mapping(bytes32 sourceBlockchainID => Message message) private _messages;
 
 Next, update `receiveTeleporterMessage` to save the message into the mapping after it is received and verified that it's sent from Teleporter. ABI decode the `message` bytes into a string. Also, emit the `ReceiveMessage` event.
 
-```solidity
+````solidity
```
cam-schultz, line 228: I don't think we need this fourth backtick

contracts/src/CrossChainApplications/GETTING_STARTED.md
```diff
@@ -242,7 +249,7 @@ function receiveTeleporterMessage(
         messageString
     );
 }
-```
+````
```
cam-schultz, line 252: Same here

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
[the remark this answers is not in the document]
cam-schultz: Is there a reason this constructor call still uses `address(this)` while the other test constructor calls use `_DEFAULT_OWNER_ADDRESS`?

cam-schultz (commented):
Just a couple of nits, otherwise LGTM

geoff-vball (approved):

Issue #263, opened 2024-01-30
[Bug]: Docker set up failing


minghinmatthewlam:
**Describe the bug**
Docker setup fails to start properly

**To Reproduce**
Run `./scripts/local/run.sh` and wait for network to try to start up

**Expected behavior**
The docker setup should keep running, where you can then enter the container to execute tests.

**Logs**
```
local_network_run  | [node4] [01-30|17:52:59.090] ERROR chains/manager.go:359 error creating chain {"subnetID": "2CZP2ndbQnZxTzGuZjPrJAm5b4s2K2Bcjh8NqWoymi8NZMLYQk", "chainID": "RjxBN1Pfxg6QWbTdL7fYj9FF2cLSCQKWuSzkXjuCKtURbYJop", "chainAlias": "RjxBN1Pfxg6QWbTdL7fYj9FF2cLSCQKWuSzkXjuCKtURbYJop", "vmID": "srEXiWaHtoWw6GFGeF3WAxWwAueYWwoD5o2HUjQY9ASJRW32P", "error": "error while creating new snowman vm rpc error: code = Unknown desc = failed to verify genesis: invalid precompile upgrades: warp cannot be activated before Durango"}
local_network_run  | [01-30|17:53:25.501] INFO ux/output.go:40 [node4] [01-30|17:52:59.090] ERROR chains/manager.go:359 error creating chain {"subnetID": "2CZP2ndbQnZxTzGuZjPrJAm5b4s2K2Bcjh8NqWoymi8NZMLYQk", "chainID": "RjxBN1Pfxg6QWbTdL7fYj9FF2cLSCQKWuSzkXjuCKtURbYJop", "chainAlias": "RjxBN1Pfxg6QWbTdL7fYj9FF2cLSCQKWuSzkXjuCKtURbYJop", "vmID": "srEXiWaHtoWw6GFGeF3WAxWwAueYWwoD5o2HUjQY9ASJRW32P", "error": "error while creating new snowman vm rpc error: code = Unknown desc = failed to verify genesis: invalid precompile upgrades: warp cannot be activated before Durango"}
local_network_run  | ----------------------------------------------------------------------
local_network_run  | [01-30|17:53:25.501] INFO ux/output.go:40 ----------------------------------------------------------------------
local_network_run  |
[...]

PR #268, opened 2024-02-01
Remove Durango config

cam-schultz:
## Why this should be merged
Fixes https://github.com/ava-labs/teleporter/issues/263.

Note: this is caused by a bug in avalanche-network-runner that has been [merged](https://github.com/ava-labs/avalanche-network-runner/pull/697), but not yet included in a release. We may need to revert this change once it is patched there.

~~We will wait to merge until https://github.com/ava-labs/awm-relayer/pull/161 is merged, and a new awm-relayer release published.~~ This PR is ready to merge.

## How this works
- Fixes the genesis config to be compatible with ANR
- Bumps awm-relayer version to latest

## How this was tested
CI, manually verified that the example script runs using the latest awm-relayer image

## How is this documented

### DIFF
--- a/docker/relayerConfigTemplate.json
+++ b/docker/relayerConfigTemplate.json
@@ -2,6 +2,7 @@
     "network-id": 1337,
     "p-chain-api-url": "http://127.0.0.1:9650",
     "encrypt-connection": false,
+    "process-missed-blocks": false,
     "source-subnets": [
         {
             "subnet-id": "11111111111111111111111111111111LpoYY",
--- a/docker/docker-compose-run.yml
+++ b/docker/docker-compose-run.yml
@@ -21,7 +21,7 @@ services:
         source: ./
         target: /code/
   relayer:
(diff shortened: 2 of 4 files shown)


michaelkaplan13 (approved):

docker/relayerConfigTemplate.json
```diff
@@ -2,6 +2,7 @@
     "network-id": 1337,
     "p-chain-api-url": "http://127.0.0.1:9650",
     "encrypt-connection": false,
+    "process-missed-blocks": false,
```
michaelkaplan13, line 5: Do we need to add the "startBlock" to the source subnet configurations here?

cam-schultz: If `process-missed-blocks` was true, then yes we would have to do that. But the starting block is only calculated is `process-missed-blocks` is enabled

minghinmatthewlam (approved):

geoff-vball (approved):

PR #269, opened 2024-02-01
Cleanup registry unit tests

minghinmatthewlam:
## Why this should be merged
Cleans up `_createWarpOffChainMessage` in Teleporter registry unit tests

## How this works
Removes extra parameter `destinationAddress`, because it already takes in registry address, and should be same value on success.

## How this was tested
Unit tests, CI

## How is this documented
n/a


geoff-vball (approved):

contracts/src/Teleporter/upgrades/tests/TeleporterRegistryTests.t.sol
```diff
@@ -252,7 +230,11 @@ contract TeleporterRegistryTest is Test {
 
         // Check if we have an invalid destination address
         warpMessage = _createWarpOffChainMessage(
-            latestVersion + 1, address(this), teleporterAddress, address(teleporterRegistry)
```
cam-schultz, line 255: Can you help me understand why `destinationAddress` was `address(this)` previously, and why this is no longer needed?

cam-schultz (commented):
One clarification question to make sure the tests are covering the same cases as before.

contracts/src/Teleporter/upgrades/tests/TeleporterRegistryTests.t.sol
minghinmatthewlam: it was set as `address(this)` to mock an invalid Warp message to the registry, we still set it in the next line

cam-schultz: Thanks!

cam-schultz (approved):

michaelkaplan13 (approved):

PR #273, opened 2024-02-05
Ginkgo filtering for e2e tests

### LINKED ISSUE #264: Filter e2e tests through a command line argument
**Context and scope**
Right now all our tests run in [e2e_test.go](https://github.com/ava-labs/teleporter/blob/main/tests/local/e2e_test.go), they can be filtered by replacing with `FIt`, but this requires goign into the code changing and remembering to change back.

Allow an argument that specifies test names to filter for running `e2e_test.sh`.

minghinmatthewlam:
## Why this should be merged
When writing a new e2e test or debugging a failing e2e tests, instead of needing to either comment out other test cases, or use `Fit` that will need to be added and removed manually, an env variable in command line makes it easier to test. Fixes #264 

## How this works
Allows `GINKGO_LABEL_FILTER` and `GINKGO_FOCUS` for the script `e2e_test.sh`. The ginkgo e2e tests are grouped together with labels, but can also be focused based on test description name. 

## How this was tested
locally

## How is this documented
README


geoff-vball (approved):

michaelkaplan13 (approved):
👍

Issue #187, opened 2023-12-08
Add multiplier to `NativeTokenDestination`


geoff-vball:
**Context and scope**
Some ERC20 contracts have `decimals` of something other than `18`. Currently, sending 1 full token from an ERC20 contract with `24` decimals, will result in `10^6` full tokens being minted on the destination. We should make this configurable.

**Discussion and alternatives**
We decided that this should be handled on the destination chain only, as the source chain does not need to know the implementation specifics of the destination, and with this implementation, we can unify the interfaces among all the different token bridge contracts.

**Open questions**

PR #277, opened 2024-02-07
Add multiplier to ERC20TokenSource

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/187

## How this works
Adds a multiplier to the constructor of `ERC20TokenSource` as well as a boolean to determine which direction the multiplier should be applied in. 

All the implementation is in the source contract. The destination does not need to know about the multiplier at all.

I had originally considered normalizing all ERC20s to 18 decimals, but it was simpler to give contract deployers the choice of what multiplier they wanted to use.

We could also extend this to the `NativeTokenSource` if desired. One subnet might want to use a multiple of another's subnet's native token as their gas token, or they might want to use a diluted version of AVAX.

## How this was tested
Unit tests

## How is this documented
Additions to the README


geoff-vball:
Closing this, will add the multiplier to the destination

PR #282, opened 2024-02-09
Minor fixes to local network scripts

geoff-vball:
## Why this should be merged
Fixes some minor issues in the local network scripts, including having one script hanging after exit.

## How this was tested
Locally

## How is this documented


minghinmatthewlam (approved):

scripts/local/run.sh
```diff
@@ -7,40 +7,38 @@ set -e # Stop on first error
...
+    esac
+    shift
+done
+
 # Set up the trap to catch the SIGINT signal (CTRL+C)
 # Note that the output of the cleanup function defined in run_setup.sh does not appear if CTRL+C is used to kill this script,
 # but the function does in fact run as expected.
-trap cleanup SIGTERM
-trap cleanup SIGINT
+trap cleanup SIGTERM SIGINT
```
minghinmatthewlam, line 41: was the script hanging after CTRL+C? I thought that's what the trap was handling, and don't think ive had that issue previously

geoff-vball: No, it was hanging on line 64 if docker errored out. CTRL+C got it to exit, but you shouldn't have to in that situation.

michaelkaplan13 (approved):

PR #285, opened 2024-02-14
Remove most uses of Websocket Client

geoff-vball:
## Why this should be merged
Removes all uses of the Websocket Client except for in `validator_churn.go`. We had originally used it to wait for transactions to be mined, but looping on the RPC client waiting for a transaction hash to be mined easier, more robust on testnets, and only requires one type of endpoint.

## How this works
`IssueTxsToActivateProposerVMFork` in `subnet-evm` requires a websocket client, so we cannot totally eliminate our use of it.

## How this was tested
Existing E2E tests

## How is this documented

### DIFF
--- a/tests/flows/erc20_to_native_token_bridge.go
+++ b/tests/flows/erc20_to_native_token_bridge.go
@@ -132,17 +132,17 @@ func ERC20ToNativeTokenBridge(network interfaces.LocalNetwork) {
 	// Create abi objects to call the contract with
 	nativeTokenDestination, err := nativetokendestination.NewNativeTokenDestination(
 		bridgeContractAddress,
-		destSubnet.WSClient,
+		destSubnet.RPCClient,
 	)
 	Expect(err).Should(BeNil())
 	erc20TokenSource, err := erc20tokensource.NewERC20TokenSource(
 		bridgeContractAddress,
-		sourceSubnet.WSClient,
+		sourceSubnet.RPCClient,
 	)
 	Expect(err).Should(BeNil())
 	exampleERC20, err := exampleerc20.NewExampleERC20(
 		exampleERC20ContractAddress,
-		sourceSubnet.WSClient,
+		sourceSubnet.RPCClient,
 	)
 	Expect(err).Should(BeNil())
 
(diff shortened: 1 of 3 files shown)


cam-schultz:
I believe the local network WS client can also be used for RPC calls. Since for the testnet flows we exclusively use the RPC client, it seems like it should be possible to simplify `SubnetTestInfo` to only consist of a single `ethclient.Client`, with testnet flows using an RPC client and local network flows using a WS client. Curious what you think about that.

We had exactly this setup in an earlier iteration, but that was before we considered running these flows on Fuji. Hindsight is 20/20 🤷

geoff-vball:
@cam-schultz So after a bunch of messing around with this, I'm going to say it's a bad idea to switch to using the WS client for local tests. The problem is the client object that the pass into the ABI bindings to get the abi objects that we can make calls on. If a websocket client is passed and is closed, we would have to regenerate the abi object by passing in a new client. For whatever reason, the abi objects with rpc clients that have been closed, actually still work. So using the websocket client actually breaks the `upgradeability` and `validator churn` tests, and it's very very messy to fix.

cam-schultz:
> @cam-schultz So after a bunch of messing around with this, I'm going to say it's a bad idea to switch to using the WS client for local tests. The problem is the client object that the pass into the ABI bindings to get the abi objects that we can make calls on. If a websocket client is passed and is closed, we would have to regenerate the abi object by passing in a new client. For whatever reason, the abi objects with rpc clients that have been closed, actually still work. So using the websocket client actually breaks the `upgradeability` and `validator churn` tests, and it's very very messy to fix.

Ah yeah, that sounds like more trouble than it's worth. Thanks for digging into it.

cam-schultz (approved):
LGTM.

One other idea to simplifying `SubnetTestInfo` to just a single (RPC) client would be to modify `IssueTxsToActivateProposerVMFork` to use `bind.WaitMined` instead of waiting for a new block with eth_subscribe. Obviously that would require a change to subnet-evm as well, so I'm cool with merging this as is in the meantime.

michaelkaplan13 (approved):

minghinmatthewlam:
are we able to get rid of the `SubnetTestInfo.WSClient` completely? Right now we have a weird case for the testnet setup wehre the WSClient is always `nil`.

geoff-vball:
@minghinmatthewlam Can't remember where I left the original comment, but the only thing keeping us from removing the WS client is the call to `IssueTxsToActivateProposerVMFork()` in `validator_churn.go`. It's implemented in `subnet-evm` and uses `SubscribeNewHead()` and therefore requires a WSclient. We could create our own implementation (all it does is send two transactions) which would let us remove the WSclient, but we mayyy want to keep it around in case we need to subscribe to the node for other tests in the future.

PR #287, opened 2024-02-14
Add sendTeleporterMessage helper to TeleporterUpgradeable

minghinmatthewlam:
## Why this should be merged
Adds a basic send Teleporter message helper to `TeleporterUpgradeable` that uses `_getTeleporterMessenger()`, which by default uses the latest Teleporter, and also handles fee transfer to Teleporter.

## How this works
Teleporter apps that inherit `TeleporterUpgradeable` can use `_sendTeleporterMessage` for their most basic send handling. Function can also be overridden for specific implementations.

Also updates `subnet-evm` commit for submodule.

## How this was tested
CI

## How is this documented
README


contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -217,6 +221,33 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
      */
     function _checkTeleporterUpgradeAccess() internal virtual;
 
+    /**
+     * @dev Sends a cross chain message using the TeleporterMessenger contract.
+     * @return `messageID` The unique identifier for the Teleporter message.
+     * @return `adjustedFeeAmount` The adjusted fee amount after transferring the fee token.
```
michaelkaplan13, line 227: Is there any need to return the `adjustedFeeAmount` here? It seems a little odd that the interface doesn't match `sendCrossChainMessage`.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -217,6 +221,33 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
...
+     */
+    function _sendTeleporterMessage(TeleporterMessageInput memory messageInput)
```
michaelkaplan13, line 229: Could you add unit tests for this new function?

contracts/lib/subnet-evm
michaelkaplan13, line 1: It's tough to tell from Github, but is this updating the submodule to https://github.com/ava-labs/subnet-evm/releases/tag/v0.6.0-fuji? I think we should stick to release version commits at this point if not.

michaelkaplan13 (commented):
Few small comments, but this will be a great addition 👍

geoff-vball:
Basically looks good! I agree with Michael's nits

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: hmm I originally just noticed in our example apps we like to emit in an event the `adjustedFeeAmount`, but that makes sense, I'll just return `sendCrossChainMessage` call

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: added

cam-schultz (commented):
Generally LGTM. It would be helpful to modify some or all of the example apps to use the new helper to demonstrate its usage in a contract.

michaelkaplan13 (dismissed):

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -217,6 +221,31 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
...
+        // For non-zero fee amounts increase the Teleporter contract's allowance.
+        if (messageInput.feeInfo.amount > 0) {
+            require(
+                messageInput.feeInfo.feeTokenAddress != address(0),
+                "TeleporterUpgradeable: zero fee token address"
+            );
+            IERC20(messageInput.feeInfo.feeTokenAddress).safeIncreaseAllowance(
+                address(teleporterMessenger), messageInput.feeInfo.amount
+            );
+        }
```
geoff-vball, line 245: I think having the call to `safeTransferFrom` outside of this function and `safeIncreaseAllowance` is a little strange, but I understand why we've done things this way and think it's the right choice. Can we leave comments here, and anywhere else that makes sense that note that fees should transferred to the contract before calling `_sendTeleporterMessage`?

geoff-vball (dismissed):

geoff-vball (dismissed):

michaelkaplan13 (approved):

geoff-vball (approved):

PR #290, opened 2024-02-16
Waiting for tx requires hash instead of full tx

geoff-vball:
## Why this should be merged
Fixes test flakiness where the RPC endpoint could not find the transaction with the hash returned from deploying the teleporter contract.

## How this works
Copies the implementation of `bind.WaitMined()` from `subnet-evm`, but takes the transaction hash as a parameter instead of the full transaction, because only the hash is needed. This saves us from looking up the full transaction from the hash when we deploy the teleporter contract, which was flaky with no retries. 

This will require code changes in the relayer repo

## How this was tested
Existing tests

## How is this documented


tests/local/network.go
```diff
@@ -302,12 +302,10 @@ func (n *LocalNetwork) DeployTeleporterContracts(
...
 
-			txHash := &common.Hash{}
-			err = rpcClient.CallContext(ctx, txHash, "eth_sendRawTransaction", hexutil.Encode(transactionBytes))
+			txHash := common.Hash{}
+			err = rpcClient.CallContext(ctx, &txHash, "eth_sendRawTransaction", hexutil.Encode(transactionBytes))
 			Expect(err).Should(BeNil())
-			tx, _, err := subnetInfo.RPCClient.TransactionByHash(ctx, *txHash)
-			Expect(err).Should(BeNil())
-			utils.WaitForTransactionSuccess(ctx, subnetInfo, tx)
+			utils.WaitForTransactionSuccess(ctx, subnetInfo, txHash)
```
geoff-vball, line 308: Not sure how to link this without leaving a comment

michaelkaplan13 (dismissed):

tests/utils/utils.go
```diff
@@ -536,6 +538,37 @@ func waitForTransaction(
 	return receipt
 }
 
+// WaitMined waits for tx to be mined on the blockchain.
+// It stops waiting when the context is canceled.
+// Takes a tx hash instead of the full tx in the subnet-evm version of this function.
```
minghinmatthewlam, line 543: nit: can you link to subnet-evm's implementation in comment

tests/utils/utils.go
```diff
@@ -536,6 +538,38 @@ func waitForTransaction(
...
+// Copied an modified from https://github.com/ava-labs/subnet-evm/blob/v0.6.0-fuji/accounts/abi/bind/util.go#L42
```
minghinmatthewlam, line 544:
```suggestion
// Copied and modified from https://github.com/ava-labs/subnet-evm/blob/v0.6.0-fuji/accounts/abi/bind/util.go#L42
```

super nit typo

michaelkaplan13 (approved):

minghinmatthewlam (approved):

PR #291, opened 2024-02-16
Add multiplier to native token destination

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/187

## How this works
Adds a multiplier to the constructor of `NativeTokenDestination` as well as a boolean to determine which direction the multiplier should be applied in. 

All the implementation is in the destination contract. The source contract does not need to know about the multiplier at all.

## How this was tested
Unit tests

## How is this documented
Additions to the README


contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -60,12 +60,24 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
 
+    // tokenMultiplier allows this contract to scale the number of tokens it sends/receives to/from
+    // the source chain. This can be used to normalize the number of decimals places between
+    // the two subnets.
+    uint256 public immutable tokenMultiplier;
+    // If multiplyOnSend is true, the number of tokens sent to the destination chain will be multiplied
+    // by `tokenMultiplier`, and tokens received from the source chain will be divided by `tokenMultiplier`.
+    // If multiplyOnSend is false, the number of tokens sent to the destination chain will be divided
+    // by `tokenMultiplier`, and tokens received from the source chain will be divided by `tokenMultiplier`.
+    bool public immutable multiplyOnSend;
```
michaelkaplan13, line 71:
```suggestion
    // If multiplyOnReceive is true, the raw token amount value will be multiplied by `tokenMultiplier` when tokens
    // are transferred from the source chain into this destination chain, and divided by `tokenMultiplier` when
    // tokens are when tokens are transferred from this destination chain back to the source chain. This is intended 
    // when the "decimals" value on the source chain is less than the native EVM denomination of 18.
    // If multiplyOnReceive is false, the raw token amount value will be divided by `tokenMultiplier` when tokens
    // are transferred from the source chain into this destination chain, and multiplied by `tokenMultiplier` when
    // tokens are when tokens are transferred from this destination chain back to the source chain. This is intended 
    // when the "decimals" value on the source chain is greater than the native EVM denomination of 18.
    bool public immutable multiplyOnReceive;
```

I think the original comment mixed up "sent to destination" and "received from source", which are actually the same transfer direction.

Multiply on send vs receive is definitely a nit...that's just how I initially thought about it for the case where the source decimals is less than 18, which is likely more far common that when its greater than 18.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/README.md
```diff
@@ -22,6 +22,7 @@ A contract that lives on the destination chain. Pairs with exactly one `TokenSou
...
 - `reportTotalBurnedTxFees`: Sends a Teleporter message to the source chain containing the total number of tokens at the burn address where transaction fees are sent (`0x0100000000000000000000000000000000000000`).
+- `NativeTokenDestination` requires the of setting a `tokenMultiplier` in the constructor. This allows for the scaling of tokens sent to the source chain, which can be used to normalize the number of decimals places for the token between the two subnets. If a 1-to-1 relationship is desired, please set `tokenMultiplier` to 1. If `multiplyOnSend` is set to true, the number of tokens sent to the source chain will be multiplied by `tokenMultiplier`, and tokens received from the source chain will be divided by `tokenMultiplier`. If `multiplyOnSend` is set to false, the number of tokens sent to the source chain will be divided by `tokenMultiplier`, and tokens received from the source chain will be multiplied by `tokenMultiplier`.
```
michaelkaplan13, line 25:
```suggestion
- `NativeTokenDestination` requires the of setting a `tokenMultiplier` in the constructor. This allows for the scaling of tokens sent to the source chain, which can be used to normalize the number of decimals places for the token between the two subnets. If a 1-to-1 relationship is desired, please set `tokenMultiplier` to 1. If `multiplyOnReceive` is set to true, the number of tokens transferred from the source chain into the destination chain will be multiplied by `tokenMultiplier`, and the number of tokens transferred from the destination chain back to the source chain will be divided by `tokenMultiplier`. If `multiplyOnReceive` is set to false, the number of tokens transferred from the source chain into the destination chain will be divided by `tokenMultiplier`, and number of tokens transferred from the destination chain back to the source chain will be multiplied by `tokenMultiplier`.
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -89,6 +101,10 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
 
         initialReserveImbalance = initialReserveImbalance_;
         currentReserveImbalance = initialReserveImbalance_;
+
+        require(tokenMultiplier_ != 0, "NativeTokenDestination: zero tokenMultiplier");
+        tokenMultiplier = tokenMultiplier_;
```
michaelkaplan13, line 106: I wonder if it would be clearer for developers if the multiplier was provided as a power of 10 (or even just requiring it to be a power of 10). If we're documenting it's purpose as a means of normalizing the "decimals" count, it wouldn't make sense for it to have any value others than powers of 10 IMO.

michaelkaplan13 (commented):
The logic all makes sense to me, my two comments are:
1. Should we have the multiplier be provided as a power of 10? i.e. For the USDC/T case where the source decimals is 6, the multiplier should be 12 rather than 1000000000000.
2. I got confused by the documentation of `multiplyOnSend`. I suggested we switch it to `multiplyOnReceive` because it feels more intuitive when thinking about it from the destination side, but either way think the comment wording can be clarified.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -60,12 +60,24 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
     uint256 public currentReserveImbalance;
     uint256 public totalMinted;
 
+    // tokenMultiplier allows this contract to scale the number of tokens it sends/receives to/from
+    // the source chain. This can be used to normalize the number of decimals places between
+    // the two subnets.
```
minghinmatthewlam, line 65:
```suggestion
    // tokenMultiplier allows this contract to scale the number of tokens it sends/receives to/from
    // the source chain. This can be used to normalize the number of decimals places between the tokens on
    // the two subnets.
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: agreed that it should be in power of 10

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -125,21 +141,30 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
         // Burn native token by sending to BURN_FOR_TRANSFER_ADDRESS
         Address.sendValue(payable(BURN_FOR_TRANSFER_ADDRESS), value);
 
+        uint256 scaledAmount;
+        if (multiplyOnSend) {
+            scaledAmount = msg.value * tokenMultiplier;
```
minghinmatthewlam, line 146: can use `value` instead of `msg.value`. Can we move this before the `sendValue` call? I like to keep internal calls before external calls when possible

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -164,6 +189,14 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
         }
 
         uint256 totalBurnedTxFees = BURNED_TX_FEES_ADDRESS.balance;
+
+        uint256 scaledAmount;
+        if (multiplyOnSend) {
```
minghinmatthewlam, line 194: can this be put in a helper function?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
```diff
@@ -87,7 +92,72 @@ contract NativeTokenDestinationTest is NativeTokenBridgeTest {
...
+
+        // We shouldn't mint anything here.
+        vm.expectCall(
+            NATIVE_MINTER_PRECOMPILE_ADDRESS,
+            abi.encodeWithSelector(INativeMinter.mintNativeCoin.selector),
+            0
+        );
+
+        vm.prank(MOCK_TELEPORTER_MESSENGER_ADDRESS);
+        nativeTokenDestination.receiveTeleporterMessage(
```
minghinmatthewlam, line 133: can you add a unit test that just checks the emit `NativeTokensMinted` for a `receiveTeleoprterMessage` where some amount is used as collateral, and the rest is minted.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -60,12 +60,28 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
     uint256 public currentReserveImbalance;
     uint256 public totalMinted;
 
+    // tokenMultiplier allows this contract to scale the number of tokens it sends/receives to/from
+    // the source chain. This can be used to normalize the number of decimals places between the tokens on
+    // the two subnets.
+    uint256 public immutable tokenMultiplier;
+    // If multiplyOnReceive is true, the raw token amount value will be multiplied by `tokenMultiplier` when tokens
+    // are transferred from the source chain into this destination chain, and divided by `tokenMultiplier` when
+    // tokens are when tokens are transferred from this destination chain back to the source chain. This is intended
```
bernard-avalabs, line 69: typo: extra "when tokens are"

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -60,12 +60,28 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
+    // when the "decimals" value on the source chain is less than the native EVM denomination of 18.
+    // If multiplyOnReceive is false, the raw token amount value will be divided by `tokenMultiplier` when tokens
+    // are transferred from the source chain into this destination chain, and multiplied by `tokenMultiplier` when
+    // tokens are when tokens are transferred from this destination chain back to the source chain. This is intended
```
bernard-avalabs, line 73: typo: extra "when tokens are"

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -125,21 +145,30 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
         // Burn native token by sending to BURN_FOR_TRANSFER_ADDRESS
         Address.sendValue(payable(BURN_FOR_TRANSFER_ADDRESS), value);
 
+        uint256 scaledAmount;
+        if (multiplyOnReceive) {
+            scaledAmount = msg.value / tokenMultiplier;
```
bernard-avalabs, line 150: I assume that integer division should never be an issue (i.e., rounding down) for any reasonable amount and multipler values.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -164,6 +193,14 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
         }
 
         uint256 totalBurnedTxFees = BURNED_TX_FEES_ADDRESS.balance;
+
+        uint256 scaledAmount;
+        if (multiplyOnReceive) {
+            scaledAmount = totalBurnedTxFees / tokenMultiplier;
```
bernard-avalabs, line 199: Perhaps consider creating a helper function for calcuating scaledAmount since we do this more than once.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/README.md
```diff
@@ -22,6 +22,7 @@ A contract that lives on the destination chain. Pairs with exactly one `TokenSou
...
 - `reportTotalBurnedTxFees`: Sends a Teleporter message to the source chain containing the total number of tokens at the burn address where transaction fees are sent (`0x0100000000000000000000000000000000000000`).
+- `NativeTokenDestination` requires the of setting a `tokenMultiplier` in the constructor. This allows for the scaling of tokens sent to the source chain, which can be used to normalize the number of decimals places for the token between the two subnets. If a 1-to-1 relationship is desired, please set `tokenMultiplier` to 1. If `multiplyOnReceive` is set to true, the number of tokens transferred from the source chain into the destination chain will be multiplied by `tokenMultiplier`, and the number of tokens transferred from the destination chain back to the source chain will be divided by `tokenMultiplier`. If `multiplyOnReceive` is set to false, the number of tokens transferred from the source chain into the destination chain will be divided by `tokenMultiplier`, and number of tokens transferred from the destination chain back to the source chain will be multiplied by `tokenMultiplier`.
```
bernard-avalabs, line 25: typo "requires the setting of a"

bernard-avalabs (commented):
Generally makes sense. I've left a few comments.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Yeah, in the worst case you could lose up to the smallest divisible unit on either of the chains, which is fine.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: So I thought about this a bit. I think restricting it to a multiple of 10 seems needlessly restrictive, and would be over-anticipating use cases. I think having the integer multiplier allows the users the most amount of freedom without being confusing. I've updated the docs slightly to allude to other use cases.

contracts/lib/subnet-evm
michaelkaplan13, line 1: Did you intend to change the subnet-evm commit used here?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -113,8 +133,11 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
             );
         }
 
+        uint256 scaledAmount = _scaleTokens(value, false);
```
michaelkaplan13, line 136: If `multiplyOnReceive` is true, should we require that the value here is divisible by `tokenMultiplier`? In practice it should be only very minor amounts that get lost due to integer division, but in theory those could be worth a significant amount. 

We're not able to handle the non-divisible amount case when receiving either way, but it may make sense to handle it here when sending.

contracts/lib/subnet-evm
geoff-vball: Nope, reverted

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: I'm not sure what you mean by that

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: I think that would be over-restrictive. The max we can lose in a transfer is less than one of the smallest divisible unit on one of the chains. If there's ever a situation where that is worth a non-negligible amount, checks could be added user-side.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
michaelkaplan13: It isn't specific to non-power-of-10 multipliers, but I'm imagining a case where a multiplier is used to make a token more divisible on the destination because a single unit of it on the source is worth a significant enough amount that the destination subnet wants to allow for smaller subdivisions of it. In that use case, I think our use of integer division is undesirable because if the multiplier is 4 and I move back 7 to the source chain, I only end up with 4.

In the use case of normalizing decimal counts, rounding down to the nearest unit of the source chain seems likely to be negligible (i.e. less than 0.000001 in the case of USDT/USDC), but in the use case of allowing further subdivision of an assets that's not divisible enough on the source, it seems less likely to be negligible.

That being said, I think requiring the amounts sent back to the source chain are divisible by the multiplier could largely avoid this. https://github.com/ava-labs/teleporter/pull/291#discussion_r1504928310

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
michaelkaplan13: Restricting transfer amounts to only valid values is a good restriction to have IMO. Otherwise we're making an additional assumption (the difference is worth a significant amount), which is likely to be true in the vast majority of use cases but feasibly not all. 

As is, it's also possible for `scaledAmount` to be 0, which I think would be an invalid case.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -55,12 +55,28 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
     uint256 public currentReserveImbalance;
     uint256 public totalMinted;
 
+    // tokenMultiplier allows this contract to scale the number of tokens it sends/receives to/from
+    // the source chain. This can be used to normalize the number of decimals places between the tokens on
+    // the two subnets.
+    uint256 public immutable tokenMultiplier;
+    // If multiplyOnReceive is true, the raw token amount value will be multiplied by `tokenMultiplier` when tokens
+    // are transferred from the source chain into this destination chain, and divided by `tokenMultiplier` when
```
minghinmatthewlam, line 63: for these multiline comments I'd prefer `/** */`

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: I also think we should only transfer amounts that are divisible, but think we can handle that logic in the contract to not restrict the passed in transfer amount argument. If the user passes an amount that is not divisible, we `removeDust`, meaning we return any amount that is non divisible back to the sender, and bridge the rest. LayerZero should have a similar concept in `OFT`s.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -236,4 +264,23 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
         // Calls NativeMinter precompile through INativeMinter interface.
         NATIVE_MINTER.mintNativeCoin(recipient, adjustedAmount);
     }
+
+    /**
+     * @dev Scales `value` based on `tokenMultiplier` and the direction of the transfer.
+     * Should be used for all tokens being transferred to/from other subnets.
+     */
+    function _scaleTokens(uint256 value, bool isReceive) private view returns (uint256) {
+        if (multiplyOnReceive == isReceive) {
```
cam-schultz, line 273: Can you add a comment above this conditional to help clarify the xor check? Something like "Scale multiplicitavely if we're scaling on the same side of the transfer as the contract instance multiplies". Then below, a comment like "Otherwise, scale by dividing the value by the multiplier".

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -236,4 +264,23 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
+            return value * tokenMultiplier;
+        } else {
```
cam-schultz, line 275: I think this else is unnecessary.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
cam-schultz: It's a matter of preference, but I think this function would be more readable (even at the cost of being more verbose) by restructuring the control flow to explicitly check the 4 possible cases in an `if-else` tree.

tests/flows/native_token_bridge.go
```diff
@@ -114,6 +114,8 @@ func NativeTokenBridge(network interfaces.LocalNetwork) {
 			sourceSubnet.BlockchainID,
 			bridgeContractAddress,
 			initialReserveImbalance,
+			big.NewInt(1),
```
cam-schultz, line 117: A multiplier > 1 would be good to test in the e2e flow.

cam-schultz (commented):
Generally makes sense to me, though I'm a bit wary of lossy integer division as others have commented. If we decide to restrict the multiplier to powers of 10, we could potentially simplify the implementation by chaing `tokenMultiplier` to a signed integer, and shifting when scaling rather than multiplying (e.g. -2 represents divide by 100; 4 represents multiply by 10,000)

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: I'm not sure if returning anything to the sender is a good idea here. Since we're dealing with native tokens, any extra operations are going to cost us more gas, and those operations are going to use up more gas than we're returning unless the multipliers are VERY large (Greater than 10^12).

tests/flows/native_token_bridge.go
geoff-vball: Since this feature is completely contained to `NativeTokenDestination`, I think it's probably fine for it to be tested by the unit tests, and leave the e2e tests to focus on the cross-chain interactions.

cam-schultz: Yeah, that's fair, though the e2e tests definitely should test the contract-specific features as well as cross-chain messaging. I think in general, we test basic contract functionality in the e2e tests, then dive deeper into edge cases in the unit tests.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -14,6 +14,9 @@ import {INativeTokenDestination} from "./INativeTokenDestination.sol";
 import {ITokenSource} from "./ITokenSource.sol";
 import {TeleporterFeeInfo, TeleporterMessageInput} from "@teleporter/ITeleporterMessenger.sol";
 import {TeleporterOwnerUpgradeable} from "@teleporter/upgrades/TeleporterOwnerUpgradeable.sol";
+// TeleporterUpgradeable import is used for natspec importdoc
+// solhint-disable-next-line no-unused-import
+import {TeleporterUpgradeable} from "@teleporter/upgrades/TeleporterUpgradeable.sol";
```
minghinmatthewlam, line 19: have you tested with the generated `solc` docs whether it's necessary to use `@inheritdoc`? Seems it might do so by default 
> Functions without NatSpec will automatically inherit the documentation of their base function. 

https://docs.soliditylang.org/en/latest/natspec-format.html

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -84,10 +147,14 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
         currentReserveImbalance = initialReserveImbalance_;
+
+        require(decimalsShift <= 18, "NativeTokenDestination: invalid decimalsShift");
+        tokenMultiplier = 10 ** decimalsShift;
+        multiplyOnReceive = multiplyOnReceive_;
     }
 
     /**
-     * @dev See {INativeTokenDestination-transferToSource}.
+     * @inheritdoc INativeTokenDestination
```
minghinmatthewlam, line 157: nit: I think we should keep this consistent throughout repo. Would recommend leaving and creating new issue, or changing it throughout repo.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -26,41 +29,101 @@ import {IAllowList} from "@avalabs/subnet-evm-contracts@1.2.0/contracts/interfac
...
+     * are transferred from the source chain into this destination chain, and divided by `tokenMultiplier` when
+     * tokens are transferred from this destination chain back to the source chain. This is intended
+     * when the "decimals" value on the source chain is less than the native EVM denomination of 18.
+     * If multiplyOnReceive is false, the raw token amount value will be divided by `tokenMultiplier` when tokens
+     * are transferred from the source chain into this destination chain, and multiplied by `tokenMultiplier` when
+     * tokens are transferred from this destination chain back to the source chain.
+     */
+    bool public immutable multiplyOnReceive;
+
     constructor(
```
minghinmatthewlam, line 119: might want some doc about `decimalShift`

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -101,11 +168,14 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
             currentReserveImbalance == 0, "NativeTokenDestination: contract undercollateralized"
         );
         uint256 value = msg.value;
-        require(value > 0, "NativeTokenDestination: zero transfer value");
+        uint256 scaledAmount = _scaleTokens(value, false);
+        require(scaledAmount > 0, "NativeTokenDestination: zero transfer value");
```
minghinmatthewlam, line 172:
```suggestion
        require(scaledAmount > 0, "NativeTokenDestination: zero scaled amount to transfer");
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -162,25 +236,25 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
                 allowedRelayerAddresses: allowedRelayerAddresses,
-                message: abi.encode(ITokenSource.SourceAction.Burn, abi.encode(totalBurnedTxFees))
+                message: abi.encode(ITokenSource.SourceAction.Burn, abi.encode(scaledAmount))
             })
         );
 
         emit ReportTotalBurnedTxFees({
             teleporterMessageID: messageID,
-            burnAddressBalance: totalBurnedTxFees
+            burnAddressBalance: scaledAmount
```
minghinmatthewlam, line 245: thinking if on the desteination chain we should keep the amounts in event emissions in the denomination of the scaled amount

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -151,7 +224,8 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
             );
         }
 
-        uint256 totalBurnedTxFees = BURNED_TX_FEES_ADDRESS.balance;
+        uint256 scaledAmount = _scaleTokens(BURNED_TX_FEES_ADDRESS.balance, false);
```
minghinmatthewlam, line 227: should we check if `scaledAmount` is zero here too?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -215,25 +287,43 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
         require(recipient != address(0), "NativeTokenDestination: zero recipient address");
         require(amount != 0, "NativeTokenDestination: zero transfer value");
 
+        uint256 scaledAmount = _scaleTokens(amount, true);
+
```
minghinmatthewlam, line 291: can you add comment if scaledAmount is zero we just return in the else block if there's `currentReserveImbalance`. Other option is to add an extra check if scaledAmount is zero before.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -215,25 +287,43 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
-                currentReserveImbalance -= amount;
-                emit CollateralAdded({amount: amount, remaining: currentReserveImbalance});
+                currentReserveImbalance -= scaledAmount;
+                emit CollateralAdded({amount: scaledAmount, remaining: currentReserveImbalance});
                 return;
             }
         }
 
         totalMinted += adjustedAmount;
         emit NativeTokensMinted(recipient, adjustedAmount);
```
minghinmatthewlam, line 309: Think these two lines should be moved under `if adjustedAmount > 0` check as well

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Maybe add both?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Michael had suggested emitting an event even in the 0 case, to make it more clear what is happening to the end user.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: refactored this a little bit. lmk if it makes sense.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: nvm, will just use the value for this chain.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: We're killing this function in another PR, so I'm not going to worry about it here

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -213,27 +280,47 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
 
         (address recipient, uint256 amount) = abi.decode(message, (address, uint256));
         require(recipient != address(0), "NativeTokenDestination: zero recipient address");
-        require(amount != 0, "NativeTokenDestination: zero transfer value");
+
+        uint256 scaledAmount = _scaleTokens(amount, true);
+        require(scaledAmount >= 0, "NativeTokenDestination: zero scaled amount received");
```
michaelkaplan13, line 285: I don't think this require is necessary since a `uint256` is always >= 0. If the scaled amount is 0, we won't call mint but still emit an event with an amount of 0, which I think is the correct behavior.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -213,27 +280,47 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
+            if (scaledAmount > currentReserveImbalance) {
                 emit CollateralAdded({amount: currentReserveImbalance, remaining: 0});
-                adjustedAmount = amount - currentReserveImbalance;
+                adjustedAmount = scaledAmount - currentReserveImbalance;
                 currentReserveImbalance = 0;
             } else {
-                currentReserveImbalance -= amount;
-                emit CollateralAdded({amount: amount, remaining: currentReserveImbalance});
-                return;
+                emit CollateralAdded({amount: scaledAmount, remaining: currentReserveImbalance});
```
michaelkaplan13, line 297: Are we emitting the stale `currentReserveImbalance` here now prior to updating it? Seems like we'd want to emit the updated value.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -213,27 +280,47 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
+                adjustedAmount = 0;
+                currentReserveImbalance -= scaledAmount;
             }
         }
 
-        totalMinted += adjustedAmount;
+        // emit an event even if we're minting 0 tokens to be clear to the user.
```
michaelkaplan13, line 303:
```suggestion
        // Emit an event even if the amount is zero to improve traceability. 
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/README.md
```diff
@@ -22,6 +22,7 @@ A contract that lives on the destination chain. Pairs with exactly one `TokenSou
...
+- `NativeTokenDestination` requires the setting of a `decimalsShift` in the constructor. This allows for the scaling of tokens sent to the source chain, which can be used to normalize the number of decimal places for the token between the two subnets. If a 1-to-1 relationship is desired, please set `decimalsShift` to 0. `tokenMultiplier` will be a publicly accessible field of the contract and will equal 10 to the power of `decimalsShift`. If `multiplyOnReceive` is set to true, the number of tokens transferred from the source chain into the destination chain will be multiplied by `tokenMultiplier`, and the number of tokens transferred from the destination chain back to the source chain will be divided by `tokenMultiplier`. If `multiplyOnReceive` is set to false, the number of tokens transferred from the source chain into the destination chain will be divided by `tokenMultiplier`, and number of tokens transferred from the destination chain back to the source chain will be multiplied by `tokenMultiplier`.
```
michaelkaplan13, line 25:
```suggestion
- `NativeTokenDestination` requires the setting of a `decimalsShift` value in the constructor. This allows for the scaling of token amounts sent from or to the source chain, and can be used to normalize the number of decimal places for the token between the two subnets. If a 1-to-1 relationship is desired, `decimalsShift` should be set to 0. `tokenMultiplier` is a publicly accessible field of the contract and is set to `10^decimalsShift`. If `multiplyOnReceive` is set to true, the number of tokens transferred from the source chain into the destination chain will be multiplied by `tokenMultiplier`, and the number of tokens transferred from the destination chain back to the source chain will be divided by `tokenMultiplier`. If `multiplyOnReceive` is set to false, the number of tokens transferred from the source chain into the destination chain will be divided by `tokenMultiplier`, and number of tokens transferred from the destination chain back to the source chain will be multiplied by `tokenMultiplier`.
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Yeah this was broken, fixed now.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -213,27 +280,46 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
 
         (address recipient, uint256 amount) = abi.decode(message, (address, uint256));
         require(recipient != address(0), "NativeTokenDestination: zero recipient address");
-        require(amount != 0, "NativeTokenDestination: zero transfer value");
+
```
michaelkaplan13, line 283: Unrelated to this PR (and definitely not a change to be in this PR), but when we port these contracts over to the `teleporter-token-bridge` repo, I'd suggest that we create a `_handleMint` private helper function here rather than having the logic directly in `_receiveTeleporterMessage`.

I think a good standard pattern for `_receiveTeleporterMessage` implementations is:
- authenticate sender
- decode message
- call private handle function for message type.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -213,27 +280,46 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
+        uint256 scaledAmount = _scaleTokens(amount, true);
 
         // If the contract has not yet been collateralized, we will deduct as many tokens
         // as needed from the transfer as needed. If there are any excess tokens, they will
         // be minted and sent to the recipient.
-        uint256 adjustedAmount = amount;
+        uint256 adjustedAmount = scaledAmount;
```
michaelkaplan13, line 289: Lets read `currentReserveImbalance` into a local variable here so that we don't read it from storage 3+ times. Ideally it would be one read and at most one write.

michaelkaplan13 (commented):
One final nit, and one future suggestion, but LGTM!

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Yep, I agree

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -287,15 +287,19 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
                 currentReserveImbalance = 0;
             } else {
-                currentReserveImbalance -= scaledAmount;
-                emit CollateralAdded({amount: scaledAmount, remaining: currentReserveImbalance});
+                emit CollateralAdded({
+                    amount: scaledAmount,
+                    remaining: reserveImbalance - scaledAmount
+                });
                 adjustedAmount = 0;
+                currentReserveImbalance -= scaledAmount;
```
michaelkaplan13, line 302: I think `-=` will actually do both a read and write to storage for `currentReserveImbalance`. I think it's probably best as:
```
uint256 updatedReserveImbalance = reserveImbalance - scaledAmount;
emit CollateralAdded({
    amount: scaledAmount,
    remaining: updatedReserveImbalance
});
currentReserveImbalance = updatedReserveImbalance;

Comment (geoff-vball) [reply contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol:302]: Good point! Thanks

File: contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol (hunk 20 above)
Comment (minghinmatthewlam) [reply contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol:116]: I see the `README` has a more detailed explanation, can you add a basic Natspec `@dev` for the overall constructor function.

File: contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol [hunk 33]
```diff
@@ -213,27 +280,48 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
-        if (currentReserveImbalance > 0) {
-            if (amount > currentReserveImbalance) {
-                emit CollateralAdded({amount: currentReserveImbalance, remaining: 0});
-                adjustedAmount = amount - currentReserveImbalance;
+        uint256 adjustedAmount = scaledAmount;
+        uint256 reserveImbalance = currentReserveImbalance;
+        if (reserveImbalance > 0) {
+            if (scaledAmount > reserveImbalance) {
+                emit CollateralAdded({amount: reserveImbalance, remaining: 0});
+                adjustedAmount = scaledAmount - reserveImbalance;
```
minghinmatthewlam, line 294: since `adjustedAmount=scaledAmount` I think we should just use `adjustedAmount -= reserveImbalance` in this `if/else` block, and not use `scaledAmount`. Easier to logic through imo

minghinmatthewlam: another option is in the line you declare `uint256 adjustedAmount = scaledAmount;` just make it `uint256 adjustedAmount;`, since you assign teh variable values in both `if/else` anyways

geoff-vball: Oh yes I like option 2, thanks!

michaelkaplan13 (approved):
LGTM

minghinmatthewlam (approved):
LGTM 🚢

Issue #306, opened 2024-03-03
[Bug]: Teleporter deployment failure


tactical-retreat:
**Describe the bug**
Deploying teleporter failed.

I got all the way through setting up a token bridge and enabling relayer, but the relayer kept failing with this error:

```
{"level":"error","timestamp":"2024-03-02T20:44:43.440-0500","logger":"awm-relayer","caller":"relayer/message_relayer.go:84","msg":"Failed to check if message should be sent","error":"no contract code at given address"}
```

Turns out, the teleporter contract doesn't exist on my subnet.

```
$ cast code  --rpc-url=https://testnet-rpc.ferdyflip.xyz/rpc 0xF7cBd95f1355f0d8d659864b92e2e9fbfaB786f7
0x
```

This surprised me, since I 'successfully' deployed it yesterday. The logs are gone though. So I tried to deploy the v0.1.0 version instead.

```
$ ./scripts/deploy_teleporter.sh --rpc-url https://testnet-rpc.ferdyflip.xyz/rpc --version v0.1.0 --fund-deployer $PRIVATE_KEY
TeleporterMessenger v0.1.0 contract address: 0x1b0505EaC728Efc30414ce0035BcC4DC5994a563
TeleporterMessenger v0.1.0 deployer address: 0xa21D65b03A07548F001b6C1422614BEa28370FdA
Funding deployer address with 10000000000000000000 wei

blockHash               0x40e82ce93c80a5b13b81dd85835acf3b7907d9273e6fed4a950f6bf160f7d3df
blockNumber             9907
contractAddress         
cumulativeGasUsed       21000
effectiveGasPrice       18000000000
from                    0xB34D4AE19393115CC5af8fa5f2b1cB4B5d682775
gasUsed                 21000
logs                    []
logsBloom               0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
root                    
status                  1
transactionHash         0x31cde0c7bba1590fd75d5c28b4bb651abc54853d0bbd973ca2073dcc3864e261
transactionIndex        0
type                    2
[...]

PR #309, opened 2024-03-04
Check deployment prior to broadcasting Nick's method tx

michaelkaplan13:
## Why this should be merged
Helps to avoid the issue described in #306. 

## How this works
Check that's the Teleporter deployer address is able to deploy the contract by simulating the transaction using `cast estimate`. 

## How this was tested
```
$ ./scripts/deploy_teleporter.sh --version v0.1.0 --rpc-url https://testnet-rpc.ferdyflip.xyz/rpc
TeleporterMessenger v0.1.0 contract address: 0x1b0505EaC728Efc30414ce0035BcC4DC5994a563
TeleporterMessenger v0.1.0 deployer address: 0xa21D65b03A07548F001b6C1422614BEa28370FdA
Error: 
(code: -32000, message: tx.origin 0xa21D65b03A07548F001b6C1422614BEa28370FdA is not authorized to deploy a contract, data: None)
```

## How this was documented
Error messages shown by script, and comments within it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
         run: |
           export PATH=$PATH:$HOME/.foundry/bin
           cd contracts/
-          forge build
+          forge build --force --extra-output-files bin
 
       - name: Build changelog
         id: build_changelog
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,5 +71,4 @@ jobs:
             ${{ env.deployer_addr_fn }}
             ${{ env.contract_addr_fn }}
             ${{ env.teleporter_messenger_bytecode_fn }}
-            ${{ env.teleporter_registry_bytecode_fn }}
     
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,8 @@ jobs:
       deployment_tx_fn: TeleporterMessenger_Deployment_Transaction_${{ github.ref_name }}.txt
       deployer_addr_fn: TeleporterMessenger_Deployer_Address_${{ github.ref_name }}.txt
       contract_addr_fn: TeleporterMessenger_Contract_Address_${{ github.ref_name }}.txt
+      teleporter_messenger_bytecode_fn: TeleporterMessenger_Bytecode_${{ github.ref_name }}.txt
+      teleporter_registry_bytecode_fn: TeleporterRegistry_Bytecode_${{ github.ref_name }}.txt
     steps:
       - name: Check out the repo
         uses: actions/checkout@v4
@@ -56,6 +58,8 @@ jobs:
           mv UniversalTeleporterDeployerTransaction.txt ${{ env.deployment_tx_fn }}
           mv UniversalTeleporterDeployerAddress.txt ${{ env.deployer_addr_fn }}
           mv UniversalTeleporterMessengerContractAddress.txt ${{ env.contract_addr_fn }}
+          mv contracts/out/TeleporterMessenger.sol/TeleporterMessenger.bin ${{ env.teleporter_messenger_bytecode_fn }}
+          mv contracts/out/TeleporterRegistry.sol/TeleporterRegistry.bin ${{ env.teleporter_registry_bytecode_fn }}
 
       - name: Create release
         uses: softprops/action-gh-release@v1
@@ -68,4 +72,6 @@ jobs:
             ${{ env.deployment_tx_fn }}
             ${{ env.deployer_addr_fn }}
             ${{ env.contract_addr_fn }}
+            ${{ env.teleporter_messenger_bytecode_fn }}
+            ${{ env.teleporter_registry_bytecode_fn }}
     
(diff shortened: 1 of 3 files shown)


geoff-vball (approved):

.github/workflows/release.yml
```diff
@@ -39,7 +40,7 @@ jobs:
         run: |
           export PATH=$PATH:$HOME/.foundry/bin
           cd contracts/
-          forge build
+          forge build --force --extra-output-files bin
```
minghinmatthewlam, line 43: why is `--force` and `--extra-output-files` needed now?

michaelkaplan13: `--force` isn't strictly needed, but I think good to add if for whatever reason output files from previous builds were around on the machine the workflow is run on (this should never be the case within Github Actions itself).

I added `--extra-output-files bin` so that the bytecode file (`.bin`) is generated to be uploaded with the release artifacts.

scripts/deploy_teleporter.sh
```diff
@@ -74,13 +85,21 @@ echo "TeleporterMessenger $teleporter_version contract address: $teleporter_cont
...
+if [[ $teleporter_contract_code != "0" ]]; then
     echo "TeleporterMessenger $teleporter_version has already been deployed on this chain." && exit 0
 fi
 
+# Estimate the amount of gas required to deploy the TeleporterMessenger bytecode from the Teleporter
+# deployer address in order to simulate the transaction. This will error if the TeleporterMessenger
+# contract is unable to be deployed from the deployer address.
+cast estimate --rpc-url $rpc_url \
+    --from $teleporter_deployer_address \
+    --create $teleporter_messenger_bytecode > /dev/null
```
cam-schultz, line 101: Any particular reason to suppress the output here?

In the error case, we should make sure this emits a clear error. The `cast` error might already be clear enough, but we may want to explicitly check the output and `echo` an error message of our own.

scripts/deploy_teleporter.sh
```diff
@@ -96,8 +115,18 @@ else
...
+# Deploy the TeleporterMessenger contract by publishing the raw Nick's method transaction.
 echo "Deploying TeleporterMessenger $teleporter_version"
-cast publish --rpc-url $rpc_url $teleporter_deploy_tx
+deployment_result=$(cast publish --rpc-url $rpc_url $teleporter_deploy_tx)
+deployment_status=$(echo $deployment_result | jq -r .status)
+deployment_tx_id=$(echo $deployment_result | jq -r .transactionHash)
+if [[ $deployment_status != "0x1" ]]; then 
+    echo "TeleporterMessenger deployment transaction failed. Transaction ID: $deployment_tx_id"
+    echo "Check failure reason and, if necessary, investigate deployment through state upgrade."
+    echo "See https://github.com/ava-labs/subnet-evm/tree/master/stateupgrade."
```
cam-schultz, line 126:
```suggestion
    echo "See https://github.com/ava-labs/subnet-evm/tree/master/stateupgrade"
```
Remove the period so that the shell emulator can generate a hyperlink (pure speculation on my part that this would make a difference)

scripts/deploy_teleporter.sh
michaelkaplan13: In the success case, I didn't think it was necessary to show the gas estimation output because the actual gas price used by Nick's method transaction is fixed since it's a legacy transaction, so I opted to suppress the stdout output.

The deployer allowlist error message was definitely clear enough from cast (seen below). I'm not 100% sure what other error cases may pop up, but another one that is helpful to have here is if the provided RPC URL is unreachable or invalid.

```
(code: -32000, message: tx.origin 0xa21D65b03A07548F001b6C1422614BEa28370FdA is not authorized to deploy a contract, data: None)
```

```
$ ./scripts/deploy_teleporter.sh --version v0.1.0 --rpc-url https://notablockchainendpoint.com
TeleporterMessenger v0.1.0 contract address: 0x1b0505EaC728Efc30414ce0035BcC4DC5994a563
TeleporterMessenger v0.1.0 deployer address: 0xa21D65b03A07548F001b6C1422614BEa28370FdA
Error: 
error sending request for url (https://notablockchainendpoint.com/): error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known

Context:
- Error #0: error trying to connect: dns error: failed to lookup address information: nodename nor servname provided, or not known
- Error #1: dns error: failed to lookup address information: nodename nor servname provided, or not known
- Error #2: failed to lookup address information: nodename nor servname provided, or not known
```

geoff-vball: Yeah I think this redirection should be fine. We're only redirecting `stdout`, not `stderr`, so we will still see any errors printed by `cast`

cam-schultz: Cool, agreed those error messages seem clear enough. I'm not seeing the line in the script in those messages though. That would be helpful to have.

geoff-vball (approved):

scripts/deploy_teleporter.sh
cam-schultz: Discussed with Michael offline, there's no easy workaround to emit the failing line number. In most cases it should be obvious from the error message itself what failed.

cam-schultz (approved):

PR #312, opened 2024-03-04
add required flags check to deploy script

felipemadero:
## Why this should be merged
required flags are not checked in the script, so the user get not so meaninful error msgs

## How this works
add the check

## How this was tested
executed the script with/without the flags

## How is this documented


michaelkaplan13:
We thought of this at the same time 👍 The checks were merged now in #309, so I'm going to close out this PR.

PR #325, opened 2024-03-05
Pin snyk action version

feuGeneA:
## How is this documented


cam-schultz (approved):
I'm generally in favor of version pinning.

PR #340, opened 2024-03-11
Fix NativeTokenDestination mint

### LINKED ISSUE #339: [Bug]: Native token bridge - `adjustedAmount` is always 0 when `currentReserveImbalance == 0`
**Describe the bug**
The bug was introduced in `v1.0.0` in the `_receiveTeleporterMessage` function of the `NativeTokenBridge/NativeTokenDestination.sol` contract.

The `adjustedAmount` used to mint tokens is only set if `reserveImbalance > 0` (and is thus 0 after the imbalance is solved). It is then impossible to bridge any token to the destination chain.

https://github.com/ava-labs/teleporter/blob/e545d5bd90d465649362884b9ed757b54c37bd29/contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol#L289-L312

**To Reproduce**
Deploy a native bridge, solve the imbalance, and try to bridge tokens to the destination chain.

**Expected behavior**
Tokens should be minted on the destination chain.

**Operating System**
Ubuntu 22.01 AMD64

**To best protect the Avalanche community security bugs should be reported in accordance to our [Security Policy](../../SECURITY.md)**

Nuttymoon:
## Why this should be merged

Fixes #339 

## How this works

Handle the case where `currentReserveImbalance == 0` to mint the right amount of tokens.

## How this was tested

Bridging tokens using `ERC20TokenSource` + `NativeTokenDestination` on a local network.

## How is this documented

NA


michaelkaplan13:
This was definitely a miss on our part. Thanks for catching it 🙏 

The issue was introduced in [this PR](https://github.com/ava-labs/teleporter/pull/291/files) that was merged last Friday. @geoff-vball could you add a unit test case that would have caught this and double check any other branches of the `NativeTokenDestination` contract that aren't covered using `forge test --coverage`?

Nuttymoon:
I missed the ABI bindings regeneration. Should be good now!

michaelkaplan13 (approved):
🙏 Thanks again. I'm down to merge this fix now with the unit tests that would have caught it originally as a fast follow as long as you are also @geoff-vball

geoff-vball (approved):

PR #344, opened 2024-03-12
Testnet E2E test cleanup

michaelkaplan13:
## Why this should be merged

Makes the E2E tests more robust when run against a live testnet.

## How this works

- Adds an extra wait condition in `waitMined` that checks that the `eth_blockNumber` RPC returns a block height at least as high as the block that the transaction was mined in. The `eth_blockNumber` RPC for endpoints behind a public load balancer are configured to return the lowest value that any node behind the load balancer is currently at. This means that subsequent RPC requests should all have seen the state update.
- Removes an unnecessary new client creation in `TraceTransaction`
- Fixes bugs in the `add_fee_amount` and `basic_send_receive` flows when run against a live testnet regarding how receipts are handled. Existing networks have prior receipt queue state, and cannot be relied on to return specific receipts in specific messages.
- Removes an unnecessary `bind.Opts` setting that did not have the proper functionality.

## How this was tested

Running the tests against the Fuji C-Chain, Dispatch, and Echo.

## How is this documented

N/A

### DIFF
--- a/tests/flows/basic_send_receive.go
+++ b/tests/flows/basic_send_receive.go
@@ -19,9 +19,7 @@ func BasicSendReceive(network interfaces.Network) {
 	teleporterContractAddress := network.GetTeleporterContractAddress()
 	fundedAddress, fundedKey := network.GetFundedAccountInfo()
 
-	//
 	// Send a transaction to Subnet A to issue a Warp Message from the Teleporter contract to Subnet B
-	//
 	ctx := context.Background()
 
 	// Clear the receipt queue from Subnet B -> Subnet A to have a clean slate for the test flow.
@@ -68,27 +66,21 @@ func BasicSendReceive(network interfaces.Network) {
 	)
 	expectedReceiptID := teleporterMessageID
 
-	//
 	// Relay the message to the destination
-	//
 	deliveryReceipt := network.RelayMessage(ctx, receipt, subnetAInfo, subnetBInfo, true)
 	receiveEvent, err := utils.GetEventFromLogs(
 		deliveryReceipt.Logs,
 		subnetBInfo.TeleporterMessenger.ParseReceiveCrossChainMessage)
 	Expect(err).Should(BeNil())
 
-	//
 	// Check Teleporter message received on the destination
-	//
 	delivered, err := subnetBInfo.TeleporterMessenger.MessageReceived(
 		&bind.CallOpts{}, teleporterMessageID,
 	)
 	Expect(err).Should(BeNil())
 	Expect(delivered).Should(BeTrue())
 
-	//
 	// Send a transaction to Subnet B to issue a Warp Message from the Teleporter contract to Subnet A
-	//
 	sendCrossChainMessageInput.DestinationBlockchainID = subnetAInfo.BlockchainID
 	sendCrossChainMessageInput.FeeInfo.Amount = big.NewInt(0)
 	receipt, teleporterMessageID = utils.SendCrossChainMessageAndWaitForAcceptance(
@@ -99,14 +91,12 @@ func BasicSendReceive(network interfaces.Network) {
 		fundedKey,
 	)
 
-	//
 	// Relay the message to the destination
-	//
 	deliveryReceipt = network.RelayMessage(ctx, receipt, subnetBInfo, subnetAInfo, true)
 
 	// Check that the receipt was received for expected Teleporter message ID
-	// This check is not performed for external networks because unrelated messages may have already changed
-	// the state of the receipt queues.
+	// This check is not performed for external networks because the specific receipt for this message
+	// may not have been included if the receipt queue had an existing build up of more than 5 messages.
 	if !network.IsExternalNetwork() {
 		Expect(utils.CheckReceiptReceived(
 			deliveryReceipt,
@@ -114,9 +104,7 @@ func BasicSendReceive(network interfaces.Network) {
 			subnetAInfo.TeleporterMessenger)).Should(BeTrue())
 	}
 
-	//
 	// Check Teleporter message received on the destination
-	//
 	delivered, err = subnetAInfo.TeleporterMessenger.MessageReceived(
 		&bind.CallOpts{}, teleporterMessageID,
 	)
@@ -124,8 +112,9 @@ func BasicSendReceive(network interfaces.Network) {
 	Expect(delivered).Should(BeTrue())
 
 	// If the reward address of the message from A->B is the funded address, which is able to send
-	// transactions on subnet A, then redeem the rewards.
-	if receiveEvent.RewardRedeemer == fundedAddress {
+	// transactions on subnet A, then redeem the rewards. This check is not performed for external
+	// networks since the specific receipt may not have been included in the message, as noted above.
+	if !network.IsExternalNetwork() && receiveEvent.RewardRedeemer == fundedAddress {
 		utils.RedeemRelayerRewardsAndConfirm(
 			ctx, subnetAInfo, feeToken, feeTokenAddress, fundedKey, feeAmount,
 		)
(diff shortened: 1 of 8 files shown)


tests/utils/utils.go
```diff
@@ -568,6 +571,35 @@ func WaitMined(ctx context.Context, b bind.DeployBackend, txHash common.Hash) (*
...
 		}
 	}
+
+	// Check that the block height endpoint returns a block height as high as the block number that the transaction was
+	// included in. This is to workaround the issue where multiple nodes behind a public RPC endpoint see
+	// transactions/blocks at different points in time. Ideally, all nodes in the network should have seen this block
+	// and transaction before returning from waitMined. The block height endpoint of public RPC endpoints is
+	// configured to return the lowest value currently returned by any node behind the load balancer, so waiting for
+	// it to be at least as high as the block height specified in the receipt should provide a relatively strong
+	// indication that the transaction has been seen widely throughout the network.
```
geoff-vball, line 581: So just double-checking my understanding: The loop above waits for the transaction to be mined, and be available on the node we happen to be hitting. Then the loop below here, will loop until that block has been seen by all nodes behind the load-balancer.

geoff-vball (approved):

tests/flows/check_upgrade_access.go
```diff
@@ -68,14 +68,11 @@ func CheckUpgradeAccess(network interfaces.Network) {
...
 	receipt := utils.WaitForTransactionSuccess(ctx, subnetInfo, tx.Hash())
-	blockNum := receipt.BlockNumber
 	pauseTeleporterEvent, err := utils.GetEventFromLogs(receipt.Logs, exampleMessenger.ParseTeleporterAddressPaused)
 	Expect(err).Should(BeNil())
 	Expect(pauseTeleporterEvent.TeleporterAddress).Should(Equal(teleporterAddress))
 
-	// We pass the block number to the call options because the Fuji C-chain endpoint is load balanced,
-	// and between consecutive calls we might be reaching different nodes with different states.
-	isPaused, err = exampleMessenger.IsTeleporterAddressPaused(&bind.CallOpts{BlockNumber: blockNum}, teleporterAddress)
+	isPaused, err = exampleMessenger.IsTeleporterAddressPaused(&bind.CallOpts{}, teleporterAddress)
```
cam-schultz, line 75: What was the issue with this `CallOpts`?

tests/utils/utils.go
```diff
@@ -542,16 +543,18 @@ func waitForTransaction(
...
 
+	var receipt *types.Receipt
 	for {
-		receipt, err := b.TransactionReceipt(cctx, txHash)
+		var err error
+		receipt, err = rpcClient.TransactionReceipt(cctx, txHash)
 		if err == nil {
-			return receipt, nil
+			break
 		}
```
cam-schultz, line 558: We can probably combine the two loops by nesting the block height check here. One possible implementation would be to set a `receivedReceipt` bool and check the block height if that bool is set, otherwise try to fetch the receipt. 

I'm not a huge fan of the back-to-back polling loops.

geoff-vball: I think it would be cleaner to just separate the two loops into distinct helper functions, one that waits on the receipt, and then one that wait for a specified block height

tests/flows/check_upgrade_access.go
michaelkaplan13: There's nothing wrong with specifying it, but if a node handled the request that had not yet seen that specified block height, it would still return a "cannot query unfinalized data" error. That error was more descriptive, but ultimately still results in a failed test.

tests/utils/utils.go
michaelkaplan13: Split the polls into helpers 👍

tests/testnet/network.go
```diff
@@ -346,12 +346,14 @@ func (n *testNetwork) getMessageDeliveryTransactionReceipt(
...
-		return nil, errors.New("Failed to find ReceiveCrossChainMessage log for relayed message")
+		return nil, errors.New("failed to find ReceiveCrossChainMessage log for relayed message")
 	} else if len(logs) > 1 {
-		return nil, errors.New("Found multiple ReceiveCrossChainMessage logs for relayed message")
+		return nil, errors.New("found multiple ReceiveCrossChainMessage logs for relayed message")
 	}
 
-	return destination.RPCClient.TransactionReceipt(ctx, logs[0].TxHash)
+	// The transaction should already be mined, but WaitMined will also wait for the eth_blockNumber
+	// endpoint to reflect the block that the transaction has been included in.
```
minghinmatthewlam, line 355: Outside of PR, for [checkMessageDelivered](https://github.com/ava-labs/teleporter/blob/main/tests/testnet/network.go#L284) can it be updated to take the `TeleporterMessenger` from `destination.SubnetInfo` instead of constructing new Teleporter

minghinmatthewlam: Also for the local network's `RelayMessage` can we add a similar `MessageReceived` check

tests/utils/utils.go
michaelkaplan13: Yep that's correct. I'm not sure how strong a guarantee it is that _all_ nodes have seen the block height returned by the load balancer, but it is definitely a good sanity check that in my testing caused additional waiting.

tests/flows/send_specific_receipts.go
```diff
@@ -96,8 +96,8 @@ func SendSpecificReceipts(network interfaces.Network) {
 	// Call send specific receipts to get reward of relaying two messages
 	receipt, messageID := utils.SendSpecifiedReceiptsAndWaitForAcceptance(
 		ctx,
-		subnetAInfo.BlockchainID,
 		subnetBInfo,
+		subnetAInfo.BlockchainID,
```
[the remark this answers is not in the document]
minghinmatthewlam: nvm just the order changed

tests/flows/basic_send_receive.go
```diff
@@ -99,33 +91,30 @@ func BasicSendReceive(network interfaces.Network) {
...
 	)
 	Expect(err).Should(BeNil())
 	Expect(delivered).Should(BeTrue())
 
 	// If the reward address of the message from A->B is the funded address, which is able to send
-	// transactions on subnet A, then redeem the rewards.
-	if receiveEvent.RewardRedeemer == fundedAddress {
+	// transactions on subnet A, then redeem the rewards. This check is not performed for external
+	// networks since the specific receipt may not have been included in the message, as noted above.
+	if !network.IsExternalNetwork() && receiveEvent.RewardRedeemer == fundedAddress {
```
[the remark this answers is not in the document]
michaelkaplan13: It should always be expected for the local network case right now, but I'd tend towards leaving it as a defensive check.

tests/testnet/network.go
michaelkaplan13: Good call, I removed `checkMessageDelivered` all together since it was reduced to a single function call with that change actually.

I don't think we need a `MessageReceived` check in `RelayMessage` given that we already find the event of the message being received in the transaction receipt.

tests/testnet/network.go
```diff
@@ -326,32 +315,32 @@ func (n *testNetwork) getMessageDeliveryTransactionReceipt(
...
 	if err != nil {
 		return nil, err
 	}
 
-	if len(logs) == 0 {
-		return nil, errors.New("Failed to find ReceiveCrossChainMessage log for relayed message")
-	} else if len(logs) > 1 {
-		return nil, errors.New("Found multiple ReceiveCrossChainMessage logs for relayed message")
+	// There should be exactly one matching event.
+	if !logIter.Next() {
```
minghinmatthewlam, line 333: I like to change to use the abi bindings filter, might be worth adding a comment that `iter.Next()` populates the `Event` value and needs to be called first.

minghinmatthewlam (approved):
LGTM thanks for cleaning up

cam-schultz (dismissed):

geoff-vball (dismissed):
👍

geoff-vball (approved):

cam-schultz (approved):

PR #346, opened 2024-03-12
Add test coverage for NativeTokenDestination

geoff-vball:
## Why this should be merged
Adds test coverage to NativeTokenDestination for the bug fixed in https://github.com/ava-labs/teleporter/pull/340.

## How this works
Reverts back to previous implementation of dealing with reserve balance. It uses an extra read/write for some operations before the reserve imbalance hits 0, but it's a rare code path that we don't need to optimize for, and I think this code structure works better.

## How this was tested
Tested unit tests against broken and fixed code. Added unit tests.

Coverage report shows this hits everything except one branch, but I swear it's being hit. If anyone minds taking a look I'd appreciate it.


contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
```diff
@@ -198,17 +217,20 @@ contract NativeTokenDestinationTest is NativeTokenBridgeTest {
...
         });
+
+        vm.expectEmit(true, true, true, true, address(nativeTokenDestination));
         emit NativeTokensMinted(_DEFAULT_RECIPIENT, firstTransfer);
 
         vm.expectCall(
             NATIVE_MINTER_PRECOMPILE_ADDRESS,
             abi.encodeWithSelector(INativeMinter.mintNativeCoin.selector),
-            1
+            2
```
minghinmatthewlam, line 233: can you put how `2` was computed instead of hardcoding it incase other values get changed.

geoff-vball: This is the number of calls we expect to the Native Minter precompile, so there's no variable to put here.

minghinmatthewlam: oh that's ncie I didn't even know `expectCall` had the amount of times, and can be set to 0.

minghinmatthewlam (approved):

cam-schultz (approved):
LGTM

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
```diff
@@ -139,6 +145,17 @@ contract NativeTokenDestinationTest is NativeTokenBridgeTest {
             )
         );
 
+        vm.expectEmit(true, true, true, true, address(nativeTokenDestination));
```
michaelkaplan13, line 148: Can't comment on the exact line, but I'm confused why the `expectCall` to the native minter precompile with a value of 0 is passing. Shouldn't we not call the precompile if the amount is 0?

michaelkaplan13: Aha! the 0 there is "expect 0 calls", nice. Could you add a comment clarifying that, I read it as "expect a call with a value of 0"

contracts/src/CrossChainApplications/examples/NativeTokenBridge/tests/NativeTokenDestinationTests.t.sol
michaelkaplan13: Could we break this up to two `expectCalls` that specify the specific expected amount to be minted in each call in the call data?

michaelkaplan13 (commented):
Thanks for adding these. Just two small nits.

cam-schultz (approved):

minghinmatthewlam (approved):

PR #347, opened 2024-03-12
NativeTokenDestination burn is a simple transfer

geoff-vball:
## Why this should be merged
Removes the separate burn mechanism for the native token bridge. This will be closer to the unified interface for moving these to the `teleporter-token-bridge` repo, and does most of the heavy lifting before implementing the reward mechanism for reporting burned tx fess.

## How this works
NativeTokenSource and ERC20TokenSource no longer receive separate burn messages.

NativeTokenDestination will take the difference between the last time it reported the amount of burned fees, and send an unlock message to send those fees to a burn address.

## How this was tested
Existing tests


contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -210,35 +205,23 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
     /**
      * @dev See {INativeTokenDestination-reportTotalBurnedTxFees}.
      */
-    function reportTotalBurnedTxFees(
-        TeleporterFeeInfo calldata feeInfo,
-        address[] calldata allowedRelayerAddresses
-    ) external {
-        uint256 adjustedFeeAmount;
```
minghinmatthewlam, line 217: why did we remove Teleporter fees here?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -210,35 +205,23 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
-                feeInfo: TeleporterFeeInfo({
-                    feeTokenAddress: feeInfo.feeTokenAddress,
-                    amount: adjustedFeeAmount
-                }),
-                requiredGasLimit: REPORT_BURNED_TOKENS_REQUIRED_GAS,
+                feeInfo: TeleporterFeeInfo({feeTokenAddress: address(0), amount: 0}),
+                requiredGasLimit: TRANSFER_NATIVE_TOKENS_REQUIRED_GAS,
                 allowedRelayerAddresses: allowedRelayerAddresses,
-                message: abi.encode(ITokenSource.SourceAction.Burn, abi.encode(scaledAmount))
+                message: abi.encode(GENERAL_BURN_ADDRESS, scaledAmount)
```
minghinmatthewlam, line 220: don't we want this to go to the blackhole address so that the sources don't need specific handling of burn report transfers?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -41,12 +40,10 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
     /**
-     * @notice Designated Blackhole Address for this contract.
-     *
-     * @dev Tokens are sent here to be "burned" before sending an unlock message to the source chain. Different
-     * from the burned tx fee address so they can be tracked separately.
+     * @dev Used when reporting burned gas fees to the source chain. Fees will be unlocked and sent to this address.
+     * Different from the burned tx fee address so they can be tracked separately.
      */
-    address public constant BURN_FOR_TRANSFER_ADDRESS = 0x0100000000000000000000000000000000010203;
+    address public constant GENERAL_BURN_ADDRESS = 0x0100000000000000000000000000000000010203;
```
minghinmatthewlam, line 46: nit: I prefer the old name, `TRANSFER_TO_SOURCE_BURN_ADDRESS` or similar. Also, can the comment above be updated.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: `subnet-evm` actually disables transfers the the blackhole address. So while we could send ERC20s to that address, we couldn't send native tokens there.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: We use it on both chains. On the destination chain as well to burn native tokens before transferring to the source chain.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/INativeTokenDestination.sol
```diff
@@ -40,7 +40,7 @@ interface INativeTokenDestination {
     /**
      * @dev Emitted when reporting total burned tx fees to source chain.
      */
-    event ReportTotalBurnedTxFees(bytes32 indexed teleporterMessageID, uint256 burnAddressBalance);
+    event ReportBurnedTxFees(bytes32 indexed teleporterMessageID, uint256 feesBurned);
```
michaelkaplan13, line 43: The dev comment here needs updating now.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -41,12 +40,10 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
     address public constant BURNED_TX_FEES_ADDRESS = 0x0100000000000000000000000000000000000000;
 
     /**
-     * @notice Designated Blackhole Address for this contract.
-     *
-     * @dev Tokens are sent here to be "burned" before sending an unlock message to the source chain. Different
-     * from the burned tx fee address so they can be tracked separately.
+     * @dev Used when reporting burned gas fees to the source chain. Fees will be unlocked and sent to this address.
+     * Different from the burned tx fee address so they can be tracked separately.
```
michaelkaplan13, line 44:
```suggestion
     * @notice Address used by this contract to blackhole funds, effectively burning them.
     * @dev Native tokens are burned by this contract by sending them to this address when transferring tokens back to
     * the source chain. When reporting burned transaction fee amounts, this address is used as the recipient
     * address for the funds to be sent to be burned on the source chain.
     * Different from BURNED_TX_FEES_ADDRESS so that the total amount burned in transaction fees and the amount 
     * total amount burned to be sent back to the source chain can be tracked separately. 
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -210,35 +205,23 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
-        TeleporterFeeInfo calldata feeInfo,
-        address[] calldata allowedRelayerAddresses
-    ) external {
-        uint256 adjustedFeeAmount;
-        if (feeInfo.amount > 0) {
-            adjustedFeeAmount = SafeERC20TransferFrom.safeTransferFrom(
-                IERC20(feeInfo.feeTokenAddress), feeInfo.amount
-            );
-        }
+    function reportBurnedTxFees(address[] calldata allowedRelayerAddresses) external {
```
michaelkaplan13, line 208: Comment for future consideration in a later PR:

I think we should consider having `allowedRelayerAddresses` be something that is set in Teleporter application constructors, with the managers of the application being allowed to update it. IMO, the use case for `allowedRelayers` is if an application wants to restrict who is able to relay it's messages to provide off-chain checks (on validity, compliance, etc). I don't think it makes sense for message senders (i.e. end users calling the applications) to be able to set it themselves, since they only care about anyone delivering their message, but not who.

michaelkaplan13: cc @minghinmatthewlam for you to think about also as we move forward with the Teleporter bridge contract implementations.

geoff-vball: Yeah I do like the idea of abstracting it away from the users.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
cam-schultz: I think it would be clearer if they were two separate variables, even if they contained the same address (I don't have a strong preference on if they should or not)

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
cam-schultz: `Used when reporting burned gas fees` seems more appropriate for `BURNED_TX_FEES_ADDRESS`

Are fees handled by this address? My understanding is that it's used to
1) burn tokens on the destination to be unlocked on the source
2) burn tokens on the source when reporting burned fees

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
cam-schultz: Can we leave the Teleporter fees in here then, and remove them in the follow up PR?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/TokenSource.sol
```diff
@@ -74,31 +71,13 @@ abstract contract TokenSource is ITokenSource, TeleporterOwnerUpgradeable {
             originSenderAddress == nativeTokenDestinationAddress, "TokenSource: unauthorized sender"
         );
 
-        // Decode the payload to recover the action and corresponding function parameters
-        (SourceAction action, bytes memory actionData) = abi.decode(message, (SourceAction, bytes));
-
         // Route to the appropriate function.
```
cam-schultz, line 74:
```suggestion
```

contracts/src/CrossChainApplications/examples/NativeTokenBridge/TokenSource.sol
```diff
@@ -74,31 +71,13 @@ abstract contract TokenSource is ITokenSource, TeleporterOwnerUpgradeable {
...
-            (address recipient, uint256 amount) = abi.decode(actionData, (address, uint256));
-            _unlockTokens(recipient, amount);
-        } else if (action == SourceAction.Burn) {
-            uint256 newBurnTotal = abi.decode(actionData, (uint256));
-            _handleBurnTokens(newBurnTotal);
-        } else {
-            revert("TokenSource: invalid action");
-        }
+        (address recipient, uint256 amount) = abi.decode(message, (address, uint256));
+        _unlockTokens(recipient, amount);
```
cam-schultz, line 76: We may want to rename this function to something more general since it now handles unlocks and burns (even though burning is implicit by sending to the burn address). Perhaps `_transferTokens`

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: would this make sense in `TeleporterUpgradeable`? Or do we want to leave it up to the dApps to implement and not set as a default

minghinmatthewlam: hmm what happens for a contract that gets called by other dApps, who have their individual allowed relayer addresses though?

contracts/src/CrossChainApplications/examples/NativeTokenBridge/TokenSource.sol
geoff-vball: I think `unlock` is the most general. We've locked the tokens, and are being told to unlock them to an address by the destination. We don't really care if that chain is "burning" them or not

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
michaelkaplan13: > would this make sense in TeleporterUpgradeable? Or do we want to leave it up to the dApps to implement and not set as a default

I think it likely makes sense as a separate utility. Along the lines of a `RestrictedRelayerSender`.

> hmm what happens for a contract that gets called by other dApps, who have their individual allowed relayer addresses though?

Sorry not sure I follow. The dApp calling `sendCrossChainMessage` is what gets to define the allowed relayers for its own messages since they may have their own restrictions.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -221,8 +230,8 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
             );
         }
 
-        uint256 amount = BURNED_TX_FEES_ADDRESS.balance;
-        uint256 scaledAmount = _scaleTokens(amount, false);
+        uint256 burnedDifference = BURNED_TX_FEES_ADDRESS.balance - latestBurnedFeesReported;
```
michaelkaplan13, line 233: We need to increment `latestBurnedFeesReported` here. Lets also add a unit test that would catch this case by calling `reportBurnedTxFees` multiple times. 

I think we should require that the `scaledAmount` is non-zero also.

cam-schultz (commented):
LGTM once the open comments are addressed

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: Lets say we have contract A that gets called by a bunch of other contracts B, C, D, E...If contract A constructor takes in empty allowed relayers list, so any relayers allowed, but the other contracts (B,C,D...) have their own `allowedRelayerAddresses` list they want to pass down to the call of contract A.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -230,8 +230,12 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
             );
         }
 
-        uint256 burnedDifference = BURNED_TX_FEES_ADDRESS.balance - latestBurnedFeesReported;
+        uint256 burnedAddressBalance = BURNED_TX_FEES_ADDRESS.balance;
+        uint256 burnedDifference = burnedAddressBalance - latestBurnedFeesReported;
+        latestBurnedFeesReported = burnedAddressBalance;
```
cam-schultz, line 235: Should this be `latestBurnedFeesReported=burnedDifference`?

geoff-vball: No, each call of this function should report the incremental change since the last time it was called, so we save the lastest `burnedAddressBalance`. I can rename the variable to `latestBurnedAddressBalance` if that would be more clear.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -230,8 +230,12 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
             );
         }
 
-        uint256 burnedDifference = BURNED_TX_FEES_ADDRESS.balance - latestBurnedFeesReported;
+        uint256 burnedAddressBalance = BURNED_TX_FEES_ADDRESS.balance;
+        uint256 burnedDifference = burnedAddressBalance - latestBurnedFeesReported;
```
cam-schultz, line 234: I think we should `require` that `burnedAddressBalance >= latestBurnedFeesReported`, or use safemath. If the destination somehow becomes corrupted such that `BURNED_TX_FEES_ADDRESS.balance` decreases, that would cause an underflow and, and up to INTMAX reported burned fees to the source chain.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Yeah, renamed.

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
geoff-vball: Yeah, I don't think this should be possible under any circumstance, but having the sanity check is worth it based on the catastrophic consequences.

cam-schultz: Thinking about it some more, a scenario in which the destination chain is corrupted such that the burn address is compromised, but arbitrary Teleporter messages are still not able to be sent seems not that likely. Still a good idea to prevent an invalid value here. Thanks for the fix.

cam-schultz (approved):

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
```diff
@@ -37,17 +36,29 @@ contract NativeTokenDestination is TeleporterOwnerUpgradeable, INativeTokenDesti
...
     address public constant BURN_FOR_TRANSFER_ADDRESS = 0x0100000000000000000000000000000000010203;
 
+    /**
+     * @notice Address used by the source chain to blackhole funds, effectively burning them.
+     *
+     * @dev When reporting burned transaction fee amounts, this address is used as the recipient
+     * address for the funds to be sent to be burned on the source chain.
+     * This address was chosen arbitrarily.
+     */
+    address public constant SOURCE_CHAIN_BURN_ADDRESS = 0x0100000000000000000000000000000000010203;
```
minghinmatthewlam, line 60: do we need this duplicate addrses variable? Can we just use `BURN_FOR_TRANSFER_ADDRESS` in reporting burned fees

geoff-vball: https://github.com/ava-labs/teleporter/pull/347#discussion_r1523699992

minghinmatthewlam (approved):

contracts/src/CrossChainApplications/examples/NativeTokenBridge/NativeTokenDestination.sol
minghinmatthewlam: Reading over again, I think i'm in agreement for dApps that want to restrict relayers, they can set it in their constructor. For dApps that don't care about restricted relayers, they can still take in `allowedRelayerAddresses` as a parameter in sends.

michaelkaplan13: Yeah I think it makes sense for the immediate caller of `sendCrossChainMessage` (i.e. the cross-chain dApp itself) to limit the allowed relayers if necessary. Exposing that option to callers of the dApp makes it possible for users of the cross-chain dApp to limit relayers for their own messages, but that's not something I would generally recommend for dApps, since users should not need to know of or about relayers at all.

Issue #342, opened 2024-03-11
Remove network restart from local setup


geoff-vball:
**Context and scope**
Remove the ability for the local network setup via `scripts/local/run.sh` to restart from where it left off. It has led to a confusing user-experience when users want to change network parameters. It also leads to issues when dependency versions are updated etc.

The utility from this function isn't very high, so we should remove it to make running the local network more reliable.

Should fix https://github.com/ava-labs/teleporter/issues/331

cam-schultz:
To fully clear out the old network, we should either delete all avalanche-cli files from the container, or delete the container altogether when the `run.sh` script exits. I've observed situations in which `avalanche network clean` is not sufficient to reset everything properly.

PR #350, opened 2024-03-13
Remove local network restart

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/342

## How this works
Removes the local network restart in `run_setup.sh`. This was causing issues and confusion with new users.

## How this was tested
Locally

## How is this documented
README


scripts/local/run_stop.sh
```diff
@@ -5,26 +5,13 @@
...
-if $clean
-then
-    echo "Cleaning network"
-    rm -f NETWORK_RUNNING
-fi
-
-docker compose -f docker/docker-compose-run.yml --project-directory ./ stop
\ No newline at end of file
+echo "Cleaning network"
+rm -f NETWORK_RUNNING
```
cam-schultz, line 15: Since we've removed `NETWORK_RUNNING`, this script just calls `docker compose stop`. How about we remove this script altogether and stop the containers from `run.sh` directly? I think that's the only remaining place we call `run_stop.sh`. Even without the specific call to `docker compose stop`, CTRL-C in the terminal that is running docker compose should stop all the containers.

docker/run_setup.sh
```diff
@@ -26,174 +26,164 @@ source $dir_prefix/scripts/utils.sh
...
-    subnet_c_subnet_id_hex=$(getBlockchainIDHex $subnet_c_subnet_id)
-    c_chain_blockchain_id_hex=$(getBlockchainIDHex $c_chain_blockchain_id)
-    c_chain_subnet_id_hex=$(getBlockchainIDHex $c_chain_subnet_id)
-    warp_messenger_precompile_addr=0x0200000000000000000000000000000000000005
-
-    # Write all vars to file so that they can be imported from another container
-    set > $dir_prefix/vars.sh
-
-    # Indicate for future runs that we started up successfully
-    touch $dir_prefix/NETWORK_RUNNING
```
cam-schultz, line 190: We should remove this from .gitignore.

scripts/local/run_stop.sh
cam-schultz: Additionally, I've oberved scenarios in which `avalanche network stop`, `avalanche network clean`, and `docker compose stop` leave the containers in an odd state for subsequent runs. The only reliable fix to this has been to remove the containers altogether.

Since creating the containers takes much less time than actually starting the network (which we do unconditionally), I think we could improve UX further by also unconditionally removing the containers. This can be done with `docker compose -f docker/docker-compose-run.yml --project-directory ./ rm -f`

cam-schultz (commented):
Thansk for cleaning this up. It's been an annoyance for awhile now. 

Left some suggestions on how to further improve UX by removing the containers altogether.

scripts/local/run_stop.sh
geoff-vball: Yeah, I left `run_stop.sh` so a user could call it to kill `run.sh`, which is probably only useful if it's called as a detached process, and in that case, the user should know how to kill it anyway. I'll remove it.

geoff-vball: AFAICT, `--force-recreate` should accomplish the same thing, so I've added that to the script.

cam-schultz: I think that should work too 👍

scripts/local/run.sh
```diff
@@ -17,9 +17,10 @@ function printHelp {
 }
 
 function cleanup {
-    echo "Shutting down network before exiting..."
-    ./scripts/local/run_stop.sh
-    echo "Network stopped"
+    echo "Shutting down and cleaning network before exiting..."
+    rm -f NETWORK_READY
+    docker compose -f docker/docker-compose-run.yml --project-directory ./ stop
+    echo "Network cleaned and stopped"
```
cam-schultz, line 23: The `trap` in `run_setup.sh` only calls `avalanche network stop`, so I think we should leave the `echo`s as is.

I could see a scenario in which someone stops the network by ctrl-c'ing out of the `run.sh` process, and then want to start the stopped containers to perform further manual testing.

We should document that subsequent calls of `run.sh` _will_ clean the network, but we should make it clear in our `echo`s exactly when that cleaning takes place, as well as document it in the README.

README.md
cam-schultz, line 1: Let's document that `run.sh` always force-recreates the containers

cam-schultz (approved):

michaelkaplan13 (approved):

PR #390, opened 2024-05-30
Update gas estimate for receiveCrossChainMessage

minghinmatthewlam:
## Why this should be merged
Fixes gas estimate of delivering `receiveCrossChainMessage` call.

## How this works
Previously we just had a static gas constant that was set higher for padding to avoid not having enough gas for delivery. But the gas cost is variable to message size, as well as number of Teleporter receipts included in the message. Now we add those two factors into account for the gas estimate, and also lower the default static gas cost.

## How this was tested
Decode gas cost was tested with a contract that only decodes, then kept passing in one more string character each time, and noticed ~2400 gas increase each time.

Teleporter message receipt gas was through unit test and receiving incrementally more receipts and noticing the increase in gas cost.

CI

## How is this documented
comments

### DIFF
--- a/tests/flows/relayer_modifies_message.go
+++ b/tests/flows/relayer_modifies_message.go
@@ -80,6 +80,7 @@ func relayAlteredMessage(
 	signedTx := createAlteredReceiveCrossChainMessageTransaction(
 		ctx,
 		signedWarpMessage,
+		&sendEvent.Message,
 		sendEvent.Message.RequiredGasLimit,
 		network.GetTeleporterContractAddress(),
 		fundedKey,
@@ -93,6 +94,7 @@ func relayAlteredMessage(
 func createAlteredReceiveCrossChainMessageTransaction(
 	ctx context.Context,
 	signedMessage *avalancheWarp.Message,
+	teleporterMesage *teleportermessenger.TeleporterMessage,
 	requiredGasLimit *big.Int,
 	teleporterContractAddress common.Address,
 	fundedKey *ecdsa.PrivateKey,
@@ -105,7 +107,12 @@ func createAlteredReceiveCrossChainMessageTransaction(
 	numSigners, err := signedMessage.Signature.NumSigners()
 	Expect(err).Should(BeNil())
 
-	gasLimit, err := gasUtils.CalculateReceiveMessageGasLimit(numSigners, requiredGasLimit, len(signedMessage.Bytes()))
+	gasLimit, err := gasUtils.CalculateReceiveMessageGasLimit(
+		numSigners,
+		requiredGasLimit,
+		len(signedMessage.Bytes()),
+		teleporterMesage,
+	)
 	Expect(err).Should(BeNil())
 
 	callData, err := teleportermessenger.PackReceiveCrossChainMessage(0, fundedAddress)
--- a/tests/flows/relayer_modifies_message.go
+++ b/tests/flows/relayer_modifies_message.go
@@ -94,7 +94,7 @@ func relayAlteredMessage(
 func createAlteredReceiveCrossChainMessageTransaction(
 	ctx context.Context,
 	signedMessage *avalancheWarp.Message,
-	teleporterMesage *teleportermessenger.TeleporterMessage,
+	teleporterMessage *teleportermessenger.TeleporterMessage,
 	requiredGasLimit *big.Int,
 	teleporterContractAddress common.Address,
 	fundedKey *ecdsa.PrivateKey,
@@ -111,7 +111,8 @@ func createAlteredReceiveCrossChainMessageTransaction(
 		numSigners,
 		requiredGasLimit,
 		len(signedMessage.Bytes()),
-		*teleporterMesage,
+		len(signedMessage.Payload),
+		len(teleporterMessage.Receipts),
 	)
 	Expect(err).Should(BeNil())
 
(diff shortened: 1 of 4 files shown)


utils/gas-utils/gas_utils.go
```diff
@@ -9,10 +9,13 @@ import (
...
 	"github.com/ava-labs/avalanchego/utils/math"
 	"github.com/ava-labs/subnet-evm/precompile/contracts/warp"
+	teleportermessenger "github.com/ava-labs/teleporter/abi-bindings/go/Teleporter/TeleporterMessenger"
 )
 
 const (
-	ReceiveCrossChainMessageStaticGasCost uint64 = 500_000
+	ReceiveCrossChainMessageStaticGasCost uint64 = 250_000
+	DecodeGasPerByte                      uint64 = 35
+	TeleporterMessageReceiptGas                  = 2500
```
michaelkaplan13, line 18:
```suggestion
	ReceiveCrossChainMessageBaseGasCost uint64 = 250_000
	MarkMessageReceiptGasCost           uint64 = 2500
	DecodeMessageGasCostPerByte         uint64 = 35
```

Nit: I think these names are slightly more consistent/accurate. "Static" was used previously, but technically it's not static so I think it should be renamed.

utils/gas-utils/gas_utils.go
```diff
@@ -23,13 +26,24 @@ const (
...
 	}
 
 	gasAmounts := []uint64{
 		executionRequiredGasLimit.Uint64(),
+		// The variable gas on message bytes is accounted for both when used in predicate verification
+		// and also when used in `getVerifiedWarpMessage`
+		uint64(numWarpMessageBytes) * warp.GasCostPerWarpMessageBytes * 2,
+		// Take into the variable gas cost for decoding the Teleporter message
+		// and marking the receipts as received.
+		uint64(len(teleporterMessage.Message)) * DecodeGasPerByte,
```
michaelkaplan13, line 45: There are 3 variable length parts of a teleporter message:
- The payload
- The receipts array (has a maximum of 5 * size of a single receipt)
- The `allowedRelayerAddresses`


Technically each of those could increase the gas code of the `abi.decode` call (although we're separately accounting for the number of receipts already), so should we account for them all here?

utils/gas-utils/gas_utils.go
```diff
@@ -23,13 +26,24 @@ const (
 // depends on the required limit for the message execution, the number of validator signatures
 // included in the aggregate signature, the static gas cost defined by the precompile, and an
 // extra buffer amount defined here to ensure the call doesn't run out of gas.
-func CalculateReceiveMessageGasLimit(numSigners int, executionRequiredGasLimit *big.Int) (uint64, error) {
+func CalculateReceiveMessageGasLimit(
+	numSigners int,
+	executionRequiredGasLimit *big.Int,
+	numWarpMessageBytes int,
```
michaelkaplan13, line 32:
```suggestion
	warpMessageSize int,
```
Super nit

tests/utils/utils.go
```diff
@@ -255,38 +255,6 @@ func CreateSendCrossChainMessageTransaction(
 	return SignTransaction(tx, senderKey, source.EVMChainID)
 }
 
-func CreateRetryMessageExecutionTransaction(
```
michaelkaplan13, line 258: Nice, didn't realize this was unused before 👍

tests/flows/relayer_modifies_message.go
```diff
@@ -105,7 +107,12 @@ func createAlteredReceiveCrossChainMessageTransaction(
 	numSigners, err := signedMessage.Signature.NumSigners()
 	Expect(err).Should(BeNil())
 
-	gasLimit, err := gasUtils.CalculateReceiveMessageGasLimit(numSigners, requiredGasLimit)
+	gasLimit, err := gasUtils.CalculateReceiveMessageGasLimit(
+		numSigners,
+		requiredGasLimit,
+		len(signedMessage.Bytes()),
```
michaelkaplan13, line 113: Should this be `len(signedMessage.Payload)`? Otherwise I think we are including the length of the aggregate signature and bit vector, which are not exposed within the EVM.

minghinmatthewlam: In `subnet-evm` the gas cost per Warp message byte is applied on each predicate byte, which includes the full signed Warp message bytes, so matched that for this estimate. https://github.com/ava-labs/subnet-evm/blob/4a75fa782de9c7b97bbcfb21b0211230234a4481/precompile/contracts/warp/config.go#L144

michaelkaplan13: Ah, nice catch. I didn't realize this.

utils/gas-utils/gas_utils.go
```diff
@@ -12,25 +12,44 @@ import (
...
-// included in the aggregate signature, the static gas cost defined by the precompile, and an
-// extra buffer amount defined here to ensure the call doesn't run out of gas.
-func CalculateReceiveMessageGasLimit(numSigners int, executionRequiredGasLimit *big.Int) (uint64, error) {
+// depends on the following:
+// - Required gas limit for the message execution
+// - The size of the Warp message
+// - The size of the Teleporter message included in the Warp message
+// - The number of Teleporter receipts
+// - Base gas cost for {receiveCrossChainMessage} call
+// -The number of validator signatures included in the aggregate signature
```
michaelkaplan13, line 31:
```suggestion
// - The number of validator signatures included in the aggregate signature
```

michaelkaplan13 (dismissed):
LGTM 🙏

michaelkaplan13 (approved):

geoff-vball (approved):

Issue #385, opened 2024-05-24
List of repo cleanup tasks


meeshhhh:
Issue to track list of cleanup items of various repos (avalanche-cli, avalanche-starter-kit, teleporter, token-bridge). Goal is to clearly define the purpose of each repo and make it clear where new pieces should be developed in the future.

For example: move the example applications from the `teleporter` repo to the `starter-kit` and add a link from `teleporter` repo. 

- [x] https://github.com/ava-labs/teleporter/pull/392
- [x] https://github.com/ava-labs/teleporter/issues/376
- [ ] https://github.com/ava-labs/teleporter/issues/378
    - Depends on:
        - [x] https://github.com/ava-labs/avalanche-cli/issues/1921
- [ ] https://github.com/ava-labs/teleporter/issues/377
    - Depends on:
        - [ ] https://github.com/ava-labs/avalanche-cli/issues/1901
- [ ] https://github.com/ava-labs/avalanche-starter-kit/issues/4
    - [ ] https://github.com/ava-labs/avalanche-starter-kit/issues/5
        - [x] https://github.com/ava-labs/teleporter/issues/381
- [ ] https://github.com/ava-labs/teleporter-token-bridge/issues/123
- [x] https://github.com/ava-labs/teleporter/issues/310 (migrate to tmpnet test framework)

PR #392, opened 2024-06-03
install avalanchego & subnetevm via avalanche-cli

feuGeneA:
## Why this should be merged
This eliminates the scripts in this repo that download and install avalanchego and subnetevm.

## How this works
It uses avalanche-cli to download those things for us.

## How this was tested
The e2e tests in CI now run with the avalanchego and subnetevm binaries downloaded and installed by CI.

## How is this documented
In the root README.md and in `.github/workflows/test.yml`.


scripts/local/e2e_test.sh
```diff
@@ -15,15 +15,34 @@ source "$TELEPORTER_PATH"/scripts/versions.sh
...
+  avalanche network clean
+fi
+
+# ensure that avalanche-cli has installed the version of subnetevm that we need:
+subnetevm_bin=${AVALANCHE_CLI_DIR}/bin/subnet-evm/subnet-evm-${SUBNET_EVM_VERSION}/subnet-evm
+if [[ ! -x "$subnetevm_bin" ]]; then
+  avalanche subnet create dummy --evm --vm-version "$SUBNET_EVM_VERSION" \
+    --force \
+    --teleporter --relayer --evm-chain-id 1 --evm-token DUM --evm-defaults # this line is just to avoid the dialog prompts
+fi
```
geoff-vball, line 36: Can you make the order of operations for these two consistent? I like 
Path
Comment
if block

cam-schultz (commented):
Does avalanche-cli have commands for directly installing the avalanchego and subnet-evm binaries? Manipulating the network just to install them seems like overkill and risks polluting or overwriting a user's avalanche-cli data.

I'm all for reducing the amount of bash we have to maintain, but am hesitant to do so if the alternative has additional overhead.

scripts/local/e2e_test.sh
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/392/commits/5c48264f2aedf30c812f49c362336172e3819a52

feuGeneA:
> Does avalanche-cli have commands for directly installing the avalanchego and subnet-evm binaries? Manipulating the network just to install them seems like overkill and risks polluting or overwriting a user's avalanche-cli data.
> 
> I'm all for reducing the amount of bash we have to maintain, but am hesitant to do so if the alternative has additional overhead.

@cam-schultz No, `avalanche-cli` does not have commands for installing those binaries. (That would be ideal...)

Aat the time I submitted this I hadn't yet realized the distinction between avalanche-cli and avalanche-network-runner, and I was envisioning that we would be incorporating avalanche-cli usage into this repo anyways (especially given all the "migrate to avalanche-cli" wording in the issues tracked by #385). Given that perspective, I was thinking that, if we're going to be migrating to avalanche-cli anyways, then there'd be no hope of avoiding the polluting/overwriting of the user's avalanche-cli data anyways, so the scripts in this PR seemed acceptable in this regard.

Having said all that, I'm happy to close this PR if that's what makes sense to the team.

feuGeneA:
see also https://github.com/ava-labs/avalanche-cli/issues/1938

geoff-vball (approved):

scripts/local/e2e_test.sh
```diff
@@ -15,15 +15,35 @@ source "$TELEPORTER_PATH"/scripts/versions.sh
...
+
+# Install the subnet-evm binary as a plugin within the avalanche-cli resources.
+# This is a hack that can probably be undone when we stop setting up the
+# avalanche network and subnet ourselves (via BeforeSuite/AfterSuite in
+# tests/local/e2e_test.go) and start having avalanche-cli do that for us
+# instead.
```
michaelkaplan13, line 43: Since we're going to use the `tempnet` fixture rather than `avalanche-cli` for use in the Ginkgo E2E tests, this comment should probably be updated.

cam-schultz:
> I was envisioning that we would be incorporating avalanche-cli usage into this repo anyways

I think incorporating avalanche-cli directly could be overkill. It's a useful interface, but all we really need is the local network it spins up. The scenario I'm considering is one in which a user maintains a Fuji or mainnet subnet named `dummy`, and when they go to run the e2e tests, they either fail or their existing subnet info is corrupted.

cam-schultz:
We discussed this offline, and came to the conclusion that the ideal solution would be to add install scripts directly to subnet-evm and avalanchego, and leverage them directly in our repos. A standalone install command in the CLI would also work, but is less preferable because the CLI is downstream of teleporter, so leveraging that in our tests would create a circular dependency. Barring that, the consensus is that we should leave the install scripts as-is, rather than add the CLI as a dependency and potentially cause conflicts with other CLI usage.

PR #397, opened 2024-06-07
Upgradeability for TeleporterUpgradeable and TeleporterOwnerUpgradeable

minghinmatthewlam:
## Why this should be merged
Fixes #420 #414 

## How this works
Makes it so the utility contracts are upgradeable, and can have easy non-upgradeable extensions that derive off these contracts in future issue.

## How this was tested
CI

## How is this documented


contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -177,6 +192,7 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
     {
         return _pausedTeleporterAddresses[teleporterAddress];
     }
+    // solhint-enable ordering
```
michaelkaplan13, line 195: Can we re-enable ordering right after `__TeleporterUpgradeable_init`?

abi-bindings/go/Teleporter/TeleporterMessenger/TeleporterMessenger.go
michaelkaplan13, line 1: Any idea why this ABI binding changed despite the contract being untouched?

michaelkaplan13: Looks like it's just the IPFS hash of the deployed byte code that changed, but still seems odd since nothing else was modified as far as I can tell.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
```diff
@@ -60,7 +69,13 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
...
      * instance and setting `_minTeleporterVersion`.
      */
-    constructor(address teleporterRegistryAddress) {
+    // solhint-disable ordering
+    // solhint-disable-next-line func-name-mixedcase
+    function __TeleporterUpgradeable_init(address teleporterRegistryAddress)
+        internal
+        onlyInitializing
+    {
+        __ReentrancyGuard_init();
```
michaelkaplan13, line 78: Just for my own understanding, do you know why `ContextUpgradeable` doesn't have a `__ContextUpgradeable_init` function to be called here?

michaelkaplan13 (commented):
Left a few clarifying questions, but the changes LGTM. Looking forward to adding in the namespaced storage layout and non-upgradeable extensions.

Given the intermediary nature of these changes, I think it could make sense to keep them in a dev branch that includes all of the upgradeability related changes prior to merging into `main`, since I don't think we intended for anyone to use this current versions even for testing purposes.

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: `ContextUPgradeable` has no state to initialize with so the init functions are empty
```
    function __Context_init() internal onlyInitializing {
    }

    function __Context_init_unchained() internal onlyInitializing {
    }
```

michaelkaplan13: Gotcha, should we include calls to that here for completeness?  In the event that state is added to those contracts in a future version I image we'd want to call it.

contracts/src/Teleporter/tests/TestMessenger.sol
```diff
@@ -20,7 +21,7 @@ import {ReentrancyGuard} from "@openzeppelin/contracts@4.8.1/security/Reentrancy
  * @dev TestMessenger is test fixture contract that exercises sending and receiving
  * messages cross chain.
  */
-contract TestMessenger is ReentrancyGuard, TeleporterOwnerUpgradeable {
+contract TestMessenger is ReentrancyGuardUpgradeable, TeleporterOwnerUpgradeable {
```
cam-schultz, line 24: We should think of a different name for `TeleporterOwnerUpgradeable`. Right now, we're mixing the meaning of `*Upgradeable` within the same codebase, which is confusing. As an illustration, to be consistent with the names of other contracts, the name should be `TeleporterOwnerUpgradeableUpgradeable`.

Maybe `UpgradeableTeleporter`? Or `UpgradeableTeleporterReceiver`?  Or bake the registry directly into the name, something like `TeleporterRegistryReceiver`.

bernard-avalabs (commented):
Generally makes sense to me.

contracts/src/Teleporter/tests/TestMessenger.sol
minghinmatthewlam: `TeleporterRegistryConsumer` @michaelkaplan13 suggestion also, or can be `TeleporterRegistryApp`

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: nope the linter still picks up the error for the following public functions

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: still need to follow up this. If you include the Context init call, for new versions we'd need to check through the inheritance hierarchy to make sure there's no duplicates of initialize calls anyways. Likely if called, we can call the `init_unchained` call. Also was looking at OZ's [OwnableUpgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/OwnableUpgradeable.sol#L51) which also inherits `ContextUpgradeable` but they do not call the Context init function, so might ask OZ about that.

contracts/src/Teleporter/tests/TestMessenger.sol
minghinmatthewlam: created separate issue https://github.com/ava-labs/teleporter/issues/421

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: was wrong about this, and will be changing some of the lint disable/enables, possibly in namespace pr

contracts/foundry.toml
```diff
@@ -2,7 +2,7 @@
 src = 'src'
 out = 'out'
 libs = ['lib']
-solc_version = '0.8.18'
+solc_version = '0.8.20'
```
michaelkaplan13, line 5: Lets go straight to `0.8.23` 👍

michaelkaplan13 (commented):
One suggestion, and one question that applies to a couple of places, but otherwise LGTM.

contracts/src/teleporter/registry/tests/TeleporterRegistryTests.t.sol
```diff
@@ -321,7 +319,7 @@ contract TeleporterRegistryTest is Test {
             payload: abi.encode(
                 ProtocolRegistryEntry({version: version, protocolAddress: protocolAddress}),
                 registryAddress
-                )
+            )
```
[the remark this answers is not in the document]
minghinmatthewlam: was there a bug before? Looks like just a formatting change to me haha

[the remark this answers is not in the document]
geoff-vball: The autoformatter had a bug where it was putting the closing paren in the wrong spot. It seems correct now :)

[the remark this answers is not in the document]
minghinmatthewlam: oh I see thought there was a bug in the test previously, sweet.

contracts/src/teleporter/registry/TeleporterOwnerUpgradeable.sol
```diff
@@ -3,25 +3,31 @@
...
+abstract contract TeleporterOwnerUpgradeable is TeleporterUpgradeable, OwnableUpgradeable {
+    // solhint-disable-next-line func-name-mixedcase
+    function __TeleporterOwnerUpgradeable_init(
         address teleporterRegistryAddress,
         address initialOwner
-    ) TeleporterUpgradeable(teleporterRegistryAddress) {
-        transferOwnership(initialOwner);
+    ) internal onlyInitializing {
+        __TeleporterUpgradeable_init(teleporterRegistryAddress);
+        __Ownable_init_unchained(initialOwner);
```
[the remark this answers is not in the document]
minghinmatthewlam: you're right, I was originally thinking to use `unchained` to avoid cases where `Ownable_init` gets some added init's in its inheritance graph and to avoid collisions for new versions, but think the standard is to stick with `init` unless necessary to avoid collisions with `unchained`.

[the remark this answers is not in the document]
minghinmatthewlam: updated

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -56,19 +73,49 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
...
-    constructor(address teleporterRegistryAddress) {
+    // solhint-disable ordering
+    // solhint-disable-next-line func-name-mixedcase
+    function __TeleporterUpgradeable_init(address teleporterRegistryAddress)
+        internal
+        onlyInitializing
+    {
+        // TODO: figure out whether best practice to call ContextUpgradeable init, even though it's empty
+        // OZ Ownable inherits ContextUpgradeable but does not call ContextUpgradeable init
+        __ReentrancyGuard_init_unchained();
```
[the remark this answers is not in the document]
minghinmatthewlam: replied in https://github.com/ava-labs/teleporter/pull/397#discussion_r1684459392, and addressed

contracts/src/Teleporter/upgrades/TeleporterUpgradeable.sol
minghinmatthewlam: following convention in https://github.com/ava-labs/teleporter/pull/397#discussion_r1684459392 to call `init` instead of `unchained` when not needed, went forward to call `Context_init` as well for completeness, and no duplicate initializations.

minghinmatthewlam: Related OZ issue with their upgradeable contract transpiler not always calling the `init` functions of parent contracts https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4690

.gitmodules
```diff
@@ -11,3 +11,6 @@
 [submodule "contracts/lib/subnet-evm"]
 	path = contracts/lib/subnet-evm
 	url = https://github.com/ava-labs/subnet-evm
+[submodule "contracts/lib/openzeppelin-contracts-upgradeable"]
+	path = contracts/lib/openzeppelin-contracts-upgradeable
+	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
```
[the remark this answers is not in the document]
cam-schultz: FWIW the submodules will need to be reviewed again in https://github.com/ava-labs/teleporter/pull/422, which is next in the pipeline

cam-schultz (dismissed):
LGTM. Calling out that it would be good to update the upgrades [UML](https://github.com/ava-labs/teleporter/blob/main/contracts/src/teleporter/registry/upgrade-uml.png) , which appears to be stale even before this change. We can do that in a followup issue.

bernard-avalabs (dismissed):
LGTM

michaelkaplan13:
Bumping this comment that may have fallen through the cracks: https://github.com/ava-labs/teleporter/pull/397#discussion_r1684507517

feuGeneA (commented):
Do we need to call `_disableInitializers()` somewhere? https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializing_the_implementation_contract

Do we plan to integrate OZ's [Upgrades Plugins](https://docs.openzeppelin.com/upgrades-plugins/1.x/foundry-upgrades)? In particular, the "upgrade safety validations" sound helpful.

geoff-vball (commented):
LGTM once Gene's comments are addressed

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -175,7 +226,8 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
         virtual
         returns (bool)
     {
-        return _pausedTeleporterAddresses[teleporterAddress];
+        TeleporterUpgradeableStorage storage $ = _getTeleporterUpgradeableStorage();
```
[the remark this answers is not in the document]
minghinmatthewlam: updated with private and external functions

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -23,21 +27,34 @@ import {IERC20} from "@openzeppelin/contracts@4.8.1/token/ERC20/IERC20.sol";
...
 
-    // The Teleporter registry contract manages different Teleporter contract versions.
-    TeleporterRegistry public immutable teleporterRegistry;
-    /**
-     * @dev A mapping that keeps track of paused Teleporter addresses.
-     */
-    mapping(address teleporterAddress => bool paused) private _pausedTeleporterAddresses;
+    // solhint-disable private-vars-leading-underscore
+    /// @custom:storage-location erc7201:teleporter.storage.TeleporterUpgradeable
+    struct TeleporterUpgradeableStorage {
```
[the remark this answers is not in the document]
minghinmatthewlam: updated comment

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -23,21 +27,34 @@ import {IERC20} from "@openzeppelin/contracts@4.8.1/token/ERC20/IERC20.sol";
...
+        uint256 _minTeleporterVersion;
+    }
+    // solhint-enable private-vars-leading-underscore
 
-    /**
-     * @dev The minimum required Teleporter version that the contract is allowed
-     * to receive messages from. Should only be updated by `_setMinTeleporterVersion`
-     */
-    uint256 private _minTeleporterVersion;
+    // keccak256(abi.encode(uint256(keccak256("teleporter.storage.TeleporterUpgradeable")) - 1)) & ~bytes32(uint256(0xff));
```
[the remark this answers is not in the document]
minghinmatthewlam: added comment

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -56,19 +73,48 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
      */
     event TeleporterAddressUnpaused(address indexed teleporterAddress);
 
+    function _getTeleporterUpgradeableStorage()
+        private
+        pure
+        returns (TeleporterUpgradeableStorage storage $)
```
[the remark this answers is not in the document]
minghinmatthewlam: I prefer to leave as is, mostly because it is the referenced example in ERC7201, and also throughout OZ contracts it is also the convention. In terms of `$` looking like a special symbol, can help point out that `$` is not like regular local variables, and is only used in reference to the storage variable.

minghinmatthewlam:
> Do we need to call `_disableInitializers()` somewhere? https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#initializing_the_implementation_contract
> 
> Do we plan to integrate OZ's [Upgrades Plugins](https://docs.openzeppelin.com/upgrades-plugins/1.x/foundry-upgrades)? In particular, the "upgrade safety validations" sound helpful.

I deferred the `_disableInitializer` to the non-upgradeable extensions Issue/PR, but yes they should be added to constructors of upgradeable contracts.

contracts/src/teleporter/registry/TeleporterOwnerUpgradeable.sol
```diff
@@ -3,25 +3,31 @@
...
+abstract contract TeleporterOwnerUpgradeable is TeleporterUpgradeable, OwnableUpgradeable {
+    // solhint-disable-next-line func-name-mixedcase
+    function __TeleporterOwnerUpgradeable_init(
         address teleporterRegistryAddress,
         address initialOwner
-    ) TeleporterUpgradeable(teleporterRegistryAddress) {
-        transferOwnership(initialOwner);
+    ) internal onlyInitializing {
+        __TeleporterUpgradeable_init(teleporterRegistryAddress);
+        __Ownable_init(initialOwner);
```
michaelkaplan13, line 25: Just confirming that the OZ standard is to not call `XXX_init_unchained()` at the end of `XXX_init()` if the unchained call is a no-op as it is here?

Doesn't make a functional difference, but want to make sure we're consistent with it going forward.

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -56,19 +80,48 @@ abstract contract TeleporterUpgradeable is Context, ITeleporterReceiver, Reentra
...
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            $.slot := _TELEPORTER_UPGRADEABLE_STORAGE_LOCATION
+        }
+    }
+
     /**
      * @dev Initializes the {TeleporterUpgradeable} contract by getting `teleporterRegistry`
-     * instance and setting `_minTeleporterVersion`.
+     * instance and setting `_minTeleporterVersion`. lkjl;kj;
```
michaelkaplan13, line 96: Blooper here

.github/workflows/abi_bindings_checker.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           export PATH=$PATH:$HOME/.foundry/bin
           export GOPATH=$HOME/go
           export PATH="$PATH:$GOPATH/bin"
+          forge --version
```
michaelkaplan13, line 35: We can remove this now

contracts/src/teleporter/registry/TeleporterOwnerUpgradeable.sol
minghinmatthewlam: they do not call it from `ContextUpgradeable`. But imo the OZ contract upgradeabls do not have many standards due to the limitation of customizability of their transpiler, so think we can set standards of our own. This would be a tradeoff for small gas increase for internal call, vs preventing future possibility of missing the `unchained` call if we add logic there. No strong pref

cam-schultz (approved):

contracts/src/teleporter/registry/TeleporterOwnerUpgradeable.sol
michaelkaplan13: Fine with leaving it as is. Any change that would make it possibly incorrect is a change to this contract itself (not just updating a version of a parent contract or the like).

michaelkaplan13 (approved):

PR #401, opened 2024-06-14
migrate to tmpnet & latest subnetevm

### LINKED ISSUE #310: Update subnet-evm to v0.6.2
**Context and scope**
Update subnet-evm dependency to v0.6.2

This version removes the `tests/utils/runner` module, in favor of `tests/fixture/e2e` and `tests/fixture/tmpnet`. We will need to adapt `network.go` to use the new packages.

Diff for the latest version of `subnet-evm`
https://github.com/ava-labs/subnet-evm/compare/v0.6.1...v0.6.2

feuGeneA:
## Why this should be merged
Fixes #310

## How this works

Stops using `avalanche-network-runner`, and uses `tmpnet` instead.

## How this was tested

By getting the existing e2e tests in this repo to pass.

## How is this documented


go.mod
```diff
@@ -3,14 +3,13 @@ module github.com/ava-labs/teleporter
 go 1.21.12
 
 require (
-	github.com/ava-labs/avalanchego v1.11.1
+	github.com/ava-labs/avalanchego v1.11.9-0.20240703222923-9d71967ccbba
```
feuGeneA, line 6: `9d719...` refers to the latest commit hash (as of this writing) in [this PR](https://github.com/ava-labs/avalanchego/pull/3179). We should change this to refer to a released version, once that PR has been merged and released.

geoff-vball (commented):
I know this is still in draft, but what you have now is looking really good :)

tests/utils/warp-genesis.json
```diff
@@ -22,7 +22,7 @@
       "blockGasCostStep": 500000
     },
     "warpConfig": {
-      "blockTimestamp": 0
+      "blockTimestamp": 1719343601
```
minghinmatthewlam, line 25: why do we need to specify a specific warp blocktimestamp, as well as the timestamp value below?

tests/local/network.go
```diff
@@ -453,74 +434,146 @@ func (n *LocalNetwork) RelayMessage(ctx context.Context,
...
+	// consume some of the extraNodes
+	newValidatorNodes := n.extraNodes[0:count]
+	slices.Delete(n.extraNodes, 0, int(count))
+
+	apiURI, err := n.tmpnet.GetURIForNodeID(subnet.ValidatorIDs[0])
 	Expect(err).Should(BeNil())
 
-	// Add the new node names
-	n.subnetNodeNames[subnetID] = append(n.subnetNodeNames[subnetID], nodeNames...)
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
geoff-vball, line 473: is 30 seconds longs enough for adding the validators and restarting the nodes on any machine?

tests/local/network.go
```diff
@@ -453,74 +434,146 @@ func (n *LocalNetwork) RelayMessage(ctx context.Context,
...
+		defer cancel()
+		err := node.SaveAPIPort()
+		Expect(err).Should(BeNil())
+
+		err = node.Stop(ctx)
+		Expect(err).Should(BeNil())
+
+		err = n.tmpnet.StartNode(ctx, os.Stdout, node)
 		Expect(err).Should(BeNil())
 	}
```
geoff-vball, line 530: How long does this take? If it's long, could we do each iteration of this in a goroutine to speed it up?

tests/utils/warp-genesis.json
feuGeneA: I was experiencing a hang in the test setup, and @ceyonur said "warp cannot be activated at timestamp = 0...also set genesis timestamp  to some non-zero value (a recent timestamp is fine)...it was changed after v0.6.3 when we introduced modifiable network upgrades and fixing a regression with using 0 timestamps in genesis." starting from https://platwizards.slack.com/archives/C075HLQL4BT/p1719327426806459?thread_ts=1718898895.417379&cid=C075HLQL4BT

it looks like this is the original issue as reported: https://github.com/ava-labs/avalanchego/issues/2877, which was fixed by https://github.com/ava-labs/subnet-evm/pull/1136 and included in https://github.com/ava-labs/subnet-evm/releases/tag/v0.6.4

.github/workflows/test.yml
```diff
@@ -69,3 +69,14 @@ jobs:
...
+      - name: Upload tmpnet network dir
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: e2e-tmpnet-data
+          path: |
+            ~/.tmpnet/networks
+            ~/.tmpnet/prometheus/prometheus.log
+            ~/.tmpnet/promtail/promtail.log
+          if-no-files-found: error
```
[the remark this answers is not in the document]
feuGeneA: the prometheus stuff I just copied along with the rest of it from avalanchego 😅 but the `~/.tmpnet/networks` part is the most pertinent: that folder contains all of the logs, configs, etc for the network and all the nodes started for it. tmpnet chooses not to dump all of those logs to stdout, instead putting them on disk in that folder (a different sub-folder per tmpnet run). uploading that folder allows us to see the full diagnostic info for a test run, since it's not all included in the direct output of the CI job.

tests/local/network.go
feuGeneA: it's a good question, but I'm not sure of the answer. all I can say is that it's good enough for my azure dev env, and for CI. (I've never seen this timeout occur, either locally or in CI.) if I remember correctly, the call that uses this context refuses to run if there isn't _some_ timeout (or deadline) specified. I'm happy to bump this up to whatever you think is reasonable.

tests/local/network.go
feuGeneA: it's not long. the `StartNode` call here doesn't wait for the node to start up, it's just sort of "fire and forget". 

I found some output in my terminal from the last time I ran this (on my azure dev box), and between the log statements just before and just after this loop, 3.574 had elapsed.

feuGeneA: also meant to add: note that in the loop just after this one, that's where we wait for all the nodes to come up, and that part in theory takes longer. I don't have a convenient log history to quantify _that_ wait, but I can say that that wait AND a subsequent call to `subnetEvmUtils.IssueTxsToActivateProposerVMFork()` in total took 8.486 seconds.

tests/local/network.go
geoff-vball: Yeah, let's bump it that should be too long to fail if things are working, but not too long that we're waiting around forever if something is broken. 60-120 seconds should be fine.

.github/workflows/test.yml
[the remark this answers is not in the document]
geoff-vball: Hmm... Would it make sense to tail the log file with a goroutine so the test suite outputs the output to stdout?

tests/flows/teleporter_registry.go
```diff
@@ -76,8 +75,9 @@ func TeleporterRegistry(network interfaces.LocalNetwork) {
 	utils.SetChainConfig(chainConfigs, subnetAInfo, warpEnabledChainConfigA)
 
 	// Restart nodes with new chain config
-	nodeNames := network.GetAllNodeNames()
-	network.RestartNodes(ctx, nodeNames, runner_sdk.WithChainConfigs(chainConfigs))
+	nodeIDs := network.GetAllNodeIDs()
+	network.SetChainConfigs(chainConfigs)
```
michaelkaplan13, line 79: For my own understanding, what's the difference between this call to `network.SetChainConfigs` and the calls to `utils.SetChainConfig` above? Can we consolidate them/remove one of them, or explain why both are needed in a comment here?

tests/local/network.go
```diff
@@ -91,81 +85,66 @@ func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
 	_, err = os.Stat(warpGenesisFile)
 	Expect(err).Should(BeNil())
 
-	anrConfig := runner.NewDefaultANRConfig()
-	anrConfig.GlobalCChainConfig = warpEnabledChainConfig
-	manager := runner.NewNetworkManager(anrConfig)
+	subnetA := subnetEvmTestUtils.NewTmpnetSubnet(
+		"A",
+		warpGenesisFile,
```
michaelkaplan13, line 90: It looks like this was probably existing already, but lets use a different EVM chain ID for each of the subnets that get created here. Otherwise the local chains will be subject to possible replayed transactions or transactions accidentally being sent to the wrong chain being valid.

It'd probably be easiest to make the `warpGenesisFile` a template with a <EVM_CHAIN_ID>, and provide a util for creating a temp file with a given chain ID value.

tests/local/network.go
```diff
@@ -42,15 +44,13 @@ var _ interfaces.LocalNetwork = &LocalNetwork{}
 type LocalNetwork struct {
 	teleporterContractAddress common.Address
 	primaryNetworkInfo        *interfaces.SubnetTestInfo
-	subnetAID, subnetBID      ids.ID
 	subnetsInfo               map[ids.ID]*interfaces.SubnetTestInfo
-	subnetNodeNames           map[ids.ID][]string
+	extraNodes                []*tmpnet.Node
```
michaelkaplan13, line 48: Could you add a comment here explaining what `extraNodes` is intended to be used for, and the requirements for its use?

tests/local/network.go
```diff
@@ -91,81 +85,66 @@ func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
...
 	Expect(err).Should(BeNil())
-
-	logFactory := logging.NewFactory(logging.Config{
-		DisplayLevel: logLevel,
-		LogLevel:     logLevel,
-	})
-	zapLog, err := logFactory.Make("main")
+	fundedKeySecp256k1, err := secp256k1.ToPrivateKey(fundedKeyBytes)
 	Expect(err).Should(BeNil())
+	network.PreFundedKeys = append(network.PreFundedKeys, fundedKeySecp256k1)
```
michaelkaplan13, line 130: Does `network.PreFundedKeys` have a default key already set? If so, we can likely get rid of the `fundedKeyStr` set up here all together, and if not, I'd recommend making a new slice here so that it's clear it's expected to start empty.

cam-schultz (commented):
Overall this looks great! Excited about how much faster we'll be able to iterate on our tests now. Just a handful of relatively minor comments.

tests/flows/teleporter_registry.go
feuGeneA: I did a tiny refactor on the `utils` one to make it more self-explanatory, in https://github.com/ava-labs/teleporter/pull/401/commits/eb248d9dd51bf6deebba45698e6714a045e9b56b. let me know what you think

tests/local/network.go
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/401/commits/6ede518c697c3315170b18dc3f45e66bea3787d8

tests/local/network.go
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/401/commits/32f349e1a4eb65f5c0491f16bd69bc2fbaac34e9

tests/local/network.go
feuGeneA: `network.PreFundedKeys` seems to have a bunch of keys, but I tried to use them instead of the `fundedKeyStr` one and it didn't seem to have enough balance to run through all the tests.

I realized that there's nothing to be gained by appending the `fundedKeyStr` one to the `network.PrefundedKeys`, so I cut that out in https://github.com/ava-labs/teleporter/pull/401/commits/f5a74112e90ec7fe43ff663511d9f5cb8369e339.

It's possible we might be able to use the `PreFundedKeys` to run the tests, but for now I just left it working with `fundedKeyStr` to opt for less work.

tests/local/network.go
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/401/commits/1e4e260794b27ac10ebb0e0d69aca70696290d12

tests/local/network_utils.go
```diff
@@ -3,26 +3,37 @@ package local
...
 
-	chainID := subnetDetails.BlockchainID
-	uri := utils.HttpToWebsocketURI(subnetDetails.ValidatorURIs[0], chainID.String())
+	chainID := subnetDetails.Chains[slices.IndexFunc(
+		subnetDetails.Chains,
+		func(c *tmpnet.Chain) bool {
+			// pick the chain that's NOT the C-Chain:
+			return c.ChainID.String() != "11111111111111111111111111111111LpoYY"
+		},
+	)].ChainID
```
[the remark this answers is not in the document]
feuGeneA: I can't remember why exactly I wrote this code, but I confirmed that there's only ever one chain here (at least in this environment), so I simplified things in https://github.com/ava-labs/teleporter/pull/401/commits/9b690bddf11f2143a4043594f5e2c8817d670744.

tests/local/network.go
```diff
@@ -69,17 +69,11 @@ func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
...
-			subnetANodeNames = append(subnetANodeNames, n)
-		} else {
-			subnetBNodeNames = append(subnetBNodeNames, n)
-		}
-	}
+	// Declare 10 new validators (which should have BLS key registered)
+	subnetANodes := subnetEvmTestUtils.NewTmpnetNodes(5)
+	subnetBNodes := subnetEvmTestUtils.NewTmpnetNodes(5)
+	// and some extra nodes to be used to add more validators later:
+	extraNodes := subnetEvmTestUtils.NewTmpnetNodes(5)
```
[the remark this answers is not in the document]
feuGeneA: thanks for the suggestion, I like that. addressed in https://github.com/ava-labs/teleporter/pull/401/commits/850c0d574cd4b29dbac31c1e287964bb0df22ece

tests/local/network.go
```diff
@@ -91,81 +85,66 @@ func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
...
+		warpGenesisFile,
+		subnetEvmTestUtils.DefaultChainConfig,
+		subnetBNodes...,
 	)
-	Expect(err).Should(BeNil())
 
-	// Issue transactions to activate the proposerVM fork on the chains
-	globalFundedKey, err := crypto.HexToECDSA(fundedKeyStr)
+	network := subnetEvmTestUtils.NewTmpnetNetwork(
+		"teleporter-test-local-network",
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/401/commits/bf37499482d8b603fb4ec8ab431401d3f03d3ff5

also parameterized the number of `extraNodes` in https://github.com/ava-labs/teleporter/pull/401/commits/eab5471972e6d8797159e43f8fc2d68902d9c3b2

tests/local/network.go
```diff
@@ -91,81 +85,66 @@ func NewLocalNetwork(warpGenesisFile string) *LocalNetwork {
...
 	Expect(err).Should(BeNil())
 
-	// On initial startup, we need to first set the subnet node names
-	// before calling setSubnetValues for the two subnets
-	subnetIDs := manager.GetSubnets()
-	Expect(len(subnetIDs)).Should(Equal(2))
-	subnetAID := subnetIDs[0]
-	subnetBID := subnetIDs[1]
+	subnetAID := network.GetSubnet("A").SubnetID
+	subnetBID := network.GetSubnet("B").SubnetID
```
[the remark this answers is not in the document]
feuGeneA: addressed within https://github.com/ava-labs/teleporter/pull/401#discussion_r1669212049

.github/workflows/test.yml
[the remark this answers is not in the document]
feuGeneA: maybe? this environment is currently spinning up 15 nodes, so it'd be tailing 15 different log files, which would be a lot of output. I guess it's no more than before, but Maru told me that the tmpnet philosophy is to dump log files to disk rather than flood stdout.

as an alternative, I used ginkgo's `AddReportEntry` function to inform the user of the directory where the the node logs can be found, in https://github.com/ava-labs/teleporter/pull/401/commits/956420b13cc5c5ca3768f97d082d37f6004082ff. what do you think of that?

geoff-vball (approved):
👍

tests/local/network.go
```diff
@@ -220,22 +199,20 @@ func (n *LocalNetwork) setPrimaryNetworkValues() {
 }
 
 func (n *LocalNetwork) setSubnetValues(subnetID ids.ID) {
-	subnetDetails, ok := n.manager.GetSubnet(subnetID)
-	Expect(ok).Should(BeTrue())
-	blockchainID := subnetDetails.BlockchainID
+	subnet := n.tmpnet.Subnets[slices.IndexFunc(
```
michaelkaplan13, line 202: It looks like in all cases `setSubnetValues` is called, we have a `*tmpnet.Subnet` and we pass the `.SubnetID` of it, only to use that ID to look up the `*tmpnet.Subnet` at the start here. Lets just pass the `*tmpnet.Subnet` directly.

tests/utils/utils.go
```diff
@@ -1175,3 +1179,40 @@ func GetChainConfigWithOffChainMessages(offChainMessages []avalancheWarp.Unsigne
...
+	templateFile, err := os.Open(templateFileName)
+	Expect(err).Should(BeNil())
+
+	subnetGenesisFile, err := os.CreateTemp(os.TempDir(), "")
+	Expect(err).Should(BeNil())
+
+	scanner := bufio.NewScanner(templateFile)
+	for scanner.Scan() {
+		var replaced string
+		replaced = scanner.Text()
```
michaelkaplan13, line 1209:
```suggestion
		replaced := scanner.Text()
```

tests/local/network.go
feuGeneA: good call, thank you. addressed in https://github.com/ava-labs/teleporter/pull/401/commits/6f790165bb4257b5fa4eed916e3366f4f87c98e3

tests/utils/utils.go
```diff
@@ -1175,3 +1179,39 @@ func GetChainConfigWithOffChainMessages(offChainMessages []avalancheWarp.Unsigne
...
+		Target string
+		Value  string
+	}{
+		{
+			"<EVM_CHAIN_ID>",
+			strconv.FormatUint(chainID, 10),
+		},
+	}
+
+	templateFile, err := os.Open(templateFileName)
```
michaelkaplan13, line 1200: do we need to `defer templateFile.close()`? (and same for the `subnetGenesisFile`)

michaelkaplan13: Alternatively, since we're reading the whole file either way, I think we can just use `os.ReadFile` here rather than scanning it line by line below.

feuGeneA: good catch, thank you. in https://github.com/ava-labs/teleporter/pull/401/commits/bc9b1b6963eefde530e41c186684330cadaf51d6 I changed to `ReadFile` and also did the `defer...Close` for the `subnetGenesisFile`.

geoff-vball (approved):

michaelkaplan13 (dismissed):

iansuvak (approved):

go.mod
iansuvak: This is probably what's blocking the e2e tests running

iansuvak: nvm. I was lacking context on the issue being around conflicting dependencies on avalanchego.

geoff-vball (approved):
🚀

tests/utils/utils.go
```diff
@@ -1175,3 +1178,37 @@ func GetChainConfigWithOffChainMessages(offChainMessages []avalancheWarp.Unsigne
...
+	templateFileBytes, err := os.ReadFile(templateFileName)
+	Expect(err).Should(BeNil())
+
+	subnetGenesisFile, err := os.CreateTemp(os.TempDir(), "")
+	defer subnetGenesisFile.Close()
+	Expect(err).Should(BeNil())
+
+	var replaced string = string(templateFileBytes[:])
+	for _, s := range substitutions {
+		replaced = strings.Replace(replaced, s.Target, s.Value, 1)
```
cam-schultz, line 1208: It seems a little odd to generalize the target text replacement in this loop, but not the number of replacements. Since we're only replacing a single value in the genesis, I think it's appropriate to just explicitly call `strings.Replace(replaces, "<EVM_CHAIN_ID>",  strconv.FormatUint(chainID, 10))` here. Or we can go in the other direction and also generalize the number of replacements by either calling `strings.ReplaceAll` or putting `n` in the `substitutions` struct.

cam-schultz (commented):
One typo and a couple of minor suggestions, which you should feel free to defer. Otherwise LGTM!

tests/local/network.go
```diff
@@ -90,7 +90,7 @@ func NewLocalNetwork(
 	Expect(err).Should(BeNil())
 	warpChainConfigPath := f.Name()
 
-	allNodes := slices.Clone(extraNodes) // to be appended w/ subnet validators
+	allNodes := extraNodes // to be appended w/ subnet validators
```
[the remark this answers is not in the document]
feuGeneA: the linter made me do it! :) it gave me some seemingly bogus type checking error that it couldn't infer the output type of `slices.Clone`, so I just did it a different way. I'm not 100% sure why the linter didn't catch this earlier.

later on in this code, we want to iterate over `allNodes`, which includes not just `extraNodes` but also all of the subnet nodes as well. this declaration is me starting to build up the list off all nodes, starting with the `extraNodes`.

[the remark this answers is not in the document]
iansuvak: Ah yeah, I understand that practically it's not necessary here just a best practice to not have two slice variables pointing to the same underlying arrays. FWIW I think it would have been solved by declaring `var allNodes []*tmpnet.Node` beforehand.

tests/local/e2e_test.go
```diff
@@ -62,6 +79,12 @@ var _ = ginkgo.BeforeSuite(func() {
 
 	LocalNetworkInstance.DeployTeleporterRegistryContracts(teleporterContractAddress, fundedKey)
 	log.Info("Set up ginkgo before suite")
+
+	ginkgo.AddReportEntry(
+		"network directory with has node logs & configs; useful in the case of failures",
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/429/commits/ec99c4d0fd133f65072254771fad61a10897fbd6

tests/local/network.go
```diff
@@ -437,8 +430,8 @@ func (n *LocalNetwork) setAllSubnetValues() {
 	subnetIDs := n.GetSubnetsInfo()
 	Expect(len(subnetIDs)).Should(Equal(2))
 
-	n.setSubnetValues(n.tmpnet.GetSubnet("A").SubnetID)
-	n.setSubnetValues(n.tmpnet.GetSubnet("B").SubnetID)
+	n.setSubnetValues(n.tmpnet.GetSubnet("A"))
+	n.setSubnetValues(n.tmpnet.GetSubnet("B"))
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/ava-labs/teleporter/pull/429/commits/05b7a9b55364a8f9deea93fff1547568340c844b

Issue #400, opened 2024-06-13
Add additional debugging tools to teleporter-cli


cam-schultz:
**Context and scope**
The following debugging tools have proven helpful in investigating Teleporter and Warp issues. We should add them to teleporter-cli. We should do this before https://github.com/ava-labs/teleporter/issues/377

- debug_traceTransaction calls
- Decode raw transaction bytes
- Decode Warp AddressedCall

These operations can be stacked with the existing decode TeleporterMessage function so that they emit the output of all functions in the call stack. For example, calling `debug_traceTransaction` would call `decodeRawTx`, `decodeAddressedCall`, and `decodeTeleporterMessage`, emitting the output of each.

PR #403, opened 2024-06-17
Add additional info and formatting to transaction command

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/400

## How this works
Formats outputs using `json.MarshalIndent` and prints them using `cmd.Println`, since `logger.Info` can only output to a single line.

Adds an optional `-d` or `--debug` flag to the `transaction` command in the Teleporter CLI. This will output the transaction info and the transaction trace.

## How this was tested
Manually using `go build && ./teleporter-cli transaction --rpc 'https://testnet-qr0606t3tp-xe0ed.avax-test.network/ext/bc/rtLueJ1R8vQ7yGe3QdveWJB1VAnvusy6PPp7TrA5suhUmgbnB/rpc?token=<REDACTED>' -d --teleporter-address "0x253b2784c75e510dD0fF1da844684a1aC0aa5fcf" "0x23a6ab766838c3e0536e6630b8b07297716aaa2773869a80b62b2285b378007e"`

Output: 
[cli-output.txt](https://github.com/user-attachments/files/15915676/cli-output.txt) (updated)

## How is this documented
Adding info for command help


cmd/teleporter-cli/transaction.go
```diff
@@ -30,46 +33,105 @@ var transactionCmd = &cobra.Command{
...
 	},
 }
 
+func checkReceipt(cmd *cobra.Command, txHash common.Hash) {
+	receipt, err := client.TransactionReceipt(context.Background(), txHash)
+	cobra.CheckErr(err)
+
+	for _, log := range receipt.Logs {
+		printTeleporterLogs(cmd, log)
+		printWarpLogs(cmd, log)
```
iansuvak, line 56: Did `printWarpLogs(cmd,log)` portion get tested? It looks like the cli-output.txt includes logs from the `printTeleporterLogs` portion only.

cmd/teleporter-cli/transaction.go
```diff
@@ -30,46 +33,105 @@ var transactionCmd = &cobra.Command{
...
+
+	cmd.Println("Teleporter Log:\n" + string(logJson) + "\n")
+
+	event, err := teleporterABI.EventByID(log.Topics[0])
+	cobra.CheckErr(err)
+
+	out, err := teleportermessenger.FilterTeleporterEvents(log.Topics, log.Data, event.Name)
+	cobra.CheckErr(err)
+
+	outJson, err := json.MarshalIndent(out, "", "  ")
```
iansuvak, line 76: This formats the `bytes` fields of `SourceBlockchainID` and `MessageID` as an array of ints. If we foresee human readability of these fields being valuable in the future should we do an additional post-processing to get blockchainID in cb58 form?

iansuvak (commented):
Generally looks good to me. The questions above are minor.

cmd/teleporter-cli/transaction.go
```diff
@@ -30,46 +33,105 @@ var transactionCmd = &cobra.Command{
...
+
+	cmd.Println("Teleporter Message:")
+	cmd.Println(string(teleporterMessageJson))
+}
+
+func traceTransaction(cmd *cobra.Command, txHash common.Hash) {
+	var result interface{}
+	ct := "callTracer"
+	err := client.Client().Call(&result, "debug_traceTransaction", txHash.String(), tracers.TraceConfig{Tracer: &ct})
+	cobra.CheckErr(err)
```
cam-schultz, line 119: Rather than exiting with an error code here, it might be better to log that an error occurred but not exit. That way if the user does not have access to a debug-enabled API node, they can still use `printTransaction` when calling the CLI with the `--debug` option.

cmd/teleporter-cli/transaction.go
```diff
@@ -30,46 +33,105 @@ var transactionCmd = &cobra.Command{
...
+	outJson, err := json.MarshalIndent(out, "", "  ")
+	cobra.CheckErr(err)
+
+	cmd.Println(event.Name + " Log:")
+	cmd.Println(string(outJson) + "\n")
+}
+
+func printWarpLogs(cmd *cobra.Command, log *types.Log) {
+	if log.Address != common.HexToAddress(warpPrecompileAddress) {
+		return
```
cam-schultz, line 85: An error message here would be helpful.

cmd/teleporter-cli/transaction.go
```diff
@@ -30,46 +33,105 @@ var transactionCmd = &cobra.Command{
...
-				cobra.CheckErr(err)
-
-				teleporterMessage, err := teleportermessenger.UnpackTeleporterMessage(warpPayload.Payload)
-				cobra.CheckErr(err)
-				logger.Info("Parsed Teleporter message",
-					zap.String("warpMessageID", unsignedMsg.ID().Hex()),
-					zap.String("teleporterMessageNonce", teleporterMessage.MessageNonce.String()),
-					zap.Any("message", teleporterMessage))
-			}
+		txHash := common.HexToHash(args[0])
```
cam-schultz, line 40: Just for my own understanding, is cobra smart enough to place the tx hash positional argument at the head of `args` after parsing the keyword arguments? For example is `teleporter-cli --debug 0x1234abcd` a valid invocation?

cmd/teleporter-cli/transaction.go
geoff-vball: I restructured this a bit to make more sense.

cmd/teleporter-cli/transaction.go
geoff-vball: Just tested this, and it looks like cobra is smart enough to take any ordering. `teleporter-cli --debug 0x1234abcd` and `teleporter-cli 0x1234abcd --debug` are both valid

cmd/teleporter-cli/transaction.go
geoff-vball: I added new readable event structs that deserialize into the proper readable formats

iansuvak: Looks great now!

iansuvak (approved):

cmd/teleporter-cli/transaction.go
cam-schultz: Rather than adding standalone types for the sole purpose of printing, we can instead extend the ABI-generated types by implementing [Stringer](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) on the types of interest. We'd need to update `FilterTeleporterEvents` to return `Stringer` instead of `interface{}`, but then we can pass the returned value directly to packages that interact with `fmt` (such as `cobra`) and log using `zap` by using `zap.String`. 

Implementing `Stringer` on these types would also make them immediately usable in other logging contexts, rather than having to manually convert to the `Readable*` version, like we do in `FilterTeleporterEvents`

cmd/teleporter-cli/transaction.go
```diff
@@ -30,46 +33,105 @@ var transactionCmd = &cobra.Command{
 	Short: "Parses relevant Teleporter logs from a transaction",
 	Long: `Given a transaction this command looks through the transaction's receipt
 for Teleporter and Warp log events. When corresponding log events are found,
-the command parses to log event fields to a more human readable format.`,
+the command parses to log event fields to a more human readable format. Optionally pass -d 
+or --debug for extra transaction output. This may require debug enpoints enable on your RPC node`,
```
bernard-avalabs, line 37:
```suggestion
or --debug for extra transaction output. This may require enabling debug enpoints on your RPC node`,
```

abi-bindings/go/Teleporter/TeleporterMessenger/event.go
```diff
@@ -86,19 +89,19 @@ func FilterTeleporterEvents(topics []common.Hash, data []byte, event string) (in
...
-		out = new(TeleporterMessengerAddFeeAmount)
+		out = new(ReadableTeleporterMessengerAddFeeAmount)
 	case MessageExecutionFailed:
-		out = new(TeleporterMessengerMessageExecutionFailed)
+		out = new(ReadableTeleporterMessengerMessageExecutionFailed)
 	case MessageExecuted:
-		out = new(TeleporterMessengerMessageExecuted)
+		out = new(ReadableTeleporterMessengerMessageExecuted)
 	case RelayerRewardsRedeemed:
 		out = new(TeleporterMessengerRelayerRewardsRedeemed)
```
bernard-avalabs, line 102: Any reason why RelayerRewardsRedeemed is different than the other cases?

bernard-avalabs (commented):
Generally makes sense to me. I left a few minor comments.

abi-bindings/go/Teleporter/TeleporterMessenger/event.go
geoff-vball: It didn't have any fields that weren't printed properly.

cmd/teleporter-cli/transaction.go
geoff-vball: I've implemented the `Stringer` interface here. It gives the same output as the previous implementation.

cam-schultz (approved):

minghinmatthewlam (approved):

PR #404, opened 2024-06-18
ValidatorSetSig Contract

### LINKED ISSUE #398: "Validator Set Signature" Contract Utility
**Context and scope**
[Off-chain Warp messages](https://github.com/ava-labs/subnet-evm/issues/729) provide a mechanism for validator to sign arbitrary messages using their BLS keys in a way that enables Warp-enabled VMs to verify that a threshold of stake weight of a given validator set has signed the message. This could be leveraged as an authentication mechanism for smart contracts.

Similarly to how [Safe multi-sig contracts ](https://github.com/safe-global/safe-smart-account/tree/main/contracts) require signatures to be provided from K of N participates in order to execute the input calldata, a `ValidatorSetSig` contract could require a threshold of a given Subnet's validator set to provide a signature in order to execute input calldata.

The would unlock the notion of "Subnet-owned" smart contracts. Instances of these could be used as the admin of other contracts such that admin actions could only be taken a threshold of stake of a given Subnet signs off on the action. These contracts could be deployed on any Warp-enabled chain, no matter which Subnet validator set is required to provide its signature. For instance, you could have a contract on the C-Chain that requires a Warp signature from a specific Subnet, or you could have a contract on a Subnet that requires a Warp signature from its own validators (a form of off-chain governance).

**Open questions**
- How to handle replay protection of messages?
   - The Safe multi-sig contracts use basic nonce tracking.
- How to ensure that messages only have effect on the target contracts on the target chains (i.e cross-chain replay protection).

iansuvak:
## Why this should be merged
closes #398 
## How this works
Adds a new contract that can be instantiated with a `blockChainID` whose validator quorum needs to sign a message for it to be passed on to a target contract. This contract is intended to be used as an owner of other contracts requiring off-chain governance for updates. 
## How this was tested

- [x] New solidity unit tests
- [x] New e2e test 

## How is this documented

- [x] Add README.md
- [x] Improve comments 

## Open questions

1. Currently the contract reverts if the call to the target contract fails. The idea is to prevent consuming the nonce if the call fails for a fixable reason such as OOG. Do we want this behavior?
2. This repo is an odd place for the contract since it doesn't directly use Teleporter at all and I'm not sure that I like the `OffChainMessageContracts` sub-directory naming.


contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,113 @@
...
+
+pragma solidity 0.8.18;
+
+import {
+    WarpMessage,
+    IWarpMessenger
+} from "@avalabs/subnet-evm-contracts@1.2.0/contracts/interfaces/IWarpMessenger.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts@4.8.1/security/ReentrancyGuard.sol";
+
+struct ValidatorSetSigMessage {
```
iansuvak, line 14: As I am writing e2e tests I've realized that `abi-gen` doesn't generate go structs for this unless it is used as an argument to an external or public function.

I can either remove this convenience wrapper and decode directly as a tuple or provide a new `validateMessage` `public view` function that checks relevant arguments including nonce. The benefit of this would be that the people preparing messages can confirm that they pass validation before passing them onto validators to get signed and sent over as a `WarpMessage`.

cam-schultz: I think you just need to add `ValidatorSetSig` to the list in `abi_gen.sh` https://github.com/ava-labs/teleporter/blob/main/scripts/abi_bindings.sh#L19

cam-schultz: Is there a strong reason why this type needs to be in the ABI generated files at all? It would be helpful for constructing the Warp message in e2e tests, but I'm not convinced that that's reason enough to add the type to the ABI. 

Adding a `validateMessage` method would work as you described, but I don't know how useful it would be in the contract itself, since it would be likely used out-of-band of the actual message provided to `executeMessage`. In which case, manually verifying by other means (i.e. inspection of the `require` statements in `executeMessage) would yield the same result.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,113 @@
...
+    address validatorSetSigAddress;
+    address targetContractAddress;
+    bytes32 targetBlockChainID;
+    uint256 nonce;
+    bytes txPayload;
+}
+
+/**
+ * @dev Contract that verifies that a set threshold of validators from a given blockchainID
```
cam-schultz, line 23: Let's add a comment that the threshold is set by the validators themselves in their Warp configs: https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/config.go#L50

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,113 @@
...
+ */
+contract ValidatorSetSig is ReentrancyGuard {
+    /**
+     * @notice A blockchain ID whose validators need to sign the message for it to be considered valid.
+     */
+    bytes32 public immutable validatorBlockchainID;
+
+    /**
+     * @dev Tracks nonces for messages sent to a specific contract address to provide replay protection.
+     * The target contract address is used as the key in order to prevent instead of keeping a global nonce in order to prevent
```
cam-schultz, line 37:
```suggestion
     * The target contract address is used as the key instead of keeping a global nonce in order to prevent
```

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,113 @@
...
+     * race conditions where the same instance of ValidatorSetSig contract is used to manage multiple downstream contracts.
```
cam-schultz, line 38: I don't think "race condition" is accurate here, since the EVM is a single threaded state machine. IIUC a global nonce would be perfectly safe in terms of replay protection, but would be poor UI since other independent contracts could consume the nonce that the target contract is intending to use.

cam-schultz (commented):
Generally makes sense to me. I think the parameterization of `ValidatorSetSig` makes a lot of sense, especially the decoupling of the signingand validating blockchains.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
iansuvak: I agree. I wasn't thinking of a software "race condition" here but of human race condition where two parties want to sign different messages at once and one of them ends up going through the process with an already consumed nonce.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,113 @@
...
+     */
+    mapping(address targetContractAddress => uint256 nonce) public nonces;
+
+    /**
+     * @notice Address that the off-chain Warp message sets as the "source" address.
+     * @dev The address is not owned by any EOA or smart contract account, so it
+     * cannot possibly be the source address of any other Warp message emitted by the VM.
+     */
+    address public constant VALIDATORS_SOURCE_ADDRESS = address(0);
```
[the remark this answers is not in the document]
iansuvak: Yeah. The idea here is to limit it to off-chain messages only. If the desired usecase is to enable on-chain governance of a contract, that can already be done with a multiSig

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
iansuvak: I just updated the comment, hopefully it's clearer now

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
iansuvak: I don't feel strongly about it. For e2e testing I can either manually define the struct in my packing.go file or I can modify the contract to not use the wrapping struct and just encode/decode arguments directly. I only included it in the first place because i have 5 arguments and wanted their uses and naming to be clear up front. 

The `validateMessage` method was a post-hoc justification and I don't feel strongly about it either. I agree that visually inspecting the code and `require`  statements would do the same thing. The utility behind this would be a "fool-proof" way to confirm that the message is valid before handing it off to validators for signing.  A practical downside here is that having another public function call where we pass the struct to the `validateMessage` is likely to consume more gas for copying the struct. I don't see this as a big problem given rarity of invocation of this contract.

All that being said my current change does include it but let me know if we want to remove it and I will go back and update the e2e tests. The options as I see them are:

:one: Leave as is with `validateMessage` method available and used by `executeCall`
:two: Remove the `validateMessage` but keep using the `ValidatorSetSigMessage` struct in the contract and manually define the matching go struct in`packing.go`
:three: Remove the `ValidatorSetSigMessage` struct from the contract and encode/decode arguments directly.

minghinmatthewlam: I am leaning towards option 1, mostly from a readability and extensibility stand point. It'd make submitting transactions to this contract more clear for what parameters need to be passed in, and also easier to extend to this struct if needed. This would also fit in nicely with our current pattern for writing unit/e2e tests. 

The downside I'd see for this option would be deployment gas costs, but shouldn't too big of an impact, especially considering this contract can be used against different target addresses.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,123 @@
...
+    address validatorSetSigAddress;
+    address targetContractAddress;
+    bytes32 targetBlockChainID;
+    uint256 nonce;
+    bytes txPayload;
+}
+
+/**
+ * @dev Contract that verifies that a set threshold of validators from a given blockchainID
+ * have signed an off-chain Warp message and forwards the payload to the destination contract specified in the payload.
```
minghinmatthewlam, line 25: nit: here says destination, vs struct was target

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,123 @@
...
+ * The threshold itself is set by the validator themselves in their Warp configs:
+ * https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/config.go#L50
+ *
+ * This is intended to be used for safe off-chain governance of enabled contracts by having the target contract be owned by
```
minghinmatthewlam, line 29: This should apply to any general access control mechanism on chain right? For example access through roles. 

Would clarify documentation to be more general access control, and use Ownable as an example.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,124 @@
...
+        // solhint-disable-next-line avoid-low-level-calls
+         validatorSetSigMessage.targetContractAddress.call(validatorSetSigMessage.txPayload);
+
+        // Use require to revert the transaction if the call fails. This is to prevent consuming the nonce if the call fails due to OOG
+        // and requiring re-signing of the message with a new nonce.
+        require(success, "ValidatorSetSig: call failed");
+        emit Delivered(validatorSetSigMessage.targetContractAddress, validatorSetSigMessage.nonce);
+    }
+
+    function validateMessage(ValidatorSetSigMessage memory message) public view returns (bool) {
```
michaelkaplan13, line 109: If `validateMessage` returns a bool, I think the implementation should be along the lines of:
```
return 
    message.validatorSetSigAddress == address(this) && 
    message.targetBlockchain = blockchainID &&
    ...;
```

And then the handling of invalid messages (i.e. reverting) is left to the caller.

Alternatively, we could have the individual reverts here and not have any return value.

iansuvak: Do you have a preference for either way? I agree that the current implementation is not very idiomatic. The main downside of switching to above implementation is that no specific reason for the revert is given to the caller. It's not a huge problem here since there are only a few fields so issues can be visually inspected.  

Overall slightly leaning towards keeping the reverts and removing the return value but don't feel strongly either way

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,130 @@
...
+    IWarpMessenger public constant WARP_MESSENGER =
+        IWarpMessenger(0x0200000000000000000000000000000000000005);
+
+    /**
+     * @notice Emited when the payload is successfully delivered to the target contract.
+     */
+    event Delivered(address indexed targetContractAddress, uint256 indexed nonce);
+
+    constructor(bytes32 validatorBlockChainID) {
+        validatorBlockchainID = validatorBlockChainID;
```
bernard-avalabs, line 79: nit: I'm not sure what our coding style guide says about this, but having these two variable names differ only by their capitalization for "chain" seems a bit error prone.

bernard-avalabs (commented):
Generally makes sense to me. Just had a minor comment about naming in the constructor.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
minghinmatthewlam: we'd make the parameter `validatorBlockchainID_`

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,130 @@
...
+pragma solidity 0.8.18;
+
+import {
+    WarpMessage,
+    IWarpMessenger
+} from "@avalabs/subnet-evm-contracts@1.2.0/contracts/interfaces/IWarpMessenger.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts@4.8.1/security/ReentrancyGuard.sol";
+
+/**
+ * @dev Message format for the WarpMessage payload to be forwarded to the target contract
```
minghinmatthewlam, line 15:
```suggestion
 * @dev Message format for the Warp message payload to be forwarded to the target contract
```

nit: dont' think we need to refer to the type here

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * validatorSetSigAddress: Address of the ValidatorSetSig contract this message is intended for
```
minghinmatthewlam, line 17: we try to use the [NatSpec](https://docs.soliditylang.org/en/latest/natspec-format.html) format whenever possible, so these struct variables we use `@param`

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,130 @@
...
+ * @dev Contract that verifies that a set threshold of validators from a given blockchainID
+ * have signed an off-chain Warp message and forwards the payload to the target contract specified in the payload.
+ * The threshold itself is set by the validator themselves in their Warp configs:
+ * https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/config.go#L50
+ *
+ * This is intended to be used for safe off-chain governance of enabled contracts. An example usecase would be
+ * to deploy a `Ownable` target contract that is owned by an instance of this contract and adding the
+ * `onlyOwner` modifier to the functions that should be governed.
+ */
+contract ValidatorSetSig is ReentrancyGuard {
```
minghinmatthewlam, line 41: can you add a `@custom:security-contact`

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
michaelkaplan13: > Overall slightly leaning towards keeping the reverts and removing the return value but don't feel strongly either way

Yeah that makes sense to me

contracts/src/OffChainMessageContracts/README.md
michaelkaplan13, line 1: I know you brought this up previously, but what are people thoughts on having the subdirectory breakdown be:
```
src/
    TeleporterMessenger/
    Utilities/
        Upgrades/
        ValidatorSetSignature/
```

I think this structure sets it up nicely for future "Teleporter" protocols other than point-to-point messaging  (i.e. event importing), and also makes it clear where new utility contracts can be added in the future (whether or not they use off-chain messages). The one odd aspect to it I can see is that `Upgrades` is really only meant for the `TeleporterMessenger` protocol to a certain extent. An alternative could be to put `TeleporterUpgradeable` and `TeleporterOwnerUpgradeable` under `TeleporterMessenger`, and move `TeleporterRegistry` under `Utilities` since it is its own standalone contract.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
iansuvak: Yeah I looked at NatSpec but looks like the `@param` is only supported for functions, events and errors (correctly since these aren't params really)  hence why I went with this for now. It looks like this issue is open though: https://github.com/ethereum/solidity/issues/12295 

Do we still want to use `@param` even though it's a misnomer?

contracts/src/OffChainMessageContracts/README.md
```diff
@@ -0,0 +1,15 @@
+## Off Chain Message Contracts
+
+This directory contains Solidity contracts that use off chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview). These are Warp messages whose source contract address is the zero address that is not owned and therefore could not have been originated on-chain.
```
michaelkaplan13, line 3:
```suggestion
This directory contains Solidity contracts that use off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview). These are Warp messages that are signed via manual approve by a chain's validators, rather than being submitted on-chain to be sent.
```

Nit: Technically off-chain messages don't _need_ to have source address of zero, that's just one way a contract receiving them can be sure that they couldn't have been published on-chain.

contracts/src/OffChainMessageContracts/ValidatorSetSig.sol
```diff
@@ -0,0 +1,130 @@
...
+// SPDX-License-Identifier: Ecosystem
+
+pragma solidity 0.8.18;
+
+import {
+    WarpMessage,
+    IWarpMessenger
+} from "@avalabs/subnet-evm-contracts@1.2.0/contracts/interfaces/IWarpMessenger.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts@4.8.1/security/ReentrancyGuard.sol";
+
```
michaelkaplan13, line 13: Could you add the "unaudited" disclaimer to this contract for the time being?

```
/**
 * THIS IS AN EXAMPLE CONTRACT THAT USES UN-AUDITED CODE.
 * DO NOT USE THIS CODE IN PRODUCTION.
 */
```

contracts/src/OffChainMessageContracts/README.md
iansuvak: I support moving `TeleporterRegistry` outside of the `TeleporterMessenger` since it's a standalone contract as you said.  However `TeleporterRegistry` is also more closely tied to the `TeleporterMessenger` than `ValidatorSetSignature` which isn't at all. 

Maybe the following makes sense? That would put Teleporter related applications at the top level while letting us keep loosely related contracts in the `Utilities` folder
```
src/
      TeleporterMessenger/
            Upgrades/
      TeleporterRegistry/
      Utilities/
          /ValidatorSetSignature
```

abi-bindings/go/Utilities/ValidatorSetSig/packing_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPackUnpackValidatorSetSigMessage(t *testing.T) {
+	msg := ValidatorSetSigMessage{
```
cam-schultz, line 16: Not for this PR, but in all our `packing_test.go`'s it would be great if we could perform validation on the struct in the test itself. For instance, if we add a new field to the Solidity struct under test, this Go test would still pass. https://github.com/GaijinEntertainment/go-exhaustruct seems like a good candidate, and it's included in golang-ci, which we already use. I'm not sure if you can target specific linters at specific files (but not others) with golang-ci, but a simple alternative would be running exhaustruct directly.

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,13 @@
+## Utility Contracts
+
+### ValidatorSetSig Contract
+
+This contract provides an alternative to MultiSig contracts for upgradeability where instead of requiring k of n signatures on an upgrade transaction we can require a quorum of validators from a given blockchain.
+
+The contract accepts off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview) signed via a manual approve by a chain's validators. It requires these messages to have source address set to the zero address to enforce this.
```
cam-schultz, line 7:
```suggestion
The contract accepts off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview), which are manually approved for signing  by a chain's validators. It requires these messages to have the source address set to the zero address to enforce this. 
```

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+### ValidatorSetSig Contract
+
+This contract provides an alternative to MultiSig contracts for upgradeability where instead of requiring k of n signatures on an upgrade transaction we can require a quorum of validators from a given blockchain.
+
+The contract accepts off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview) signed via a manual approve by a chain's validators. It requires these messages to have source address set to the zero address to enforce this. 
+
+Note:
+1. The blockchain validating the message may or may not be the same chain where the target contract and the ValidatorSetSig contract are deployed.
+2. The quorum of signers required for a Warp Message to be considered valid is configured by the validators themselves via their [Warp configs](https://github.com/ava-labs/subnet-evm/issues/729)
```
cam-schultz, line 11: This links to the issue describing off-chain Warp messages, which is distinct from a chain's stake weight threshold requirement for signature verification. I don't think we need to mention signature verification here, as it's covered extensively elsewhere, but keeping the link and rewording the bullet point to describe off-chain Warp messages in detail would be good to include.

contracts/src/Utilities/README.md
cam-schultz, line 1: Can you include a description of how the off-chain Warp message is expected to be formatted? I believe the steps are:
1.  ABI-encode a `ValidatorSetSigMessage`
2. Pack that as the payload of an `[AddressedCall](https://github.com/ava-labs/avalanchego/blob/0c4efd743e1d737f4e8970d0e0ebf229ea44406c/vms/platformvm/warp/payload/addressed_call.go#L15)` Warp Message format
3. Pack the `AddressedCall` as the payload  of the `[UnsignedWarpMessage](https://github.com/ava-labs/avalanchego/blob/f17ea6a7ab4036c41b693e47b94d8f0c81cb69ec/vms/platformvm/warp/unsigned_message.go#L14)`

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -11,20 +11,39 @@ import {
...
+    bytes32 targetBlockchainID;
     uint256 nonce;
-    bytes txPayload;
+    bytes payload;
 }
 
 /**
  * @dev Contract that verifies that a set threshold of validators from a given blockchainID
- * have signed an off-chain Warp message and forwards the payload to the destination contract specified in the payload.
+ * have signed an off-chain Warp message and forwards the payload to the target contract specified in the payload.
```
cam-schultz, line 38:
```suggestion
 * have signed an off-chain Warp message and forwards the payload to the target contract specified in the ValidatorSetSigMessage.
```
Reword to avoid confusion over the various payloads involved.

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -11,20 +11,39 @@ import {
...
+ * The threshold itself is set by the validator themselves in their Warp configs:
+ * https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/config.go#L50
```
cam-schultz, line 40:
```suggestion
 * https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/config.go#L50
```
It'd be better to link to a fixed commit/tag, so that it doesn't break as `master` changes.

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -83,31 +103,35 @@ contract ValidatorSetSig is ReentrancyGuard {
...
-            "ValidatorSetSig: invalid nonce"
-        );
+        validateMessage(validatorSetSigMessage);
 
         nonces[validatorSetSigMessage.targetContractAddress] = validatorSetSigMessage.nonce;
 
         (bool success,) =
         // solhint-disable-next-line avoid-low-level-calls
-         validatorSetSigMessage.targetContractAddress.call(validatorSetSigMessage.txPayload);
+         validatorSetSigMessage.targetContractAddress.call(validatorSetSigMessage.payload);
```
cam-schultz, line 115: Can you add a comment explaining that since the caller is expected to have full control over the contract called (since it requires a threshold of validators to agree), we don't need to protect against return bomb vectors. For context, we _do_ have to protect against that in `TeleporterMessenger` as well as `TeleporterTokenBridge`: https://github.com/ava-labs/teleporter-token-bridge/blob/afdb3c21691c0ddb389e2a2ae359a43074249ce1/contracts/src/utils/CallUtils.sol#L24

tests/utils/utils.go
```diff
@@ -1035,3 +1147,30 @@ func ParseTeleporterMessage(unsignedMessage avalancheWarp.UnsignedMessage) *tele
 
 	return teleporterMessage
 }
+
+func GetWarpEnabledChainConfig(offChainMessages []avalancheWarp.UnsignedMessage) string {
```
cam-schultz, line 1151: `GetChainConfigWithOffChainMessages` would be more descriptive. Warp is enabled on all chains post-Durango

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/ava-labs/teleporter/tests/interfaces"
+	"github.com/ava-labs/teleporter/tests/utils"
+	. "github.com/onsi/gomega"
+)
+
+func ValidatorSetSig(network interfaces.LocalNetwork) {
+	// Deploy ValidatorSetSig contract to contract subnet to be validated from validator subnet
+	// Deploy a mock ERC20 contract to contract subnet
+	// Construct ValidatorSetSig message with mock ERC20 as the target contract
+	// Create a OffChainWarp Message using the ValidatorSetSig message to be signed by the validator subnet
```
cam-schultz, line 20:
```suggestion
	// Create an OffChainWarp Message using the ValidatorSetSig message to be signed by the validator subnet
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,157 @@
...
+		networkID,
+		validatorSubnet,
+		validatorSetSigContractAddress,
+		[]validatorsetsig.ValidatorSetSigMessage{vssMessage1, vssMessage2})
+
+	// Create chain config with off chain messages
+	chainConfigs := make(map[string]string)
+	utils.SetChainConfig(chainConfigs, cChainInfo, noMessageWarpEnabledChainConfig)
+	utils.SetChainConfig(chainConfigs, validatorSubnet, warpEnabledChainConfigWithMsg)
+	utils.SetChainConfig(chainConfigs, contractSubnet, noMessageWarpEnabledChainConfig)
```
cam-schultz, line 90: Is it necessary to set the chain config for chains that do not sign off-chain Warp messages?

tests/flows/validator_set_sig.go
cam-schultz, line 1: We should also test the case where validatorSubnet == contractSubnet

cam-schultz (commented):
I left a handful of relatively small comments. Overall, I think this looks great!

tests/flows/validator_set_sig.go
iansuvak: No it's not. Good catch =)

abi-bindings/go/Utilities/ValidatorSetSig/packing_test.go
iansuvak: 👍 cut an issue for it #411

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,138 @@
...
+ * THIS IS AN EXAMPLE CONTRACT THAT USES UN-AUDITED CODE.
+ * DO NOT USE THIS CODE IN PRODUCTION.
+ */
+
+/**
+ * @dev Message format for the WarpMessage payload to be forwarded to the target contract
+ *
+ * validatorSetSigAddress: Address of the ValidatorSetSig contract this message is intended for
+ * targetContractAddress: Address of the contract that the payload should be forwarded to
+ * targetBlockchainID: Blockchain ID of the chain the message is intended for
```
michaelkaplan13, line 24: Is the `targetBlockchainID` always the same blockchain as the chain that the `validatorSetSigAddress` is on? If so, it may be slightly more clear to have it as the first parameter of `ValidatorSetSigMessage` so that it goes from most general to most specific

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,138 @@
...
+            "ValidatorSetSig: invalid validatorSetSigAddress"
+        );
+        require(
+            message.targetBlockchainID == blockchainID,
+            "ValidatorSetSig: invalid targetBlockchainID"
+        );
+        require(
+            nonces[message.targetContractAddress] + 1 == message.nonce,
+            "ValidatorSetSig: invalid nonce"
+        );
```
michaelkaplan13, line 136: Just calling out that this is slight deviation from EVM transaction nonces, where the EVM stores the next nonce to be used for a given account (starting from 0), rather than the last nonce use by the account (starting from 1).

This is still correct I think, but may want to document that nonces need to start from 1, or make the small change to have it match transaction nonces.

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,214 @@
...
+	// Deploy a mock ERC20 contract to contract subnet
+	// Construct ValidatorSetSig message with mock ERC20 as the target contract
+	// Create a OffChainWarp Message using the ValidatorSetSig message to be signed by the validator subnet
+	// Confirm the event is emitted
+	// Retry the same message and confirm it fails
+	// Send a new message with incremented nonce and confirm it is accepted.
+	// Deploy a new mock ERC20 contract to the validator subnet
+	// Send a new message with the new ERC20 contract as the target contract and confirm it is accepted
+
+	contractSubnet, validatorSubnet := utils.GetTwoSubnets(network)
```
michaelkaplan13, line 27: Could just be me, but I find "contractSubnet" and "validatorSubnet" to be confusing names here. I think the intention was that "validatorSubnet" was that the `validatorSetSig contract deployed there is owned by that subnet itself.

michaelkaplan13 (commented):
Last set of super minor comments hopefully 🙏

tests/flows/validator_set_sig.go
iansuvak: I agree. Original intention was to have `contractSubnet` have both contracts deployed on it, `validatorSetSig` and the `ExampleERC20` and the `validatorSubnet` is the subnet that signs the warp message to trigger actions. 

Given @cam-schultz 's feedback I added a test case where the signing chain is the same as the chain that contracts are deployed on which meant that I had to either:
:one: deploy an instance of the `validatorSetSig` contract on the `validatorSubnet` as well
:two: change how I was handling the warp messages and have the `contractSubnet` sign off-chain warp messages as well.

I opted for the former but kept the confusing naming. 
I'm changing to `subnetA` and `subnetB` now but happy to pivot to option :two: if that is preferred.

michaelkaplan13: I'm good with just `subnetA` and `subnetB` 👍  the test itself looks correct and provides good coverage as-is IMO

michaelkaplan13 (dismissed):
🚀

contracts/README.md
```diff
@@ -1,7 +1,7 @@
-## Teleporter Messenger Contract and Example Cross Chain Applications
```
minghinmatthewlam, line 1: thanks for cleaning this up

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,18 @@
+## Utility Contracts
```
minghinmatthewlam, line 1: nit: thoughts on changing directoyr name `Utilities` -> `utils` or `utilities`? My pref is `utils` but might look off for the abi-binding generated file

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+### ValidatorSetSig Contract
+
+This contract provides an alternative to traditional multi-signature contracts where instead of requiring signatures from `K` of `N` pre-specified signers, an aggregate signature is required from a quorum of the current validators a given blockchain.
```
minghinmatthewlam, line 5:
```suggestion
This contract provides an alternative to traditional multi-signature contracts where instead of requiring signatures from `K` of `N` pre-specified signers, an aggregate signature is required from a quorum of the current validators for a given blockchain.
```

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The contract leverages off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview), which are manually approved for signing  by a chain's validators. It requires these messages to have the source address set to the zero address to enforce this.
```
minghinmatthewlam, line 7:
```suggestion
The contract leverages off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview), which are manually approved for signing by a chain's validators. It requires these messages to have the source address set to the zero address to enforce this.
```

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Note:
+1. The blockchain validating the message may or may not be the same chain where the target contract and the `ValidatorSetSig` contract are deployed.
+2. [Off-Chain Warp messages](https://github.com/ava-labs/subnet-evm/issues/729) are described in the linked issue. They are Warp messages that validators can include in their config to indicate that they are willing to sign them even though they are not a result of on-chain activity.
+
+#### Creating a valid Off-chain Warp Message for interaction with the ValidatorSetSig Contract
```
minghinmatthewlam, line 13: nit: should we keep capitalization for "Off-chain" same throughout?

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+1. ABI-encode a `ValidatorSetSigMessage` as defined in `ValidatorSetSig.sol`
+2. OPTIONAL: call `validateMessage` view method of the intended contract to confirm that the message has correct fields, including the nonce.
+3. Pack that as the payload of na [AddressedCall](https://github.com/ava-labs/avalanchego/blob/0c4efd743e1d737f4e8970d0e0ebf229ea44406c/vms/platformvm/warp/payload/addressed_call.go#L15) Warp Message format.
+4. Pack the `AddressedCall` as the payload of the [UnsignedWarpMessage](https://github.com/ava-labs/avalanchego/blob/f17ea6a7ab4036c41b693e47b94d8f0c81cb69ec/vms/platformvm/warp/unsigned_message.go#L14).
```
minghinmatthewlam, line 18: should we reference aggregating signatures, and possibly point to awm-relayer here

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * THIS IS AN EXAMPLE CONTRACT THAT USES UN-AUDITED CODE.
+ * DO NOT USE THIS CODE IN PRODUCTION.
+ */
+
+/**
+ * @dev Message format for the WarpMessage payload to be forwarded to the target contract
+ *
+ * targetBlockchainID: Blockchain ID of the chain the message is intended for
```
minghinmatthewlam, line 22: we tend to use `@param` even on struct variables

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+     */
+    event Delivered(address indexed targetContractAddress, uint256 indexed nonce);
+
+    constructor(bytes32 validatorBlockchainID_) {
+        validatorBlockchainID = validatorBlockchainID_;
+        blockchainID = WARP_MESSENGER.getBlockchainID();
+    }
+
+    function executeCall(uint32 messageIndex) external nonReentrant {
+        // Get the WarpMessage from the WarpMessenger pre-compile and verify that it is valid
```
minghinmatthewlam, line 91:
```suggestion
        // Get the WarpMessage from the WarpMessenger precompile and verify that it is valid
```

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+        (WarpMessage memory message, bool valid) =
```
minghinmatthewlam, line 92: nit: can we use `success` instead of `valid`? To match `TeleporterMessenger`

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+        require(valid, "ValidatorSetSig: invalid warp message");
+
+        // Ensure that the sourceChainID of the validator set that signed the message is the authorized one
+        require(
+            message.sourceChainID == validatorBlockchainID, "ValidatorSetSig: invalid sourceChainID"
+        );
+        // Ensure that the originSenderAddress is the zero address to prevent on-chain messages from being forwarded
+        require(
+            message.originSenderAddress == VALIDATORS_SOURCE_ADDRESS,
+            "ValidatorSetSig: non-zero originSenderAddress"
```
minghinmatthewlam, line 103:
```suggestion
            "ValidatorSetSig: non-zero origin sender address"
```
we generally havent' used the variable names in revert messages

contracts/src/Utilities/ValidatorSetSig.sol
minghinmatthewlam: is there downside to using the same assembly execute call?

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+        validateMessage(validatorSetSigMessage);
+
+        nonces[validatorSetSigMessage.targetContractAddress] = validatorSetSigMessage.nonce + 1;
+
+        // We don't need to protect against return bomb vectors below here since the caller is expected to have full control over the contract called.
+        (bool success,) =
+        // solhint-disable-next-line avoid-low-level-calls
+         validatorSetSigMessage.targetContractAddress.call(validatorSetSigMessage.payload);
+
+        // Use require to revert the transaction if the call fails. This is to prevent consuming the nonce if the call fails due to OOG
```
minghinmatthewlam, line 118:
```suggestion
        // Use require to revert the transaction if the call fails. This is to prevent consuming the nonce if the call fails due to out of gas
```

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+        // and requiring re-signing of the message with a new nonce.
+        require(success, "ValidatorSetSig: call failed");
+        emit Delivered(validatorSetSigMessage.targetContractAddress, validatorSetSigMessage.nonce);
```
minghinmatthewlam, line 121: do we want to include payload bytes in the event but not indexed?

contracts/src/Utilities/README.md
```diff
@@ -2,12 +2,17 @@
...
+1. The blockchain validating the message may or may not be the same chain where the target contract and the `ValidatorSetSig` contract are deployed.
+2. [Off-Chain Warp messages](https://github.com/ava-labs/subnet-evm/issues/729) are described in the linked issue. They are Warp messages that validators can include in their config to indicate that they are willing to sign them even though they are not a result of on-chain activity.
 
-A public `validateMessage` view method is provided as a way to verify that the message has correct fields before being passed on to validators for signing.
\ No newline at end of file
+#### Creating a valid Off-chain Warp Message for interaction with the ValidatorSetSig Contract
+
+1. ABI-encode a `ValidatorSetSigMessage` as defined in `ValidatorSetSig.sol`
+2. OPTIONAL: call `validateMessage` view method of the intended contract to confirm that the message has correct fields, including the nonce.
+3. Pack that as the payload of na [AddressedCall](https://github.com/ava-labs/avalanchego/blob/0c4efd743e1d737f4e8970d0e0ebf229ea44406c/vms/platformvm/warp/payload/addressed_call.go#L15) Warp Message format.
```
cam-schultz, line 17:
```suggestion
3. Pack that as the payload of an [AddressedCall](https://github.com/ava-labs/avalanchego/blob/0c4efd743e1d737f4e8970d0e0ebf229ea44406c/vms/platformvm/warp/payload/addressed_call.go#L15) Warp Message format.
```

contracts/src/Utilities/README.md
```diff
@@ -2,12 +2,17 @@
...
+The contract leverages off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview), which are manually approved for signing  by a chain's validators. It requires these messages to have the source address set to the zero address to enforce this.
 
 Note:
-1. The blockchain validating the message may or may not be the same chain where the target contract and the ValidatorSetSig contract are deployed.
-2. The quorum of signers required for a Warp Message to be considered valid is configured by the validators themselves via their [Warp configs](https://github.com/ava-labs/subnet-evm/issues/729)
+1. The blockchain validating the message may or may not be the same chain where the target contract and the `ValidatorSetSig` contract are deployed.
+2. [Off-Chain Warp messages](https://github.com/ava-labs/subnet-evm/issues/729) are described in the linked issue. They are Warp messages that validators can include in their config to indicate that they are willing to sign them even though they are not a result of on-chain activity.
```
cam-schultz, line 11:
```suggestion
2. [Off-Chain Warp messages](https://github.com/ava-labs/subnet-evm/issues/729) are Warp messages that validators can include in their config to indicate that they are willing to sign them even though they are not a result of on-chain activity.
```

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    function validateMessage(ValidatorSetSigMessage memory message) public view {
+        require(
+            message.targetBlockchainID == blockchainID,
```
minghinmatthewlam, line 126: same nits here: generally have stayed away from variable name directly in require statements. Think makes sense since people using tools or submitting the txs that see the errors might not be viewing the code or have code context.

contracts/src/Utilities/tests/ValidatorSetSigTests.t.sol
```diff
@@ -0,0 +1,238 @@
+// (c) 2023, Ava Labs, Inc. All rights reserved.
```
minghinmatthewlam, line 1:
```suggestion
// (c) 2024, Ava Labs, Inc. All rights reserved.
```

contracts/src/Utilities/README.md
cam-schultz: I don't think that's necessary since signature aggregation is separate from the validator signing the message provided via config

contracts/src/Utilities/tests/ValidatorSetSigTests.t.sol
```diff
@@ -0,0 +1,238 @@
...
+        mockERC20.transfer(_ERC20_RECIPIENT_ADDRESS, 10);
+        assertEq(mockERC20.balanceOf(_ERC20_RECIPIENT_ADDRESS), 10);
+    }
+
+    function testInvalidMessage() public {
+        vm.mockCall(
+            WARP_PRECOMPILE_ADDRESS,
+            abi.encodeWithSelector(IWarpMessenger.getVerifiedWarpMessage.selector, uint32(0)),
+            abi.encode(
+                WarpMessage({
```
minghinmatthewlam, line 54: we can use `_getValidWarpMessage` here

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+func ValidatorSetSig(network interfaces.LocalNetwork) {
+	// ************************************************************************************************
+	// Setup
+	// ************************************************************************************************
+	// Deploy ValidatorSetSig expecting signatures from subetB instances to both subnets
+	// Deploy exampleERC20 instance to  both subnets
+	// Construct ValidatorSetSig message with mock ERC20 as the target contract
+	// Create OffChainWarp Messages using the ValidatorSetSig message to be signed by the subnetB
```
minghinmatthewlam, line 23:
```suggestion
	// Create off-chan Warp messages using the ValidatorSetSig message to be signed by the subnetB
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	runner_sdk "github.com/ava-labs/avalanche-network-runner/client"
+	"github.com/ava-labs/subnet-evm/accounts/abi/bind"
+	exampleerc20 "github.com/ava-labs/teleporter/abi-bindings/go/Mocks/ExampleERC20"
+	validatorsetsig "github.com/ava-labs/teleporter/abi-bindings/go/Utilities/ValidatorSetSig"
+	"github.com/ava-labs/teleporter/tests/interfaces"
+	"github.com/ava-labs/teleporter/tests/utils"
+	. "github.com/onsi/gomega"
+)
+
+func ValidatorSetSig(network interfaces.LocalNetwork) {
```
minghinmatthewlam, line 16: doesn't have to be this PR, but we new `utils` folder for contracts, I'm thinking our e2e tests structure should follow that of the contracts. So would be `tests/flows/utilities/validator_set_sig.go` same for `teleporter_registry` tests, etc.

minghinmatthewlam: and teleporter e2e tests would be in `tests/flows/teleporter`

contracts/src/Utilities/README.md
minghinmatthewlam: I'm asking about including signature aggregation because the section `Creating a valid Off-chain Warp Message for interaction with the ValidatorSetSig Contract` is for "valid off-chain Warp message", and semantically we've used "valid Warp message" as one that includes valid aggregate signature

minghinmatthewlam: would suggest removing "valid" from the header, but no strong pref

cam-schultz: Fair point, though I think the word "valid" should still be included. If we want to be technically accurate, what we're doing is providing a "valid unsigned off-chain Warp message". I don't think it's necessary to mention signature aggregation or any of the other details around actual message delivery here. Or if we do, keep it high level and say that signatures can be aggregated and packaged into a transaction in the same manner as any other Warp message.

minghinmatthewlam: i agree high level is good enough

contracts/src/Utilities/README.md
iansuvak: Right now I prefer `Utilities` to match `Mocks` and `Teleporter` directories as we have them now. We do have #410 issue open and if we want to lower-case all of them we can do it there?

tests/flows/validator_set_sig.go
iansuvak: Yeah we can add it to #410

contracts/src/Utilities/ValidatorSetSig.sol
iansuvak: I thought about it originally but decided not to. It would be specific to the target contract and not relevant to the operation of this contract. Happy to change if we have a quorum that disagrees

contracts/src/Utilities/ValidatorSetSig.sol
iansuvak: We can but I prefer `valid` since that is what the Warp calls it in the interface definition. Thoughts?

contracts/src/Utilities/ValidatorSetSig.sol
iansuvak: But what's the benefit? If it's just style preference that's fine but if future solidity version provides a way to document start fields in NatSpec It's very unlikely to be `@param` and we would have to change it then again.

contracts/src/Utilities/ValidatorSetSig.sol
iansuvak: I think it makes sense for very user facing transactions but anyone seeing this is manually constructing a off-chain warp message and seeing the variable name gives them something to search the source code for. It's even more pronounced in later require message of `invalid validatorSetSigAddress`. IMO `invalid validator set sig address` is less clear and useful. And explaining of what the field is in the revert message would be even less clear? `warp message used is addressed to a different validator set sig contract address`

contracts/src/Utilities/ValidatorSetSig.sol
minghinmatthewlam: ye I tend to agree, and they will likely having more applicable events on their target contracts to look for.

contracts/src/Utilities/ValidatorSetSig.sol
minghinmatthewlam: works for me

contracts/src/Utilities/ValidatorSetSig.sol
minghinmatthewlam: that's fair in regards to easier searching or relating to the variable, sgtm.

contracts/src/Utilities/ValidatorSetSig.sol
minghinmatthewlam: better readability, keeping to a consistent style throughout the repo

bernard-avalabs (approved):
LGTM

contracts/src/Utilities/README.md
```diff
@@ -0,0 +1,18 @@
+## Utility Contracts
+
+### ValidatorSetSig Contract
+
+This contract provides an alternative to traditional multi-signature contracts where instead of requiring signatures from `K` of `N` pre-specified signers, an aggregate signature is required from a quorum of the current validators for a given blockchain.
+
+The contract leverages off-chain [Avalanche Warp Messages](https://docs.avax.network/build/cross-chain/awm/overview), which are manually approved for signing by a chain's validators. It requires these messages to have the source address set to the zero address to enforce this, since on-chain warp messages cannot have zero source address.
+
+Note:
+1. The blockchain validating the message may or may not be the same chain where the target contract and the `ValidatorSetSig` contract are deployed.
```
geoff-vball, line 10:
```suggestion
1. The blockchain validating the message may or may not be the same chain where the target contract and the `ValidatorSetSig` contract is deployed.
```

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+    bytes payload;
+}
+
+/**
+ * @dev Contract that verifies that a set threshold of validators from a given blockchainID
+ * have signed an off-chain Warp message and forwards the payload to the target contract specified in the ValidatorSetSigMessage.
+ * The threshold itself is set by the validator themselves in their Warp configs:
+ * https://github.com/ava-labs/subnet-evm/blob/master/precompile/contracts/warp/config.go#L50
+ *
+ * This is intended to be used for safe off-chain governance of enabled contracts. An example usecase would be
```
geoff-vball, line 42:
```suggestion
 * This is intended to be used for safe off-chain governance of enabled contracts. An example use case would be
```

contracts/src/Utilities/ValidatorSetSig.sol
```diff
@@ -0,0 +1,137 @@
...
+ * to deploy a `Ownable` target contract that is owned by an instance of this contract and adding the
```
geoff-vball, line 43:
```suggestion
 * to deploy an `Ownable` target contract that is owned by an instance of this contract and adding the
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	// ************************************************************************************************
+	// Setup
+	// ************************************************************************************************
+	// Deploy ValidatorSetSig expecting signatures from subetB instances to both subnets
```
geoff-vball, line 20:
```suggestion
	// Deploy ValidatorSetSig expecting signatures from subnetB instances to both subnets
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	// Deploy exampleERC20 instance to  both subnets
```
geoff-vball, line 21:
```suggestion
	// Deploy exampleERC20 instance to both subnets
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	// Test Case 1: validatorChain (subnetB) != targetChain (subnetA)
+	// ************************************************************************************************
+	// Send the off chain message to subnetA instance of ValidatorSetSig and confirm it is accepted.
+	// Confirm the event is emitted
+	// Retry the same message and confirm it fails
+	// Send a new message with incremented nonce and confirm it is accepted.
+	// ************************************************************************************************
+	// Test Case 2: validatorChain (subnetB) == targetChain (subnetB)
+	// ************************************************************************************************
+	// Send a new message to subnetB instance of ValidatorSetSig and  and confirm it is accepted
```
geoff-vball, line 34:
```suggestion
	// Send a new message to subnetB instance of ValidatorSetSig and confirm it is accepted
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+		Payload:                callData,
+	}
+
+	// Create chain config file with off chain validatorsetsig message
+	networkID := network.GetNetworkID()
+	offchainMessages, warpEnabledChainConfigWithMsg := utils.InitOffChainMessageChainConfigValidatorSetSig(
+		networkID,
+		subnetB,
+		validatorSetSigContractAddress,
+		[]validatorsetsig.ValidatorSetSigMessage{vssMessage1, vssMessage2, vssMessage3})
```
geoff-vball, line 127:
```suggestion
		[]validatorsetsig.ValidatorSetSigMessage{vssMessage1, vssMessage2, vssMessage3},
	)
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	// Setup
+	// ************************************************************************************************
+	// Deploy ValidatorSetSig expecting signatures from subetB instances to both subnets
+	// Deploy exampleERC20 instance to  both subnets
+	// Construct ValidatorSetSig message with mock ERC20 as the target contract
+	// Create off-chain Warp messages using the ValidatorSetSig message to be signed by the subnetB
+	// ************************************************************************************************
+	// Test Case 1: validatorChain (subnetB) != targetChain (subnetA)
+	// ************************************************************************************************
+	// Send the off chain message to subnetA instance of ValidatorSetSig and confirm it is accepted.
```
geoff-vball, line 27:
```suggestion
	// Send the off-chain message to subnetA instance of ValidatorSetSig and confirm it is accepted.
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	// Create a message for the case where validatorSetSig contract and targetContract are on the same chain.
+	vssMessage3 := validatorsetsig.ValidatorSetSigMessage{
+		ValidatorSetSigAddress: validatorSetSigContractAddress2,
+		TargetContractAddress:  exampleERC20ContractAddressB,
+		TargetBlockchainID:     subnetB.BlockchainID,
+		Nonce:                  big.NewInt(0),
+		Payload:                callData,
+	}
+
+	// Create chain config file with off chain validatorsetsig message
```
geoff-vball, line 121:
```suggestion
	// Create chain config file with off-chain validatorsetsig message
```

tests/flows/validator_set_sig.go
```diff
@@ -0,0 +1,235 @@
...
+	networkID := network.GetNetworkID()
+	offchainMessages, warpEnabledChainConfigWithMsg := utils.InitOffChainMessageChainConfigValidatorSetSig(
+		networkID,
+		subnetB,
+		validatorSetSigContractAddress,
+		[]validatorsetsig.ValidatorSetSigMessage{vssMessage1, vssMessage2, vssMessage3})
+
+	// Create chain config with off chain messages
```
geoff-vball, line 129:
```suggestion
	// Create chain config with off-chain messages
```

tests/utils/utils.go
```diff
@@ -370,6 +401,36 @@ func AddProtocolVersionAndWaitForAcceptance(
 	Expect(versionUpdatedEvent.NewVersion.Cmp(expectedLatestVersion)).Should(Equal(0))
 }
 
+// Returns Receipt for the transaction unlike TeleporterRegistry version since this is a non-teleporter case
+// and we don't want to add the  ValidatorSetSig ABI to the subnetInfo
```
geoff-vball, line 405:
```suggestion
// and we don't want to add the ValidatorSetSig ABI to the subnetInfo
```

tests/utils/utils.go
```diff
@@ -989,6 +1062,45 @@ func CreateOffChainRegistryMessage(
 	return unsignedMessage
 }
 
+func InitOffChainMessageChainConfigValidatorSetSig(
+	networkID uint32,
+	subnet interfaces.SubnetTestInfo,
+	validatorSetSigAddress common.Address,
+	validatorSetSigMessages []validatorsetsig.ValidatorSetSigMessage) ([]avalancheWarp.UnsignedMessage, string) {
```
geoff-vball, line 1069:
```suggestion
	validatorSetSigMessages []validatorsetsig.ValidatorSetSigMessage,
) ([]avalancheWarp.UnsignedMessage, string) {
```

geoff-vball (commented):
Just some spelling and formatting nits, but overall lgtm :)

contracts/src/Utilities/README.md
iansuvak: Both target contract and the validator  are included in the sentence above so `are` is correct.

geoff-vball: Oh yeah, my bad 😓

geoff-vball (approved):
👍

contracts/src/Utilities/README.md
minghinmatthewlam: sgtm

minghinmatthewlam (commented):
LGTM, only two questions left are high level on signatures for "valid off-chain Warp messsage", and `@param` for struct variables

contracts/src/Utilities/ValidatorSetSig.sol
iansuvak: Looks like we don't use `@param` anywhere in this repo yet. Since there is no special handling for this in terms of documentation generation and `@param` is a misnomer, what do people think about using `@field` or `@member` instead if we do want to prefix them?

michaelkaplan13 (approved):

minghinmatthewlam (approved):
LGTM 🚀

PR #413, opened 2024-07-05
Golang 1.21.12

michaelkaplan13:
## Why this should be merged

Bump to latest 1.21 Golang release. Removes unnecessary explicit version used by `actions/setup-go` in favor of pulling directly from `go.mod`.


cam-schultz (commented):
We should be able to remove `export GO_VERSION=${GO_VERSION:-$(getDepVersion go)}` from `versions.sh` (as well as `getDepVersion` from `utils.sh`. We were only using those in CI from what I can tell.

michaelkaplan13:
> We should be able to remove `export GO_VERSION=${GO_VERSION:-$(getDepVersion go)}` from `versions.sh` (as well as `getDepVersion` from `utils.sh`. We were only using those in CI from what I can tell.

Looks like `getDepVersion` is still used for the AvalancheGo/SubnetEVM versions, but I removed the GO_VERSION export. 👍

cam-schultz (approved):

iansuvak (approved):

Issue #410, opened 2024-07-01
[Code organization] - Move TeleporterRegistry to Utilities sub-folder 


iansuvak:
**Context and scope**
Once the Validator Set Sign contract #404 is merged we will have added a `Utilities` folder for standalone contracts. Based on the discussion in that PR we should move `TokenBridge` code outside into the Utilities subfolder and update relevant READMEs. 

 **Other potential improvements** 
Do we want to lowercase the directory names? I.e 'utilities', 'teleporter', 'mock'

PR #418, opened 2024-07-10
Create utilities and governance contract folders

geoff-vball:
## Why this should be merged
closes https://github.com/ava-labs/teleporter/issues/410

## How this works
Reorganized directory structure.

## How this was tested
Existing tests

## How is this documented
Edited existing documentation.


contracts/src/teleporter/ReceiptQueue.sol
cam-schultz, line 1: While we're here, what do you think of moving this file, `RentrancyGuards.sol`, and `SafeERC20TransferFrom.sol` to a new folder `contracts/src/utilities/utils`?

cam-schultz (commented):
One comment about adding an additional generic `utilities` folder, but otherwise LGTM. Also, I know that lowercasing the directory names was mentioned in the issue, but did we come to a conclusion on that? If we decide to move forward with that, we should do the same in `avalanche-interchain-token-transfer`.

iansuvak:
> Also, I know that lowercasing the directory names was mentioned in the issue, but did we come to a conclusion on that? If we decide to move forward with that, we should do the same in `avalanche-interchain-token-transfer`.

It was brought up in #404 and I think it makes sense overall. FWIW OZ uses lower cased directory names and ProperCase contract file names. I agree that if  we did it here we should do the same in `avalanche-interchain-token-transfer` repo as well. 

Overall LGTM. I just have one minor question below.

contracts/src/utilities/validators/README.md
```diff
@@ -1,6 +1,6 @@
-## Utility Contracts
```
iansuvak, line 1: I agree that it makes sense to separate different logical groups in sub-directories but I'm not sure about the `validators` name. Does anyone have a better idea? Maybe `governance`? I think we are more likely to have another governance type contract that would make sense to group this with than a validator related one?

contracts/src/utilities/upgrades/TeleporterOwnerUpgradeable.sol
iansuvak, line 1: In the original directory mockup on the #404 we agreed to keep Upgradeable files inside the `teleporter` folder.  Apologies for not copying it properly into the issue. I'm not opposed to keeping them here though

geoff-vball: Hmm, I do think it makes sense to keep `TeleporterRegistry` and `TeleporterUpgradeable` in the same spot. Would love to hear other opinions if anyone has one :)

contracts/src/teleporter/ReceiptQueue.sol
geoff-vball: Sounds good to me!

contracts/src/utilities/upgrades/TeleporterOwnerUpgradeable.sol
cam-schultz: I agree we should keep them together, especially if we rename `TeleporterUpgradeable` as discussed [here](https://github.com/ava-labs/teleporter/pull/397#discussion_r1671342631)

contracts/src/utilities/validators/README.md
minghinmatthewlam: i like governance

minghinmatthewlam (commented):
I overall still feel it's a bit strange to classify `TeleporterRegistry` and the other `upgrades` contracts under `utilities`, and the same for `governance` files. Thinking if those can sibling directories to `teleporter` or under `teleporter`. That way we can also remove the need to have another `utils` folder under `utilities`.

bernard-avalabs (commented):
Generally looks good to me. I agree with some of the name change suggestions made by others, such as validators->governance.

geoff-vball:
I'd be okay having `governance` and `upgrades` as top level directories, then we don't have to have `utilities/utils`.

geoff-vball:
@iansuvak I think I like the lowercase names. If we go ahead with https://github.com/ava-labs/teleporter/issues/419, then we can lowercase the directories in ICTT when we move them over.

iansuvak:
> @iansuvak I think I like the lowercase names. If we go ahead with #419, then we can lowercase the directories in ICTT when we move them over.

I agree and I think it makes sense to do it together with #419 implementation instead of a separate issue.

contracts/src/utilities/upgrades/TeleporterOwnerUpgradeable.sol
michaelkaplan13: I was originally thinking we'd put the registry contract under `utilities` but keep the `TeleporterUpgradeable`, etc under `Teleporter`. Names aside, the registry is a standalone contract that isn't inherently tied to Teleporter itself (could be used a general account versioning mechanism controlled by the validator set). The `TeleporterUpgradeable` contracts OTOH are abstract contracts that are tied to Teleporter in the sense that they import Teleporter interfaces and provide helpers specifically for sending and receiving Teleporter messages (by leveraging a registry instance).

contracts/src/governance/README.md
```diff
@@ -1,6 +1,6 @@
-## Utility Contracts
+# ValidatorSetSig Contract
```
iansuvak, line 1:
```suggestion
# Governance Contracts
```

nit: Can we change this back and `## Overview` below to `## ValidatorSetSig Contract` now that we have a `governance` folder?

minghinmatthewlam:
> I'd be okay having `governance` and `upgrades` as top level directories, then we don't have to have `utilities/utils`.

Definitely agree for `governance`. For `upgrades` I think the standard is to use it only in context of upgradeability in contracts, and think we should stick with that, so prefer `upgrades`-> `registry`. Then because the `registry` directory and all contracts under are tied to `teleporter`, i'd prefer to leave under `teleporter` folder but with the rename.

cam-schultz (approved):

iansuvak (approved):
LGTM!

minghinmatthewlam (approved):
LGTM

PR #420, opened 2024-07-11
Namespace storage

### LINKED ISSUE #415: Add namespace storage for upgradeable contracts
**Context and scope**
Implement ERC7201 namespace storage to upgradeable contracts to prevent storage collision between logic contract versions.

blocked by https://github.com/ava-labs/teleporter/pull/397

Alternative was using _gap variables, but namespace storage has advantages:

Is the new standard and used in OZv5
For non-upgradeable extensions no unnecessary _gap variables
Ideally we'll include upgrading OZ version to v5, and sol version to ^0.8.20 as well, but this would also be compatible with OZ v4 with our current setup.

minghinmatthewlam:
## Why this should be merged
Fixes #415 

## How this works
Adds namespace storage to all contracts according to ERC7201. Namespace storage helps prevent collision between storage of different logic contracts when used in the context of upgradeable contracts and proxies. 

Changes:
- all state variables are added to the contract's storage struct
- the storage struct is specified a specific storage slot with `@custom:storage-location` according to erc7201
- private variable for getting back the storage location
- private getter function and assembly to save storage struct inside `$` struct

## How this was tested
CI
TODO: should ideally have unit tests to make sure that the storage location are computed correctly.

## How is this documented
TBD


contracts/src/teleporter/registry/TeleporterUpgradeable.sol
```diff
@@ -35,18 +35,26 @@ abstract contract TeleporterUpgradeable is
...
     using SafeERC20 for IERC20;
 
-    // The Teleporter registry contract manages different Teleporter contract versions.
-    TeleporterRegistry public teleporterRegistry;
-    /**
-     * @dev A mapping that keeps track of paused Teleporter addresses.
-     */
-    mapping(address teleporterAddress => bool paused) private _pausedTeleporterAddresses;
+    // solhint-disable private-vars-leading-underscore
+    /// @custom:storage-location erc7201:teleporter.storage.TeleporterUpgradeable
```
cam-schultz, line 39: For my own understanding, do you know why the custom storage location has three forward slashes? This seems to be the convention in OZ's docs as well as the EIP 7201 spec, but I couldn't find any justification for it.

cam-schultz (approved):

contracts/src/teleporter/registry/TeleporterOwnerUpgradeable.sol
```diff
@@ -22,7 +22,15 @@ abstract contract TeleporterOwnerUpgradeable is TeleporterUpgradeable, OwnableUp
         address initialOwner
     ) internal onlyInitializing {
         __TeleporterUpgradeable_init(teleporterRegistryAddress);
-        __Ownable_init();
+        __Ownable_init_unchained();
```
michaelkaplan13, line 25: IIUC, `__Ownable_init` vs `__Ownable_init_unchained` doesn't have any functional difference here. 

For my understanding, is there a reason why `__Ownable_init_unchained` should be preferred? Seems like the normal `_init` would be recommended unless there is a specific need to ensure none of the `Ownable ` parent initializers are called

contracts/src/teleporter/registry/TeleporterUpgradeable.sol
michaelkaplan13: Not understanding the rationale behind this was bothering me also...dug into it and my mind was blown 😅 
image

michaelkaplan13: So "///" is for single line natspec comments, and "/**" is for multi-line natspec comments.

(Where all natspec comments start with a @ tag, and can be used to auto-generate docs)

michaelkaplan13 (approved):
One question, but LGTM

minghinmatthewlam:
merging into https://github.com/ava-labs/teleporter/pull/397

Issue #419, opened 2024-07-10
Repo Structure


michaelkaplan13:
**Context and scope**
In order to promote a more maintainable repo structure for smart contracts built on top of both AWM and Teleporter, and create a clear single entry point for anyone looking to learn about those contracts, we should structure the repo in such a way that provides an intuitive home for current and future projects.

Also related, "Warp" and "Teleporter" going forward will be referred to as "Avalanche Interchain Messaging (ICM)", so we should begin to have the repo reflect that.

**Discussion and alternatives**
This repository (`teleporter`) can become `avalanche-icm-contracts` to be the clear home for smart contract publish that build on top of the Avalanche ICM protocol. This includes the existing Teleporter protocol, but could also in the future include protocols for EVM event importing, ACP-77 staking contracts, etc. 

A rough breakdown of how the repository could be structured is:
```
audits/
lib/
contracts/
	Teleporter/
		Interfaces/
		Upgrades/
		Utils/
			ReceiptQueue
			ReentrancyGuards
			SafeERC20TransferFrom
	ICTT/
	Utilities/
		AccountRegistry/
		ValidatorSetSig/
scripts/
e2e-tests/
	Setup/
	Teleporter/
	ICTT/
```
**This is a very rough high level suggestion. The exact names can be determined by the implementer.**

**Open questions**
- On a go forwards basis, how can we make clear that the Teleporter contract should only be deployed via the static Nick's method transaction for official major release versions, which doesn't necessarily reflect the changes in `main` or in minor/patch releases?
- How do we iteratively move the "teleporter" name to follow the more standard "Avalanche ICM" conventions? (Should not be included in this issue either way).
- Where does the `teleporter-cli` fit in the above structure, or can it be removed?
- When should the Avalanche ICTT contracts be moved in? Can be now or in a later sub-issue.

PR #422, opened 2024-07-11
Contracts top level

cam-schultz:
## Why this should be merged

Converts the repo to be a top-level forge project. The new repo structure is as follows:
```
teleporter/
    lib/
    contracts/
        Teleporter/
        Utilities/
        Mocks/
    foundry.toml
    remappings.txt
```

`contracts-monorepo` is the staging branch for all changes related to #419

## How this works
Reogranizes the repo into the above structure

## How this was tested
CI

## How is this documented
Updated paths in READMEs


iansuvak (dismissed):

foundry.toml
```diff
@@ -1,9 +1,10 @@
 [profile.default]
-src = 'src'
+src = 'contracts'
 out = 'out'
 libs = ['lib']
 solc_version = '0.8.18'
 test = 'tests'
+ignored_warnings_from = ["contracts/Teleporter/tests", "contracts/Teleporter/upgrades/tests"]
```
cam-schultz, line 7: Some of the test files were throwing "Function state mutability can be restricted to pure/view" warnings that failed to compile when the suggested change was applied. Unfortunately forge doesn't support wildcard filters for this option, so we'll have to keep an eye on emitted warnings for any future tests we add.

.gitmodules
```diff
@@ -1,13 +1,14 @@
...
+	path = lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
+	ignore = dirty
 [submodule "lib/openzeppelin-contracts"]
 	branch = v4.8.1
-[submodule "contracts/lib/subnet-evm"]
-	path = contracts/lib/subnet-evm
+	path = lib/openzeppelin-contracts
+	url = https://github.com/openzeppelin/openzeppelin-contracts
+	ignore = dirty
```
michaelkaplan13, line 10: Did you mean to change the `lib/forge-std` branch and/or add the `ignore = dirty` flags here? Seems like just changing the paths would be sufficient.

contracts/src/.solhintignore
michaelkaplan13, line 1: Should this file be put in the base of the repo now?

foundry.toml
michaelkaplan13: Is there any reason why we couldn't fix the state mutability of those functions?

README.md
```diff
@@ -126,11 +123,19 @@ The user wallet set in `.env` must have native tokens for each of the Subnets us
 
 ## Upgradeability
 
-The Teleporter contract is non-upgradeable and can not be changed once it is deployed. This provides immutability to the contracts, and ensures that the contract's behavior at each address is unchanging. However, to allow for new features and potential bug fixes, new versions of the Teleporter contract can be deployed to different addresses. The [TeleporterRegistry](./contracts/src/Teleporter/TeleporterRegistry.sol) is used to keep track of the deployed versions of Teleporter, and to provide a standard interface for dApps to interact with the different Teleporter versions.
+<<<<<<< HEAD
```
geoff-vball, line 126: Looks like there are some merge artifacts, and some duplicated paragraphs in this file

geoff-vball (commented):
It looks like `contracts/src/.solhintignore` needs to be deleted

README.md
cam-schultz: Whoops, should be fixed now

cam-schultz:
> It looks like `contracts/src/.solhintignore` needs to be deleted

Is that not the case? I don't see it in the remote [branch](https://github.com/ava-labs/teleporter/tree/contracts-top-level/contracts) nor my local branch.

foundry.toml
cam-schultz: Fixed now 👍 I had my wires crossed while putting the original change together.

.gitmodules
cam-schultz: When putting together this change, I noticed a few things:
1. our forge-std submodule was using a different branch than openzeppelin-contract's submodule
2. We weren't specifying a path to our forge-std submodule in remappings.txt

From what I could tell, this means that the exact forge-std version that's used is whatever one is found first when resolving include paths. To get around this, I added forge-std to remappings.txt explicitly, and updated our submodule to use the same version as openzeppelin-contracts

As for `ignore = dirty`, that is added as a developer convenience. I noticed that when playing with submodules it was quite easy to get to a state in which no submodule changes were made, but it was still showing as modified in `git status`. `ignore = dirty` shouldn't impact CI or fresh clones.

.gitmodules
```diff
@@ -1,14 +1,16 @@
...
+	url = https://github.com/foundry-rs/forge-std
+	ignore = dirty
+[submodule "lib/openzeppelin-contracts"]
+	path = lib/openzeppelin-contracts
+	url = https://github.com/openzeppelin/openzeppelin-contracts
+	ignore = dirty
+[submodule "lib/openzeppelin-contracts-upgradeable"]
+	path = lib/openzeppelin-contracts-upgradeable
 	url = https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable
+	ignore = dirty
```
minghinmatthewlam, line 13: do we want ignore=dirty? IIUC if I make local changes to these submodules, git will not tell me in git status, but will it still build with these local changes without me knowing?

Also how come for subnet-evm there isn't the ignore tag

cam-schultz: Yeah, on further consideration masking local changes would likely cause more headaches than it's worth. Unintended local changes were very common when putting together this PR since the submodules themselves were moving around, but for more regular changes that should not be the case. Removed now.

contracts/governance/tests/ValidatorSetSigTests.t.sol
```diff
@@ -5,7 +5,7 @@
 
 pragma solidity 0.8.25;
 
-import {Test} from "forge-std/Test.sol";
+import {Test} from "@forge-std/Test.sol";
```
minghinmatthewlam, line 8: did we need this change? I think this was working previously from the auto remappings, but we could consider turning auto remappings off to be more explicit

contracts/teleporter/registry/tests/GetTeleporterMessengerTests.t.sol
```diff
@@ -10,14 +10,6 @@ import {ITeleporterMessenger} from "@teleporter/ITeleporterMessenger.sol";
 import {TeleporterMessenger} from "@teleporter/TeleporterMessenger.sol";
 
 abstract contract GetTeleporterMessengerTest is BaseTeleporterRegistryAppTest {
-    function testGetTeleporterMessengerBasic() public {
```
minghinmatthewlam, line 13: was there a reason to move this

minghinmatthewlam: seems like a lint with the new version, did it hit a lint rule?

contracts/governance/tests/ValidatorSetSigTests.t.sol
minghinmatthewlam: nvm I see that it's updated in remappings. If we're going to explicitly set it for forge-std that differs from the default, might want to consider turning off auto remapping detection

minghinmatthewlam (commented):
mostly LGTM, left some comments

contracts/governance/tests/ValidatorSetSigTests.t.sol
cam-schultz: I was encountering a scenario where the auto remapping used `open-zeppelin-contracts`' `forge-std` instead of ours.  IMO we should be explicit wherever possible to prevent this sort of ambiguity.

contracts/teleporter/registry/tests/GetTeleporterMessengerTests.t.sol
cam-schultz: Yeah, the linter complained about `public view` functions being before `public` ones. `solc` was throwing warnings saying that this function could be marked `view`. I'm not sure what change prompted `solc` to start warning though.

minghinmatthewlam:
do we need to update `.gitignore`? For example I see `contracts/doc` but dk if we still use that at all.

cam-schultz:
> do we need to update `.gitignore`? For example I see `contracts/doc` but dk if we still use that at all.

Let's continue ignoring generated docs from `forge docs`. I've updated the path to reflect the updated repo structure.

contracts/governance/tests/ValidatorSetSigTests.t.sol
minghinmatthewlam: i'm in favor of copying the remappings we need, then disabling auto remapping detection

geoff-vball (dismissed):
👍

minghinmatthewlam (approved):
lgtm 🙌

geoff-vball (approved):

michaelkaplan13 (approved):
🚢

meaghanfitzgerald:
Next time the file structures of any of the `.md`s change, pls ping @laviniat1996. These files are referenced by a plugin in the docs and when the path changes it causes the build of https://github.com/ava-labs/avalanche-docs to break.

PR #429, opened 2024-07-18
Tmpnet deferrals

feuGeneA:
## Why this should be merged

This PR addresses some comments from [the tmpnet PR](https://github.com/ava-labs/teleporter/pull/401) that were deferred.

### DIFF
--- a/tests/local/network.go
+++ b/tests/local/network.go
@@ -338,7 +338,7 @@ func (n *LocalNetwork) DeployTeleporterRegistryContracts(
 }
 
 func (n *LocalNetwork) GetSubnetsInfo() []interfaces.SubnetTestInfo {
-	subnetsInfo := make([]interfaces.SubnetTestInfo, 0)
+	subnetsInfo := make([]interfaces.SubnetTestInfo, 0, len(n.subnetsInfo))
 	for _, subnetInfo := range n.subnetsInfo {
 		subnetsInfo = append(subnetsInfo, *subnetInfo)
 	}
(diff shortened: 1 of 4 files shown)


iansuvak (dismissed):

tests/local/network.go
```diff
@@ -336,10 +338,11 @@ func (n *LocalNetwork) DeployTeleporterRegistryContracts(
 }
 
 func (n *LocalNetwork) GetSubnetsInfo() []interfaces.SubnetTestInfo {
-	return []interfaces.SubnetTestInfo{
-		*n.subnetsInfo[n.tmpnet.GetSubnet("A").SubnetID],
-		*n.subnetsInfo[n.tmpnet.GetSubnet("B").SubnetID],
+	subnetsInfo := make([]interfaces.SubnetTestInfo, 0)
```
michaelkaplan13, line 341: Performance doesn't matter much here, but when populating a new slice like this it is generally best to either:
- `make` the slice an initial size of 0 but an initial capacity large enough for all the entries to be added, and use `append` to add the elements to it (which will not require any additional allocs)
- `make` the slice with an initial size of the length of elements to be added, and the assign directly to the index within the for loop.

As is, it `append` could need to alloc a new array and copy all of the contents of the slice there on each iteration.

cam-schultz (commented):
LGTM once Michael's comment is addressed.

michaelkaplan13 (approved):

tests/local/network.go
geoff-vball: It doesn't actually allocate a new array on every iteration. Slices have an underlying capacity. If you append to a full slice, golang will allocate many extra slots, which you may or may not end up using. This gives an amortized runtime cost of O(1), but it's still good practice to allocate a slice of exactly the length you need if it is known. 

https://blog.lu4p.xyz/posts/slices/

cam-schultz (dismissed):

geoff-vball (dismissed):

cam-schultz (approved):

michaelkaplan13 (approved):

geoff-vball (approved):

PR #442, opened 2024-07-24
Set EVM version to shanghai, and Solidity to 0.8.25

michaelkaplan13:
## Why this should be merged

We previously have only used Solidity versions whose default EVM version were supported on Avalanche EVMs, but @ARR4N called out that there is no reason to have these be coupled. We can set the EVM version explicitly while still using the latest versions of Solidity to take advantage of other improvements. 

I suggest we stay one version behind the latest release (currently v0.8.26) going forward in order to stay up to date while also allowing time for bugs to be identified in the newest releases (Solidity does not maintain patch releases for one minor version behind latest as Golang does).

## How this works

Sets the EVM version explicitly to `shanghai` in `foundry.toml`, and bumps the Solidity version to 0.8.25.


contracts/src/.solhint.json
cam-schultz, line 1: This file seems to have a lot of formatting churn, probably due to differing IDE settings from different contributors. I agree with the changes here, but we should either add formatting to the repo itself, or reject future formatting only changes.

cam-schultz (dismissed):

contracts/src/.solhint.json
michaelkaplan13: Yeah I haven't spent the time to pin down the source of the auto-formatter changes but don't mind either way.

minghinmatthewlam: same pretty sure Michael and I have some differing settings, but fine for now.

geoff-vball (approved):

PR #443, opened 2024-07-24
Staking Contract Docs

cam-schultz:
## Why this should be merged
Adds overview documentation to `contracts/src/Staking` describing the high level usage of the Staking Manager contract, including specifications of the Warp messages involved in the subnet validator registration flow.

## How is this documented
New README


contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The `RegisterSubnetValidatorMessage` is delivered to the P-Chain as the payload of a `RegisterSubnetValidatorTx`. Please see the transaction [specification](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#step-2-issue-a-registersubnetvalidatortx-on-the-p-chain) for validity requirements. The successful `RegisterSubnetValidatorTx` results in a [`SubnetValidatorRegistrationMessage`](#subnetvalidatorregistrationmessage) Warp message indicating that the specified validation was successfully registered on the P-Chain.
+
+The `SubnetValidatorRegistrationMessage` is delivered to the Staking Manager contract via a call to `completeValidatorRegistration`. Staking rewards begin accruing at this time.
+
+#### Handling a Missed Expiry
+
+In the case of a missed expiry, the `RegisterSubnetValidatorTx` will result in a `SubnetValidatorRegistrationMessage` Warp message with the `valid` field set to `0`. This serves as proof that the corresponding validation has not started and may never start. The `SubnetValidatorRegistrationMessage` can be provided to the Staking Manager contract by calling `completeEndValidation` with `setWeightMessageType=false`.
```
geoff-vball, line 31:
```suggestion
In the case of a missed expiry, the `RegisterSubnetValidatorTx` will result in a `SubnetValidatorRegistrationMessage` Warp message with the `valid` field set to `0`. This serves as proof that the corresponding validation has not started and will never start. The `SubnetValidatorRegistrationMessage` can be provided to the Staking Manager contract by calling `completeEndValidation` with `setWeightMessageType=false`.
```

geoff-vball: Do we know if delivering the `RegisterSubnetValidatorTx` to the p-chain is idempotent? We have "resend" functions for all our communications with the p-chain, does the p-chain have the same?

contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Removing a Validator
```
geoff-vball, line 33:
```suggestion
### Removing yourself as a Validator
```

geoff-vball: There are usually ways to forcibly remove bad-acting validators, so the name of this section confused me at first.

contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+As an optional first step, an uptime proof may be requested from the P-Chain, which will be used the calculate the staking rewards on the Subnet. The proof is embedded in a [`ValidationUptimeMessage`](#validationuptimemessage) Warp message.
+
+Validators exit is initiated with a call to `initializeEndValidation` on the Staking Manager contract. A `ValidationUptimeMessage` Warp message may optionally be provided, otherwise the validator's uptime will be set to`0`. Once `initializeEndValidation` staking rewards cease accruing. The Staking Manager contract constructs a [`SetSubnetValidatorWeightMessage`](#setsubnetvalidatorweightmessage) Warp message, setting the weight to `0`.
```
geoff-vball, line 37: Just thinking it would probably make sense to provide two external functions - one that expects rewards to be disbursed, and one that exits without rewards. This would avoid someone accidentally not including their uptime proof, which could be a costly mistake.

contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The `SetSubnetValidatorWeightMessage` is delivered to the P-Chain as the payload of a `SetValidatorWeightTx`. This results in another `SetSubnetValidatorWeightMessage` Warp message, this time constructed by the P-Chain.
+
+This `SetSubnetValidatorWeightMessage` is delivered to the Staking Manager contract with a call to `completeEndValidation` with `setWeightMessageType=true`, at which point staking rewards are disbursed and the validation is removed from the contract's state.
```
geoff-vball, line 41:
```suggestion
This `SetSubnetValidatorWeightMessage` is delivered to the Staking Manager contract with a call to `completeEndValidation` with `setWeightMessageType=true`, at which point staking rewards are disbursed, stake is returned, and the validation is removed from the contract's state.
```

contracts/src/staking/StakingMessages.sol
```diff
@@ -30,7 +30,7 @@ library StakingMessages {
     // Full test coverage should be provided, and the implementation should be optimized for gas efficiency.
 
     /**
-     * @notice Packs a RegisterSubnetValidator message into a byte array.
+     * @notice Packs a SetSubnetValidatorWeightMessage message into a byte array.
```
geoff-vball, line 33:
```suggestion
     * @notice Packs a RegisterSubnetValidatorMessage message into a byte array.
```

contracts/src/staking/README.md
bernard-avalabs, line 37: Is an optional "ValidationUptimeMessage" message being provided the same as an uptime proof being requested by the P-Chain (line 35)? If they are not, when would you use one over the other?

bernard-avalabs: Adding to this, my understanding (which may be incorrect) is that Warp messages can be delivered out of order. If this is correct, then is there a chance that the SetSubnetValidatorWeightMessage message be delivered before the ValidationUptimeMessage message?

contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Removing a Validator
+
+As an optional first step, an uptime proof may be requested from the P-Chain, which will be used the calculate the staking rewards on the Subnet. The proof is embedded in a [`ValidationUptimeMessage`](#validationuptimemessage) Warp message.
+
+Validators exit is initiated with a call to `initializeEndValidation` on the Staking Manager contract. A `ValidationUptimeMessage` Warp message may optionally be provided, otherwise the validator's uptime will be set to`0`. Once `initializeEndValidation` staking rewards cease accruing. The Staking Manager contract constructs a [`SetSubnetValidatorWeightMessage`](#setsubnetvalidatorweightmessage) Warp message, setting the weight to `0`.
+
+The `SetSubnetValidatorWeightMessage` is delivered to the P-Chain as the payload of a `SetValidatorWeightTx`. This results in another `SetSubnetValidatorWeightMessage` Warp message, this time constructed by the P-Chain.
```
bernard-avalabs, line 39: Maybe I'm misunderstanding something, but how is a Warp message delivered as a payload of a transaction? Is the message included in the payload?

bernard-avalabs (commented):
The README reads well. It might be helpful to include a few diagrams to illustrate the interactions between the P-chain and the staking contract.

contracts/src/staking/README.md
cam-schultz: Replay protection is provided by the `expiry` field in the `RegisterSubnetValidatorTx` Warp message. I need to think more about the conditions under which resending is necessary. Since the staking contracts represent the only path by which validators can be added, the semantics around validator churn are slightly different than the general case.

contracts/src/staking/README.md
cam-schultz: An alternative would be to update the redeemable amount internally on validation completion, and have disbursement be a separate step, similar to `redeemRelayerRewards` in `TeleporterMessenger`.

cam-schultz: > is there a chance that the SetSubnetValidatorWeightMessage message be delivered before the ValidationUptimeMessage message?

This is only possible if `completeEndValidation` is called before `initializeEndValidation`, in which case the default uptime value of 0 will be used.

contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+## Usage
+
+The common case for adding or removing a validator from a Subnet's validator set follows these general steps:
+1. Inform the Staking Manager contract of the change. The Staking Manager contract will construct a Warp message attesting to the change in the validator set.
+2. Deliver the Warp message containing the validator set change to the P-Chain. The P-Chain will construct a Warp message acknowledging the change to the validator set.
+3. Deliver the Warp message containing the P-Chain acknowledgement back to the Staking Manager contract on the Subnet. The Staking Manager contract finalizes the validator set change.
+
+### Initializing a Validator Set
+
+If the Subnet has no validators registered on the P-Chain, then it will not be able to sign a Warp message to add a validator (step 1 above). In this case, Subnet recovery must be performed on the P-Chain, which allows an initial validator set to be specified. The P-Chain constructs a Warp message (format specification TBD) describing the initial validator set (step 2 above).
```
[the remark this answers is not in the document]
cam-schultz: This draft PR has the first existing description that I'm aware of: https://github.com/avalanche-foundation/ACPs/pull/135/files#diff-d2befccafcc84a6c9bf56f76def8e04647ed5087fe8480f28de11fea63fb769aL75

I'll update the docs here to reflect this.

contracts/src/staking/README.md
cam-schultz: Yup, the new P-Chain transaction types specified in ACP-77 include a `warp.Message` field.

contracts/src/staking/README.md
cam-schultz: The semantics of the `valid` bit guarantee that the corresponding validation _has not_ and _will_ never start if `valid=0`.

contracts/src/staking/README.md
cam-schultz: They're part of the same optional step. I reorganized this section to make this a bit clearer.

contracts/src/staking/README.md
cam-schultz: I added that this is specifically a Warp message payload field of the transaction type.

contracts/src/staking/README.md
cam-schultz: Renamed to `Exiting the Validator Set`

contracts/src/staking/README.md
geoff-vball: I guess my question is - What happens on the p-chain if we deliver the `RegisterSubnetValidatorTx` twice? Does it return a valid `SubnetValidatorRegistrationMessage` in both cases?

cam-schultz: I'm not sure what will happen on the P-Chain in that case. However, a `SubnetValidatorRegistrationMessage` with `valid=0` should only mean that the validation in question is not active and can never become active, and should not reflect the `RegisterSubnetValidatorTx` transaction status otherwise.

contracts/src/staking/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The `SubnetValidatorRegistrationMessage` is delivered to the Staking Manager contract via a call to `completeValidatorRegistration`. Staking rewards begin accruing at this time.
+
+#### Handling a Missed Expiry
+
+In the case of a missed expiry, the `RegisterSubnetValidatorTx` will result in a `SubnetValidatorRegistrationMessage` Warp message with the `valid` field set to `0`. This serves as proof that the corresponding validation has not started and may never start. The `SubnetValidatorRegistrationMessage` can be provided to the Staking Manager contract by calling `completeEndValidation` with `setWeightMessageType=false`.
+
+### Removing a Validator
+
+As an optional first step, an uptime proof may be requested from the P-Chain, which will be used the calculate the staking rewards on the Subnet. The proof is embedded in a [`ValidationUptimeMessage`](#validationuptimemessage) Warp message.
```
bernard-avalabs:
```suggestion
As an optional first step, an uptime proof may be requested from the P-Chain, which will be used to the calculate the staking rewards on the Subnet. The proof is embedded in a [`ValidationUptimeMessage`](#validationuptimemessage) Warp message.
```

bernard-avalabs (commented):
The updates look good to me. Left a comment about a typo.

geoff-vball (approved):

Issue #438, opened 2024-07-22
Implement resendEndValidatorMessage


cam-schultz:
**Context and scope**
This method is currently unimplemented in the PoC

geoff-vball:
Completed via https://github.com/ava-labs/teleporter/pull/447

PR #447, opened 2024-07-25
Implement resendEndValidatorMessage

geoff-vball:
## Why this should be merged
closes https://github.com/ava-labs/teleporter/issues/438

Also fixes a couple TODOs


contracts/src/staking/StakingManager.sol
```diff
@@ -355,7 +356,18 @@ abstract contract StakingManager is ReentrancyGuard, IStakingManager {
...
+        StakingManagerStorage storage $ = _getStakingManagerStorage();
+        Validator memory validator = $._validationPeriods[validationID];
+
+        require(
+            validator.status == ValidatorStatus.PendingRemoved,
+            "StakingManager: Validator not pending removal"
+        );
+
+        bytes memory setValidatorWeightPayload = StakingMessages.packSetSubnetValidatorWeightMessage(
+            validationID, validator.messageNonce, 0
```
bernard-avalabs, line 368: We are reusing the previous message nonce. Is there any case where we may need to manually set the message nonce? I suppose no additional new messages should be sent after initializeEndValidation.

bernard-avalabs (commented):
Generally makes sense to me.

contracts/src/staking/StakingManager.sol
cam-schultz: The dynamics of this nonce are specified [here](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#setsubnetvalidatorweighttx). Specifically, the `minNonce` associated with the `validationID` tracked by the P-Chain is incremented whenever a `SetSubnetValidatorWeightTx` is accepted. Since this method is only intended to be used when the `SetSubnetValidatorWeightTx` cannot be accepted due to validator churn, it should be safe to reuse the nonce.

contracts/src/staking/StakingManager.sol
```diff
@@ -332,6 +332,7 @@ abstract contract StakingManager is ReentrancyGuard, IStakingManager {
             uptimeSeconds = uptime;
         }
         validator.uptimeSeconds = uptimeSeconds;
+        validator.messageNonce++;
```
cam-schultz, line 335: Is incrementing the nonce here necessary? My understanding is that at this point, `minNonce == 0` for the validation ID, so the resulting `SetSubnetValidatorWeightTx` should be accepted with `nonce = 0`. The nonce should be incremented for any `SetSubnetValidatorWeightTx` that do not exit the validator set.

geoff-vball: That sounds right. After this function is called, no weight adjustments should be allowed, so we shouldn't need to update the nonce.

michaelkaplan13 (approved):

cam-schultz (approved):

PR #450, opened 2024-07-25
e2e test: use fewer nodes

feuGeneA:
## Why this should be merged
Because it reduced my test run time from 3m39s to to 2m57s, a 19.2% reduction

## How this works
Uses fewer nodes for each of the subnets and for the pool of non-subnet-validator nodes.

## How this was tested
I tried to reduce the number of nodes to 1 for each category (subnetA, subnetB, non-subnet-validator) but it gave errors about:
- for the non-subnet-validator: not being connected to enough stake (connected to 66.6% but want 80%
- for the subnet validators: not connected to enough peers (have 0, need 1)


tests/flows/validator_churn.go
```diff
@@ -14,10 +14,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-const (
-	nodesPerSubnet = 5
```
feuGeneA, line 18: this variable was apparently not used anywhere so I removed it

geoff-vball (approved):

iansuvak (approved):

tests/flows/validator_churn.go
```diff
@@ -14,10 +14,7 @@ import (
...
-	newNodeCount   = 5
-)
+const newNodeCount = 2
```
cam-schultz, line 17: Just noticing this now, but `newNodeCount` is a bit of a misnomer. It represents the number of nodes added to the validator set, not the number of validators after the operation.

feuGeneA: ah I can see that interpretation now that you mention it. I had interpreted it as "count of new nodes", not "new count of nodes". definitely ambiguous!

PR #453, opened 2024-07-26
WIP: Staking contract

geoff-vball:
This is a draft PR for the Staking Contract feature branch. This is to make sure CI is passing, and to have a place for team members to leave comments.

This should roughly track with this milestone https://github.com/ava-labs/teleporter/milestone/2


contracts/staking/StakingManager.sol
```diff
@@ -0,0 +1,447 @@
...
+        uint256 value,
+        uint64 registrationExpiry,
+        bytes memory signature
+    ) internal nonReentrant returns (bytes32) {
+        StakingManagerStorage storage $ = _getStakingManagerStorage();
+
+        // Ensure the registration expiry is in a valid range.
+        require(
+            registrationExpiry > block.timestamp && block.timestamp + 2 days > registrationExpiry,
+            "StakingManager: Invalid registration expiry"
```
ARR4N, line 155: [Solidity 0.8.26 allows you to use custom errors in `require()`](https://soliditylang.org/blog/2024/05/21/solidity-0.8.26-release-announcement/#:~:text=Custom%20errors%20support%20in%20require). They result in smaller contract sizes and are IMO much better for tests because you can assert the exact reason for a failure without risk of introducing a [change-detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) if you modify the error string. If you want to stick to 0.8.25 then I'd recommend switching to `if (!cond) { revert SpecificError() }` where you originally had `require(cond, ...)`.

contracts/staking/NativeTokenStakingManager.sol
```diff
@@ -0,0 +1,47 @@
+// (c) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: Ecosystem
```
ARR4N, line 4: (nit) Confirm with a lawyer, but I think this should be `UNLICENSED` per [Solidity conventions for non-OSS licenses](https://docs.soliditylang.org/en/v0.8.26/layout-of-source-files.html#:~:text=If%20you%20do%20not%20want%20to%20specify%20a%20license%20or%20if%20the%20source%20code%20is%20not%20open%2Dsource%2C%20please%20use%20the%20special%20value%20UNLICENSED.) (_not_ `UNLICENSE` without the `D`). There are [specific license identifiers for SPDX](https://spdx.org/licenses/).

contracts/staking/StakingMessages.sol
```diff
@@ -0,0 +1,289 @@
...
+     * | validationID : [32]byte | 32 bytes |
+     * +--------------+----------+----------+
+     * |        nonce :   uint64 |  8 bytes |
+     * +--------------+----------+----------+
+     * |       weight :   uint64 |  8 bytes |
+     * +--------------+----------+----------+
+     *                           | 52 bytes |
+     *                           +----------+
+     */
+    function packSetSubnetValidatorWeightMessage(
```
ARR4N, line 131: This can be simplified to:

```solidity
return abi.encodePacked(
    SET_SUBNET_VALIDATOR_WEIGHT_MESSAGE_TYPE_ID,
    validationID,
    nonce,
    weight
);
```

ARR4N: There are similar improvements that can be made to unpacking, but it was easier to write code so I created [a PR, based off this branch, to demonstrate](https://github.com/ava-labs/teleporter/pull/482).

Together they save about 39,500 gas on the pack-unpack round trip.

ARR4N: I only just saw the `TODO` re gas efficiency 🤦 sorry!

contracts/staking/StakingManager.sol
cam-schultz: Our versioning policy is to use the second-most-recent release to hedge against potential issues in the latest release. That said, we've struggled to choose between revert strings and custom errors in the past, so getting the best of both worlds will be a welcome change.

contracts/staking/StakingManager.sol
```diff
@@ -0,0 +1,447 @@
...
+    function completeEndValidation(uint32 messageIndex, bool setWeightMessageType) external {
+        StakingManagerStorage storage $ = _getStakingManagerStorage();
+
+        // Get the Warp message.
+        (WarpMessage memory warpMessage, bool valid) =
+            WARP_MESSENGER.getVerifiedWarpMessage(messageIndex);
+        require(valid, "StakingManager: Invalid warp message");
+
+        bytes32 validationID;
+        if (setWeightMessageType) {
```
ceyonur, line 374: did we confirm that both SetWeightTx=0 and InvalidMsg can be signed for this one? I wonder if it makes a better UX to expect a single msg type to be signed/relayed.

PR #456, opened 2024-07-26
ABIPacker interface + testing

### LINKED ISSUE #411: [tests] Improve struct packing testing in packing_test.go files
**Context and scope**
There are multiple `packing_test.go` files accompanying `packing.go` files that test packing methods. These tests don't check for exhaustiveness of the packing and the structs and would still succeed if additional fields were added to the Solidity structs but the packing methods weren't updated.  We should fix this. 
 
@cam-schultz pointed out that https://github.com/GaijinEntertainment/go-exhaustruct is a potential candidate for implementing these exhaustiveness tests and is already included in golang-ci. We need to check whether we can point the linter at specific files only if not. 

Adding `//exhauststruct:enforce` to struct definitions with it being off globally would do the trick but since the structs are auto-generated, adding this would be a change to the `abigen` in `subnet-evm` that would probably have to be flag gated in order to not affect existing usage.

iansuvak:
## Why this should be merged
Closes #411
## How this works
Defines `ABIPacker` interface and replaces existing Pack<typeName> functions with Pack/Unpack pointer receiver methods to make the structs implement the new interface.
## How this was tested
Added a new test that confirms that any struct that implements `ABIPacker` interface populated with random values encodes and decodes to same values. 

Randomizing all fields of the struct ensures exhaustiveness of the test. I confirmed this by removing the last line in the `abi.NewType`  call and confirmed that the test fails in that case. 

## How is this documented
Need to add note about this to `abi-bindings/README.md`


abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,190 @@
...
+		encoded, err := v1.Pack()
+		require.NoError(t, err)
+		v2 := reflect.New(reflect.TypeOf(v1).Elem()).Interface().(ABIPacker)
+		err = v2.Unpack(encoded)
+		require.NoError(t, err)
+		require.Equal(t, v1, v2)
+	}
+}
+
+func randomizeStruct(packerStruct interface{}) (interface{}, error) {
```
iansuvak, line 123: If we think that these randomization functions are useful we can maybe add them to a test utils location?

abi-bindings/README.md
```diff
@@ -1,3 +1,7 @@
...
\ No newline at end of file
+This directory contains ABI bindings for the Solidity contracts in the `contracts/src/CrossChainApplications` and `contracts/src/Teleporter` directories. The files with the same name as the Solidity source files are automatically generated by the `scripts/abi_bindings.sh` script. Other files in this directory (such as the packing utilities) are manually created and maintained.
+
+## Packing
+
+`packing.go` files in individual subfolders define utilities for ABI packing instances of structs auto-generated by `abigen` as well as method calls. For structs we implement `ABIPacker` interface defined in `./packer/packer.go` and add references to the `packer_test.go` file to ensure that our tests are exhaustive and don't fail silently if additional fields are added to the structs in the future on the Solidty side.
```
minghinmatthewlam, line 7: generally have tried to avoid "we" and "our"

iansuvak: Thanks. Made the update.

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,190 @@
...
+										}
+									}
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
michaelkaplan13, line 96: I'm not exactly sure what this part of the test is doing, but generally when something has this many nested statements it gets hard to read. Can the if statements be inverted (i.e. return early in failure case and have no nesting beneath it), and loop bodies broken out into helpers?

abi-bindings/README.md
```diff
@@ -1,3 +1,7 @@
...
\ No newline at end of file
+This directory contains ABI bindings for the Solidity contracts in the `contracts/src/CrossChainApplications` and `contracts/src/Teleporter` directories. The files with the same name as the Solidity source files are automatically generated by the `scripts/abi_bindings.sh` script. Other files in this directory (such as the packing utilities) are manually created and maintained.
+
+## Packing
+
+`packing.go` files in individual subfolders define utilities for ABI packing instances of structs auto-generated by `abigen` as well as method calls. For structs, `ABIPacker` interface defined in `./packer/packer.go` needs to be implemented and mapping to it's instance added to the `packer_test.go` file to ensure that the tests are exhaustive and don't fail silently if additional fields are added to the structs in the future on the Solidty side.
```
michaelkaplan13, line 7:
```suggestion
This directory contains ABI bindings for the Solidity contracts in the `contracts/src/` directory. The files with the same name as the Solidity source files are automatically generated by the `scripts/abi_bindings.sh` script. 

The `packing.go` files in individual subfolders define utilities for ABI packing instances of structs auto-generated by `abigen` as well as method calls. For structs, the `ABIPacker` interface defined in `./packer/packer.go` needs to be implemented and mapped to its instance added to the `packer_test.go` file to ensure that the tests are exhaustive and don't fail silently if additional fields are added to the structs in the future on the Solidity side.
```

abi-bindings/go/governance/ValidatorSetSig/packing.go
```diff
@@ -29,37 +29,29 @@ func init() {
...
-	type validatorSetSigWarpPayload struct {
-		ValidatorSetSigMessage ValidatorSetSigMessage `json:"validatorSetSigMessage"`
-	}
-	var payload validatorSetSigWarpPayload
-	err = args.Copy(&payload, unpacked)
-	if err != nil {
-		return ValidatorSetSigMessage{}, err
-	}
-	return payload.ValidatorSetSigMessage, nil
+	err = args.Copy(&m, unpacked)
```
michaelkaplan13, line 53: Is the `Copy` here necessary, or could we `Unpack` directly into `&m`? If it is necessary, lets just `return args.Copy(...)`.

abi-bindings/go/teleporter/registry/TeleporterRegistry/packing_test.go
```diff
@@ -16,11 +16,23 @@ func TestPackUnpackProtocolRegistryEntry(t *testing.T) {
 		Version:         big.NewInt(1),
 		ProtocolAddress: common.HexToAddress("0x0123456789abcdef0123456789abcdef01234567"),
 	}
+	//
```
michaelkaplan13, line 19: Stray comment 🙂

abi-bindings/go/packer/packer_test.go
iansuvak: I absolutely agree in general.   Ast traversal for very particular types is the  only reason why I would do nesting like this without breaking it out into helpers. I prefer it like this for 2 reasons:
:one: The flow here is of zeroing down on a particular thing. In this case instances of the `ABIPacker` interface. With each nesting you are closer to the specific thing that you are looking for.
:two: If breaking out in helpers the interfaces types on the functions wouldn't be helpful. These are all `types.Object` interfaces and get cast to different more specific interfaces (or fail the cast). 

Also FWIW. Github makes it look a lot more egregious visually since it uses 8 spaces for a tab.

It does sound like we have two people who prefer doing something to this so I would be happy to but I just wanted to get my 2 cents in 😄

iansuvak: I unrolled them and added some comments. I think I still prefer the previous structure but it sounds like I'm the odd one out 😄

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+func findAllImplementers(t *testing.T) []string {
+	cfg := &packages.Config{
+		Mode: packages.NeedName | packages.NeedFiles | packages.NeedSyntax | packages.NeedTypes | packages.NeedTypesInfo,
+		Fset: fs,
+	}
+	interfacePkg, err := packages.Load(cfg, currentPackagePath)
+	require.NoError(t, err)
+
+	var iface *types.Interface
+	// find the interface in the currnet package
```
minghinmatthewlam, line 47:
```suggestion
	// find the interface in the current package
```

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+var fs = token.NewFileSet()
+
+var packerTypes = map[string]ABIPacker{
+	"ValidatorSetSigMessage": &validatorsetsig.ValidatorSetSigMessage{},
+	"TeleporterMessage":      &teleportermessenger.TeleporterMessage{},
+	"ProtocolRegistryEntry":  &teleporterregistry.ProtocolRegistryEntry{},
+}
+
+func findAllImplementers(t *testing.T) []string {
```
cam-schultz, line 38: What's the advantage of dynamically fetching the implementers if we still have to maintain a hardcoded list of them in `packerTypes`? If a type implements `ABIPacker`, then the exhaustiveness check will be run against it, which fails if it's not also named in `packerTypes`.

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+const interfaceName = "ABIPacker"
+
+// Paths below are relative to the current package.
+const packagePath = "../..."
+const currentPackagePath = "."
+
+var fs = token.NewFileSet()
+
+var packerTypes = map[string]ABIPacker{
+	"ValidatorSetSigMessage": &validatorsetsig.ValidatorSetSigMessage{},
```
cam-schultz, line 33: If we ever have multiple tests in here that modify the values in this map, then the exhaustiveness test may incorrectly pass. For instance if an additional test runs before `TestExhaustivePacking` that populates a field that is erroneously left out of the corresponding `Pack` method, then `TestExhaustivePacking` would still pass.

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+		encoded, err := v1.Pack()
+		require.NoError(t, err)
+		v2 := reflect.New(reflect.TypeOf(v1).Elem()).Interface().(ABIPacker)
+		err = v2.Unpack(encoded)
+		require.NoError(t, err)
+		require.Equal(t, v1, v2)
+	}
+}
+
+func randomizeStruct(packerStruct interface{}) (interface{}, error) {
```
cam-schultz, line 137: I'd recommend renaming these to `populateStruct` and `populateField` so that there's no misconceptions that the test relies on randomness.

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return packerStruct, nil
+}
+
+func randomizeField(field reflect.Value) error {
+	fieldType := field.Type()
+	switch fieldType.Kind() {
+	// The list only includes types supported by abigen
+	case reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		field.SetInt(rand.Int63())
```
cam-schultz, line 155: I'm not sure how ints are capped, but it would be clearer to read if we instead bounded the random value to be within the smallest integer's range, or else hardcoded to a static value also within that range.

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+		field.SetUint(rand.Uint64())
+	case reflect.Bool:
+		field.SetBool(rand.Intn(2) == 1)
+	case reflect.String:
+		field.SetString(randomString(rand.Intn(256)))
+	case reflect.Slice:
+		n := rand.Intn(256)
+		field.Set(reflect.MakeSlice(fieldType, n, n))
+		for i := 0; i < field.Len(); i++ {
+			err := randomizeField(field.Index(i))
```
cam-schultz, line 166: What if it's an slice of structs?

cam-schultz: Or a multi dimensional slice?

abi-bindings/go/packer/packer_test.go
```diff
@@ -0,0 +1,204 @@
...
+		for j := 0; j < field.Len(); j++ {
+			field.Index(j).SetUint(uint64(rand.Intn(256)))
+		}
+
+	case reflect.Pointer:
+		if fieldType.Elem() == reflect.TypeOf(big.Int{}) {
+			newBigInt := big.NewInt(rand.Int63())
+			field.Set(reflect.ValueOf(newBigInt))
+		} else {
+			return fmt.Errorf("Unsupported pointer type %s", fieldType.Elem().String())
```
cam-schultz, line 188: Where is it documented that big.Int is the only supported pointer type?

abi-bindings/go/teleporter/registry/TeleporterRegistry/packing.go
```diff
@@ -33,6 +33,35 @@ func init() {
...
+
+func (p *ProtocolRegistryEntry) Unpack(b []byte) error {
+	args := abi.Arguments{
+		{
+			Name: "protocolRegistryEntry",
+			Type: protocolRegistryEntryType,
+		},
+	}
+	unpacked, err := args.Unpack(b)
+	fmt.Println("unpacked: ", unpacked)
```
cam-schultz, line 57: Let's remove this line

abi-bindings/go/governance/ValidatorSetSig/packing_test.go
cam-schultz, line 1: Are the type-specific `packing_tests` necessary anymore? AFAIK they all just pack and unpack an arbitrary instance of the struct and ensure that they're equal, which is exactly what we do in the randomized variant at the `ABIPacker` interface level.

abi-bindings/go/packer/packer_test.go
cam-schultz, line 1: There are a number of references to `abigen` in this file. It would be helpful to provide links to the code in question when referencing `abigen` features.

abi-bindings/go/packer/packer_test.go
iansuvak: If you forget to add it to the `packerTypes`, which is very likely without this enforcement mechanism, then the test is rather useless. It only checks for types that we remembered to add to the list. 

Originally I didn't want to require the explicit list and only find them programmatically but I didn't find a way to convert from `types` types that the `"golang.org/x/tools/go/packages"` uses and the `reflect` types that the randomizer uses hence I use the names of the types as a "bridge".

abi-bindings/go/packer/packer_test.go
iansuvak: Honestly, initially ChatGPT gave me a type conversion table but I confirmed it by looking at [Solidity abi-spec ](https://docs.soliditylang.org/en/latest/abi-spec.html) and the abi-gen code itself. All solidity types should be accounted for but if this changes in the future we will get an error in tests and it should be an easy fix to account for additional supported types.

abi-bindings/go/governance/ValidatorSetSig/packing_test.go
iansuvak: They shouldn't be. I'd be happy to remove it but some do extra things like `TeleporterRegistry` one.

abi-bindings/go/teleporter/registry/TeleporterRegistry/packing.go
iansuvak: Good catch, I removed the line in other files but missed this one.

abi-bindings/go/packer/packer_test.go
iansuvak: What features are you thinking of? The main thing that I see is the type conversion mapping and there is no good documentation that I found but I included the table of assumed type conversions to the README and a reference to it in the `randomizeField` function.

geoff-vball (dismissed):
👍

abi-bindings/go/packer/packer_test.go
cam-schultz: Cleared this up. The values are only used to get an instance of a concrete type for reflection.

abi-bindings/go/packer/packer_test.go
cam-schultz: > None of this matters since we don't rely on randomness for the validity of the test

Good enough for me!

abi-bindings/go/governance/ValidatorSetSig/packing_test.go
cam-schultz: My vote is to remove all of the one's that simply pack/unpack, but happy to defer to a followup task.

abi-bindings/go/packer/packer_test.go
cam-schultz: It was mostly the supported pointer types that I noticed. The README additions you added suffice 👍

abi-bindings/go/packer/packer_test.go
cam-schultz: Thanks for the explanation!

abi-bindings/go/packer/packer_test.go
```diff
@@ -29,12 +29,17 @@ const currentPackagePath = "."
 
 var fs = token.NewFileSet()
 
+// These mapping of names to empty type instances is necessary to be able to bridge
+// between the name of the type obtained by the AST parsing in findAllImplementers
+// and the actual type that we can instantiate and populate using reflection.
```
cam-schultz, line 34:
```suggestion
// and the actual type that we can instantiate and populate using reflection.
// If a new type is added that implements ABIPacker, but is not added to this list,
// then the test will fail.
```

cam-schultz (commented):
One small suggestion to document what happens if a new type is not added to the hardcoded list (the test should fail). Otherwise LGTM.

cam-schultz (approved):

geoff-vball (approved):

PR #462, opened 2024-07-30
E2E flaky test fix

geoff-vball:
## Why this should be merged
Fixes test flakiness introduced with the switchover to tmpnet.

Also contains some formatting

## How this works
Sorts the slice created by `GetSubnetsInfo()` before returning it. We were adding items to a slice by ranging over a map, which gives a random order. Therefore calling `GetTwoSubnets()` could return the two subnets in the opposite order if called twice in one test.

## How this was tested
Existing E2E tests.

## How is this documented
N/A


tests/local/network.go
```diff
@@ -464,7 +471,7 @@ func (n *LocalNetwork) AddSubnetValidators(ctx context.Context, subnetID ids.ID,
 
 	// consume some of the extraNodes
 	newValidatorNodes := n.extraNodes[0:count]
-	defer slices.Delete(n.extraNodes, 0, int(count))
+	copy(n.extraNodes, n.extraNodes[count:])
```
cam-schultz, line 474: This will still preserve the nodes to be deleted in `n.extraNodes`, since `n.extraNodes[count:]` is a subset. See example here: https://go.dev/play/p/z8RhPp136sr

iansuvak (approved):
Lots of formatting only changes but the core of the fix makes sense to me.

cam-schultz (commented):
Nice find! One comment to revert an unrelated change.

tests/local/network.go
geoff-vball: Fixed by copying the `newValidatorNodes` from `extraNodes` and then just assigning `extraNodes` to the subset.

tests/local/network.go
```diff
@@ -470,8 +470,9 @@ func (n *LocalNetwork) AddSubnetValidators(ctx context.Context, subnetID ids.ID,
 	)]
 
 	// consume some of the extraNodes
-	newValidatorNodes := n.extraNodes[0:count]
-	copy(n.extraNodes, n.extraNodes[count:])
+	var newValidatorNodes []*tmpnet.Node
```
cam-schultz, line 473:
```suggestion
	newValidatorNodes := make([]*tmpnet.Node, len(n.extraNodes[0:count]))
```
We need to preallocate the destination, otherwise nothing will be copied. [Example](https://go.dev/play/p/bZ4utbW6N2l)

cam-schultz: The `append` approach you adopted also works.

cam-schultz (approved):

michaelkaplan13 (approved):

iansuvak (approved):

PR #469, opened 2024-07-31
Upgradability typos

minghinmatthewlam:
## Why this should be merged
Upgradeability -> Upgradability

## How is this documented


geoff-vball (approved):

cam-schultz (approved):
According to Merriam Webster, both spellings are correct: https://www.merriam-webster.com/dictionary/upgrade

I don't have a preference one way or another, but we should be consistent across our repos.

iansuvak (approved):

PR #479, opened 2024-08-02
Add codec id to message formats

### LINKED ISSUE #470: Incorporate codec and type ID into validationID hash
**Context and scope**
The P-Chain expects the codec and type ID to be included in the `validationID` hash. See [here](https://github.com/avalanche-foundation/ACPs/pull/135#discussion_r1697013241)

cam-schultz:
## Why this should be merged
Fixes #470 
Updates the message formats in `StakingMessages` to reflect the latest iteration of ACP-77: https://github.com/avalanche-foundation/ACPs/pull/135

## How this works
- Adds `codecID` to all Warp message fields. 
- Combines `packRegisterSubnetValidatorMessage` and `packValidationInfo`. `packValidationInfo` previously did not include the `codecID` and `typeID` in the `validationID` calculation. `packRegisterSubnetValidatorMessage` packs the full message, including the `codecID` and `typeID`.

## How this was tested
forge test

## How is this documented
n/a


geoff-vball (approved):
👍

contracts/staking/StakingMessages.sol
```diff
@@ -84,46 +109,53 @@ library StakingMessages {
         pure
```
michaelkaplan13, line 109: @minghinmatthewlam pointed out that it's kind of odd for the `pack/unpackRegisterSubnetValidatorMessage` use different message types (one the request to the P-Chain, and the other the response). 

To match the updated ACP-77 spec, maybe we should go with `packRegisterSubnetValidtorMessage` and `unpackSubnetValidatorRegistrationMessage`? Not opposed to adding the request/response nomenclature otherwise either if people think that's more clear.

michaelkaplan13 (commented):
Generally makes sense to me. The [ACP-77 spec has updated](https://github.com/avalanche-foundation/ACPs/pull/135/files) these message format and defined the actual type IDs to be used, so we'll need to update them further here to be functional.

Given this PR doesn't actually match the current spec as is, I'd lean towards including those updates here, but not opposed to doing it in a follow up either.

cam-schultz:
> Generally makes sense to me. The [ACP-77 spec has updated](https://github.com/avalanche-foundation/ACPs/pull/135/files) these message format and defined the actual type IDs to be used, so we'll need to update them further here to be functional.
> 
> Given this PR doesn't actually match the current spec as is, I'd lean towards including those updates here, but not opposed to doing it in a follow up either.

Given the active iteration on the spec, I've marked this PR as draft and will bring it up to date once the spec is stable.

cam-schultz:
Closing in favor of https://github.com/ava-labs/teleporter/pull/492

Issue #436, opened 2024-07-22
Initial rewards curve implementation


cam-schultz:
**Context and scope**
Implement a static, linear rewards curve as a starting point. This will be replaced by a more robust rewards curve later.

michaelkaplan13:
Discussed offline, lets go with an initial implementation that of a flat parameterized X% inflation per year.

PR #481, opened 2024-08-06
Initial stakings rewards curve

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/436

## How this works
Just a simple linear non-compounding rewards calculation.

## How this was tested
Unit tests TODO once that PR is merged.

## How is this documented


contracts/staking/ExampleRewardCalculator.sol
```diff
@@ -0,0 +1,28 @@
...
+     */
+    function calculateReward(
+        uint256 stakeAmount,
+        uint64 startTime,
+        uint64 endTime,
+        uint256, // initialSupply
+        uint256, // endSupply
+        uint256 rewardBasisPoints
+    ) external pure returns (uint256) {
+        return stakeAmount * rewardBasisPoints * (endTime - startTime) / SECONDS_IN_YEAR / 1000;
```
cam-schultz, line 26:
```suggestion
        return stakeAmount * (rewardBasisPoints  / 1000 ) * (endTime - startTime) / SECONDS_IN_YEAR;
```
Suggest to associate the `1000` factor with the bp param to improve readability.

geoff-vball: That would equal 0, unfortunately. This is all integer arithmetic, solidity doesn't have floating point numbers, so all the division is at the end to prevent truncation.

cam-schultz: Ah, right duh

cam-schultz (approved):

contracts/staking/tests/ExamplesRewardCalculatorTests.t.sol
```diff
@@ -0,0 +1,35 @@
...
+    uint64 public constant DEFAULT_START_TIME = 1000;
+    uint64 public constant DEFAULT_END_TIME = 31537000; // a year + 1000 seonds
+    uint64 public constant DEFAULT_REWARD_BASIS_POINTS = 42;
+
+    function setUp() public {
+        exampleRewardCalculator = new ExampleRewardCalculator();
+    }
+
+    function testRewardCalculation() public view {
+        // solhint-disable-next-line func-named-parameters
```
geoff-vball, line 24: If anyone knows how to do named parameters for params ignored in the function body, let me know 😄

contracts/staking/interfaces/IRewardCalculator.sol
```diff
@@ -7,10 +7,11 @@ pragma solidity 0.8.25;
 
 interface IRewardCalculator {
     function calculateReward(
-        uint64 stakeAmount,
+        uint256 stakeAmount,
```
michaelkaplan13, line 10: JFYI, the P-Chain stake amounts are always represented as `uint64`s, so I originally thought it'd make sense to keep that here to help avoid the possibility of overflows in messages to/from the P-Chain.

contracts/staking/interfaces/IRewardCalculator.sol
```diff
@@ -7,10 +7,11 @@ pragma solidity 0.8.25;
...
         uint64 startTime,
         uint64 endTime,
         uint256 initialSupply,
-        uint256 endSupply
-    ) external returns (uint256);
+        uint256 endSupply,
+        uint64 rewardBasisPoints
```
michaelkaplan13, line 15: I think the curve definition should be the left to implementations, and not be in the interface. The interface should be kept to inputs needed to calculate a point on the curve itself.

contracts/staking/interfaces/IRewardCalculator.sol
geoff-vball: The P-Chain is going to have stake weights, which are going to be an arbitrary scaling of the actual stake amount. I think the contract could use either, but it make more sense to me to use the stake amount directly here

contracts/staking/interfaces/IRewardCalculator.sol
minghinmatthewlam: random thought, would it make sense to take in arbitrary bytes as input as well for output, to make it open to any implemetnation

michaelkaplan13: I think @cam-schultz may have brought this up previously. I don't think it makes sense because the caller and implementation would still need to agree on the expected data format (i.e. what the parameters are), and changes to that could just as easily be made as changes to the interface parameters themselves.

cam-schultz: I think in the conversation you're referencing I was suggesting providing arbitrary bytes to the rewards curve constructor, but having the rewards curve be a standalone contract from the staking manager acheives the same generality. In either case, yes, we need an agreed upon interface.

contracts/staking/ExampleRewardCalculator.sol
```diff
@@ -0,0 +1,31 @@
...
+// See the file LICENSE for licensing terms.
+
+// SPDX-License-Identifier: Ecosystem
+
+pragma solidity 0.8.25;
+
+import {IRewardCalculator} from "./interfaces/IRewardCalculator.sol";
+
+contract ExampleRewardCalculator is IRewardCalculator {
+    uint256 public constant SECONDS_IN_YEAR = 31536000;
```
bernard-avalabs, line 11: I assume we don't need to account for leap years in this initial staking reward curve.

bernard-avalabs (approved):
LGTM. Just left a comment about interest calculations on leap years.

contracts/staking/ExampleRewardCalculator.sol
geoff-vball: I think it would add a whole lot of complexity without any real tangible benefit

contracts/staking/tests/ExamplesRewardCalculatorTests.t.sol
michaelkaplan13: I think this ignore directive can be removed all together now since there are only 5 params

michaelkaplan13 (approved):
Makes sense to me. We may also want to have a default compounding rewards calculator so users are incentivized to relatively frequently end/re-stake to compound rewards.

contracts/staking/interfaces/IRewardCalculator.sol
minghinmatthewlam: I agree that the validator manager and the reward manager still need to have the same expectations of format, but take in bytes makes it so that reward managers that take into different inputs can still fall under the same interface. Right now if if a reward manager were to take a different input, it'd need to define a separate interface

geoff-vball: @minghinmatthewlam I think that having a bytes interface would actually be coutner-productive. In order for the staking contract and reward calculator to work together, they would both have to change their implementations. If they're both changing their implementations, it would be easy to change the interface between them, and I think it would be a lot clearer, where a bytes interface obfuscates the fact that the contracts might not actually agree what form the parameters should take

PR #487, opened 2024-08-08
fix doc links

cam-schultz:
## Why this should be merged
- Fixes some links in the docs that were broken when the foundry project was moved to the top level of the repo.
- Updates `contracts/README.md` to be generic to apply to all of the subdirectories, rather than Teleporter specific.
- Documents `retrySendCrossChainMessage` in the Teleporter protocol README

## How is this documented


contracts/README.md
```diff
@@ -1,16 +1,17 @@
...
 
-This directory is set up as a [Foundry](https://github.com/foundry-rs/foundry) project. Use the `scripts/install_foundry.sh` to install the correct version of the ava-labs fork of foundry. Further documentation about given contracts can be found in`src/teleporter/`.
+This directory is set up as a [Foundry](https://github.com/foundry-rs/foundry) project. Use the `scripts/install_foundry.sh` to install the correct version of the ava-labs fork of foundry. Further documentation about given contracts can be found in`contracts/teleporter/`.
 
 ## Building and Running
 - To compile the contracts run `forge build` from this directory.
 - Similarly, to run unit tests, run `forge test`.
 - See additional testing and deployment options [here](https://book.getfoundry.sh/forge/).
 
-## Deployment
```
iansuvak, line 11: Do we still want to include this in the teleporter readme? It's a common pitfall and having a **bold** warning might still be good on top of the Deployment section there

iansuvak (dismissed):
LGTM! my comment is more of a question

contracts/README.md
cam-schultz: IMO the existing warnings are sifficient. There's a note in both the top-level README, as well as a section in `contracts/teleporter/README.md`

contracts/teleporter/README.md
```diff
@@ -72,3 +73,9 @@ The `TeleporterMessenger` contract must be deployed to the same contract address
...
+
+## Resending a Message
+
+If the sending Subnet's validator set changes, then it's possible for the receiving Subnet to reject the underlying Warp message due to insufficient signing stake. For example, suppose Subnet A has 5 validators with equal stake weight who all sign a Teleporter message sent to Subnet B. 100% of Subnet A's stake has signed the message. Also suppose Subnet B requires 67% of the sending Subnet's stake to have signed a given Warp message in order for it to be accepted. Before the message can be delivered, however, 5 _more_ validators are added to Subnet A's validator set (all with the same stake weight as the original validators), meaning that the Teleporter message was signed by _only 50%_ of Subnet A's stake. Subnet B will reject this message.
+
+The Warp protocol does not support re-signing in such a scenario, but Teleporter does. `retrySendCrossChainMessage` can be called for any message that has not been acknowledged as delivered to its destination. Under the hood, this packages the Teleporter message into a brand new Warp message that is re-signed by the current validator set.
```
iansuvak, line 81: I might be misunderstanding but isn't this behavior specific to the way that signable Warp messages are being handled right now? I.e. in the future there could be messages that are signable regardless of when the validator joined. The Teleporter comment still stands and I agree that we should mention `retrySendCrossChainMessage` but I don't think that the statement about Warp protocol as a whole is true in general.

cam-schultz: Fair point. We should think of a better way to distinguish Warp messages that are sent at a point in time (with a specific validator set's aggregate signature) versus a Warp message that's constructed on demand (with the current validator set's signature). I've rephrased it to make it clear that we're talking about Warp messages sent on chain.

contracts/teleporter/README.md
```diff
@@ -65,10 +66,16 @@ Teleporter messages are delivered by calling the `receiveTeleporterMessage` func
...
+
+If the sending Subnet's validator set changes, then it's possible for the receiving Subnet to reject the underlying Warp message due to insufficient signing stake. For example, suppose Subnet A has 5 validators with equal stake weight who all sign a Teleporter message sent to Subnet B. 100% of Subnet A's stake has signed the message. Also suppose Subnet B requires 67% of the sending Subnet's stake to have signed a given Warp message in order for it to be accepted. Before the message can be delivered, however, 5 _more_ validators are added to Subnet A's validator set (all with the same stake weight as the original validators), meaning that the Teleporter message was signed by _only 50%_ of Subnet A's stake. Subnet B will reject this message.
+
+Once sent on chain, Warp messages cannot be re-signed by a new validator set in such a scenario. Teleporter, however, does support re-signing via the function `retrySendCrossChainMessage`, which can be called for any message that has not been acknowledged as delivered to its destination. Under the hood, this packages the Teleporter message into a brand new Warp message that is re-signed by the current validator set.
```
geoff-vball, line 81: I think this part is a little unclear. I think it would be more accurate to say that validators that were not online when the message was sent are not able to sign it, because we don't bootstrap validators.

geoff-vball (commented):
Just one comment, otherwise LGTM

iansuvak (approved):

contracts/teleporter/README.md
iansuvak: To me, this reads well and is a clear explanation of the scenario. I am fine with adding Geoff's summary up top and leaving the detailed explanation in as an example scenario.

cam-schultz: I tried a handful of variations that mentioned bootstrapping, but IMO they were all less clear than this description of "if the validator set changes, then signature verification may fail". This example isn't meant to be exhaustive, just to illustrate how Teleporter handles such a scenario.

geoff-vball (approved):

geoff-vball (approved):

contracts/teleporter/README.md
```diff
@@ -7,8 +7,9 @@
   - [Fees](#fees)
     - [Message Receipts and Fee Redemption](#message-receipts-and-fee-redemption)
   - [Required Interface](#required-interface)
-  - [Teleporter Contract Deployment](#teleporter-contract-deployment)
   - [Message Delivery and Execution](#message-delivery-and-execution)
+  - [Resending a Message](#resending-a-message)
```
michaelkaplan13, line 11: Nit: Github provides an outline/table of contents by default. Can we remove the manual one while we're here so it doesn't fall out of date in the future?

image

cam-schultz: 100%

michaelkaplan13 (approved):

geoff-vball (approved):

PR #491, opened 2024-08-14
Poa staking

### LINKED ISSUE #449: Extensions for Proof-of-Authority support
**Context and scope**
Subnet validators are registered on the P-Chain in the same way for proof of authority subnets as for proof of stake subnets. A variant of the Staking Manager contract could support PoA validator registration by requiring all validator management operations be conducted by or signed by an admin address. No tokens would need to be locked; instead the admin sets the weight arbitrarily.

**Design**
The PoA variant will be implemented as an upgradeable contract that inherits from `StakingManager`. Migrating to PoS would involve upgrading to a different contract that also inherits from `StakingManager`; in doing so, the contract state pertaining to the registered validators can be migrated cleanly.

**Open questions**
We will need to exclude migrated PoA validators from being able to claim staking rewards. There are many ways to do so, here are a few options:
- On migration, set the PoA validator weights to 0 in the staking contract. This does not affect the weights registered on the P-Chain and used for consensus
- When PoA validator registration is complete, set it's weight to zero. This saves us from having to do so as part of migration
- Mark the block timestamp at which migration occured, and exclude any validators that began before that time from claiming staking rewards.

minghinmatthewlam:
## Why this should be merged
Fixes #449 

## How this works
Adds a PoA contract that requires an initial owner address
- initialization of registration can only be called by the owner
- There is no associated token, so the weight is set by the initialization call
- owner of the validator is set to the PoA owner, so after migraiton to PoS only the previous PoA owner can initialize ending validation of previous existing validators

Refactors staking related logic from base contract to only apply to PoS
Renames `StakingManager` -> `ValidatorManager` same for anything that is not specific to PoS like the p-chain validator messages.
Abstracts the unit tests to conform to same layout, separate PoS unit tests

TODO: https://github.com/ava-labs/teleporter/issues/496 as followup

## How this was tested
unit tests

## How is this documented
readme


cam-schultz (commented):
I left some comments suggesting some interface changes, but the implementation LGTM.

Once the E2E test framework is ready, we should add tests exercising the PoA -> PoS migration.

contracts/staking/PoAStakingManager.sol
```diff
@@ -0,0 +1,59 @@
...
+        return _initializeValidatorRegistration(nodeID, stakeAmount, registrationExpiry, signature);
+    }
+    // solhint-enable ordering
+
+    function _lock(uint256 value) internal virtual override returns (uint256) {
+        return value;
+    }
+
+    // solhint-disable-next-line no-empty-blocks
+    function _unlock(uint256 value, address to) internal virtual override {}
```
geoff-vball, line 58: We should probably move these to a PoS-only contract

contracts/staking/ERC20TokenStakingManager.sol
```diff
@@ -50,18 +56,35 @@ contract ERC20TokenStakingManager is Initializable, StakingManager, IERC20TokenS
     }
 
     function initialize(
-        StakingManagerSettings calldata settings,
+        ValidatorManagerSettings calldata settings,
+        uint256 minimumStakeAmount,
+        uint256 maximumStakeAmount,
+        uint64 minimumStakeDuration,
+        IRewardCalculator rewardCalculator,
```
cam-schultz, line 63: It may be cleaner to combine these into a `PoSValidatorManagerSettings` in `IPoSValidatorManager.sol`. That way all PoS validator managers can use the same type to initialize.

cam-schultz (approved):
One small suggestion to add a `PoSValidatorManagerSettings` type, otherwise LGTM

PR #495, opened 2024-08-19
Wait for Teleporter message delivery test utility

cam-schultz:
## Why this should be merged
Unblocks https://github.com/ava-labs/awm-relayer/issues/437

## How this works
Adds a utility function that blocks until `TeleporterMessenger.messageReceived` returns true for the given Teleporter message ID, or the timeout expires.

## How this was tested
CI

## How is this documented
N/A

### DIFF
--- a/tests/utils/utils.go
+++ b/tests/utils/utils.go
@@ -1207,3 +1207,40 @@ func InstantiateGenesisTemplate(
 
 	return subnetGenesisFile.Name()
 }
+
+//
+// Teleporter message utils
+//
+
+// Blocks until the given teleporter message is delivered to the specified TeleporterMessenger
+// before the timeout, or if an error occurred.
+func WaitTeleporterMessageDelivered(
+	ctx context.Context,
+	teleporterMessenger *teleportermessenger.TeleporterMessenger,
+	teleporterMessageID ids.ID,
+) error {
+	cctx, cancel := context.WithTimeout(ctx, 20*time.Second)
+	defer cancel()
+
+	queryTicker := time.NewTicker(200 * time.Millisecond)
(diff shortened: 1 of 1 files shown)


tests/utils/utils.go
```diff
@@ -1207,3 +1207,40 @@ func InstantiateGenesisTemplate(
...
+	queryTicker := time.NewTicker(200 * time.Millisecond)
+	defer queryTicker.Stop()
+	for {
+		delivered, err := teleporterMessenger.MessageReceived(
+			&bind.CallOpts{}, teleporterMessageID,
+		)
+		if err != nil {
+			return err
+		}
+
```
geoff-vball, line 1234: super nit: I prefer no newline when checking return values from a function

geoff-vball (approved):

iansuvak (approved):

minghinmatthewlam (approved):

minghinmatthewlam:
If this utility isn't used in `teleporter`, not sure if this repo should need to hold utilities used by other repos dependent on teleporter, and instead live in `awm-relayer` unless more dependencies end up needing it.

cam-schultz:
> If this utility isn't used in `teleporter`, not sure if this repo should need to hold utilities used by other repos dependent on teleporter, and instead live in `awm-relayer` unless more dependencies end up needing it.

OTOH, I'm sure we could find plenty of places in the Teleporter E2E tests that could use this utility instead of waiting for a transaction to be mined. It's made a little easier to do so here since we actually have the tx hash directly from the relayer utility. All that said, I think this utility is tightly scoped enough to Teleporter that it makes sense for it to live here.

PR #498, opened 2024-08-21
Remove tesnet flows

minghinmatthewlam:
## Why this should be merged
Removes testnet e2e flows that we don't use and is extra burden to maintain. Mostly will rely on local network e2e tests, only main difference with testing on testnet is the difference in the networks, so confirming in manual deployments of Teleporter's interaction with Warp and network validators should cover

## How this works
still leaves e2e test that allow for non local networks being passed. Might be useful when we want to manually set up testnet testing.

## How is this documented


geoff-vball (commented):
LGTM. I think the upload tmpnet files step just needs to be fixed

geoff-vball (dismissed):

iansuvak (approved):

feuGeneA (approved):

bernard-avalabs (approved):

tests/network.go
cam-schultz, line 1: We should remove `interfaces/network.go`, `interfaces/local_network.go`, and have all e2e tests take a concrete `LocalNetwork` parameter instead. There's no need for either interface, since there's only one implementation.

minghinmatthewlam:
deferring for now to fully handle removing https://github.com/ava-labs/teleporter/pull/498#discussion_r1727250386

tests/network.go
geoff-vball: Done 👍

iansuvak (dismissed):
I like the simplification with removal of interfaces and resulting shrinking of the utils file. 

FWIW I didn't mind the earlier commit you had with `n` instead of `localnetwork` either but this works too.

tests/network/network_utils.go
cam-schultz, line 1: Can we move the `LocalNetwork` methods in this file to `tests/network/network.go` with all the others?

cam-schultz (commented):
Overall lgtm, just one request to put the `LocalNetwork` methods all in the same place.

cam-schultz (approved):

cam-schultz (dismissed):

geoff-vball (approved):

Issue #480, opened 2024-08-02
StakingManager initialStake & ValidatorChurnPeriod


minghinmatthewlam:
**Context and scope**
Currently in the WIP progress [branch](https://github.com/ava-labs/teleporter/pulls) we have a concept to check for validator churn on updates, to avoid too much churn in a short period of time. Some parts of the initialization steps for the validator set of the staking manager have changed, how do we initialize the churn period, and also update the stake each period? 

**Discussion and alternatives**
Check if other protocols like Ethereum have a queue for validators to be added/removed.

PR #501, opened 2024-08-22
Moving uptime to PoS

### LINKED ISSUE #496: refactor staking info out of ValidatorManager
**Context and scope**
Follow up from https://github.com/ava-labs/teleporter/pull/491, which starts the initial inheritance hierarchy between the contracts. `ValidatorManager` should only keep state/logic that applies to both PoA and PoS validator managers, in order to support migration from PoA -> PoS #488 related. 

At least the `uptime` and `rewarded` infos on the `ValidatorInfo` is currently saved in the base contract's state, and only applies to PoS. 

Could have separate state variables that keep track of all stake related info for a validation ID. Will likely live in `PoSValidatorManager` as it does not apply to PoA, and applies to both native or erc20 token staking.

minghinmatthewlam:
## Why this should be merged
Fixes #496 

## How this works
Moves PoS logic out of `ValidatorManager` for uptime. Validator churn tracking will be moved in #480 

Verifies the uptime proof and sets the uptime if valid, then calls the underlying initialize end validation.

## How this was tested
Unit tests for uptime

## How is this documented


contracts/staking/PoSValidatorManager.sol
```diff
@@ -21,6 +21,7 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         uint256 _maximumStakeAmount;
         uint64 _minimumStakeDuration;
         IRewardCalculator _rewardCalculator;
+        mapping(bytes32 validationID => uint64) _uptimes;
```
bernard-avalabs, line 24: small nit: perhaps validatorUptimes would be more descriptive.

bernard-avalabs (commented):
Generally makes sense to me. Looks like there are a few Slither alerts to cleanup.

geoff-vball (approved):
Generally LGTM 👍

contracts/staking/ValidatorMessages.sol
```diff
@@ -7,8 +7,8 @@ pragma solidity 0.8.25;
 library ValidatorMessages {
     // The information that uniquely identifies a subnet validation period.
     // The validationID is the SHA-256 hash of the concatenation of the CODEC_ID,
-    // REGISTER_SUBNET_VALIDATOR_MESSAGE_TYPE_ID, and the concatenated ValidationInfo fields.
-    struct ValidationInfo {
+    // REGISTER_SUBNET_VALIDATOR_MESSAGE_TYPE_ID, and the concatenated ValidationPeriod fields.
+    struct ValidationPeriod {
```
cam-schultz, line 11: I like this name. We should adopt this nomenclature whenever we're describing the entity specified by `validationID`.

cam-schultz (approved):

PR #502, opened 2024-08-23
staking: small README improvements

feuGeneA:
## Why this should be merged
- ensures all hyperlinks are applied to the first appearance of a term in the doc
- adds some hyperlinks that were missing (things we can link to but weren't yet)
- corrects `SetValidatorWeightTx` to `SetSubnetValidatorWeightTx`
- changes some more instances of "staking manager" to "validator manager"
- adds to message format descriptions a mention of where the message is issued from
- makes a few tiny grammar/style improvements

### DIFF
--- a/contracts/staking/README.md
+++ b/contracts/staking/README.md
@@ -20,7 +20,7 @@ If the Subnet has no validators registered on the P-Chain, then it will not be a
 
 Validator registration is initiated with a call to `initializeValidatorRegistration` on the concrete Validator Manager contract. The Validator Manager contract constructs a [`RegisterSubnetValidatorMessage`](#registersubnetvalidatormessage) Warp message to be sent to the P-Chain. Each validator registration request includes all of the information needed to identify the validator and its stake weight, as well as an `expiry` timestamp before which the `RegisterSubnetValidatorMessage` must be delivered to the P-Chain.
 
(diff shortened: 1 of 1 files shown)


contracts/staking/README.md
```diff
@@ -30,21 +30,21 @@ In the case of a missed expiry, the `RegisterSubnetValidatorTx` will result in a
...
 
 ## Warp Message Format Reference
 
 ### `RegisterSubnetValidatorMessage`
 
-Description: Used to register a Subnet validator on the P-Chain
+Description: Issued by the Validator Manager contract in order to register a Subnet validator
 
-Signed by: Subnet
+Signed by: Subnet validators
```
cam-schultz, line 47: nit: It's a bit odd that we specify the subnet validators here, but not the P-Chain validators in other sections. If we're going to be specific, the other sections should _also_ be `Signed by the subnet validators, via their P-Chain VM` due to the self-signing optimization for Warp messages from the primary network. 

IMO it's clearer to leave this as is, and omit validators altogether.

cam-schultz (approved):
One nit, otherwise LGTM

geoff-vball (approved):

iansuvak (approved):

bernard-avalabs (approved):

PR #503, opened 2024-08-23
Simplify completeEndValidation

cam-schultz:
## Why this should be merged
Makes a handful of changes to `completeEndValidation`:
- Removes `SetSubnetValidatorWeight` as a valid Warp message type to be consumed. When a validator's weight is set to. 0 on the P-Chain, the P-Chain signs a `SubnetValidatorRegistration` message with `registered=false`. It does _not_ sign a `SetSubnetValidatorWeight` message. This message type is only signed by the P-Chain to attest to the weight of an active validation.
- Removes the requirement that only the validator owner can call this function. As long as a valid Warp message from the P-Chain attesting to the removal of the validator is provided, then it should also be removed from the staking contract. 
- Adds checks that the Warp message is from the pre-determined address on the P-Chain
- Removes `ValidatorStatus.Active` as a valid status. This was to support cases in which a validator removes itself from validating a subnet unilaterally on the P-Chain via `ExitSubnetValidatorTx`. This does _not_, however, remove the validator's weight from the validator set, but merely treats the validator as inactive. The validator may become active again in the future by increasing its balance. The only valid way to change the validator's weight is through a Warp message signed by the validator manager contract. As such, the only expected status here is `ValidatorStatus.PendingRemoval`
- Adds _defensive_ handling of other validator statuses that also remove the validator from the contract's active validator state. In these cases, the status is set to  `Invalidated` rather than `Completed`.

## How this works
See above

## How this was tested
CI

## How is this documented
N/A


contracts/staking/ValidatorManager.sol
```diff
@@ -53,7 +53,7 @@ abstract contract ValidatorManager is
 
     // solhint-disable ordering
     function _getValidatorManagerStorage()
-        private
+        internal
```
minghinmatthewlam, line 56: would be wary of making this `internal` as it uses assembly to assign slot of value `$`, the same for other contracts

contracts/staking/ValidatorManager.sol
```diff
@@ -275,49 +275,44 @@ abstract contract ValidatorManager is
...
         // Get the Warp message.
         (WarpMessage memory warpMessage, bool valid) =
             WARP_MESSENGER.getVerifiedWarpMessage(messageIndex);
         require(valid, "ValidatorManager: Invalid warp message");
+        require(
+            warpMessage.sourceChainID == $._pChainBlockchainID,
+            "ValidatorManager: Invalid source chain ID"
+        );
+        require(
+            warpMessage.originSenderAddress == address(0),
```
minghinmatthewlam, line 291: we should refactor into a modifier like `onlyPChainMessage`

cam-schultz: I added a helper `_getPChainWarpMessage` in this [PR](https://github.com/ava-labs/teleporter/pull/504/)

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
-        if (
-            validator.status == ValidatorStatus.PendingRemoved
-                || validator.status == ValidatorStatus.Active
-        ) {
-            // Remove the validator from the active validators mapping.
-            delete $._activeValidators[validator.nodeID];
+        if (validator.status == ValidatorStatus.PendingRemoved) {
             endStatus = ValidatorStatus.Completed;
         } else {
             endStatus = ValidatorStatus.Invalidated;
```
feuGeneA, line 302: If this method is called twice with the same `messageIndex`, will this transition from `PendingRemoved` to `Completed` the first time, and then `Completed` to `Invalidated` the second time? Should there be another `if` branch for `Completed` that's just a no-op and return?

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -105,9 +105,6 @@ interface IValidatorManager {
      * {registrationExpiry} being reached.
      * @param messageIndex The index of the Warp message to be received providing the proof the validation is not active
      * and never will be active on the P-Chain.
-     * @param setWeightMessageType Whether or not the message type is a SetValidatorWeight message, or a
-     * SubnetValidatorRegistration message (with valid set to false). Both message types are valid for ending
```
feuGeneA, line 109: I think this doc block should still mention the expectation that the message is a `SubnetValidatorRegistration` message, I guess in the `@param` description

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
-            require(weight == 0, "ValidatorManager: Weight must be 0");
-        } else {
-            bool validRegistration;
-            (validationID, validRegistration) =
-                ValidatorMessages.unpackSubnetValidatorRegistrationMessage(warpMessage.payload);
-            require(!validRegistration, "ValidatorManager: Registration still valid");
-        }
+        (bytes32 validationID, bool validRegistration) =
+            ValidatorMessages.unpackSubnetValidatorRegistrationMessage(warpMessage.payload);
+        require(!validRegistration, "ValidatorManager: Registration still valid");
```
minghinmatthewlam, line 291: nit: can we lower case the error messages after the contract name

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
 
-        // Note: The validation status is not necessarily PendingRemoved here. It could be PendingAdded if the
-        // registration was never successfully registered on the P-Chain, or it could be Active if the validator
-        // removed themselves directly on the P-Chain without going through the contract.
+        // The validation status is PendingRemoved if validator removal was initiated with a call to initiateEndValidation.
```
minghinmatthewlam, line 293:
```suggestion
        // The validation status is PendingRemoved if validator removal was initiated with a call to {initiateEndValidation}.
```

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
+        // The validation status is PendingAdded if the validator was never registered on the P-Chain.
+        // All other statuses are unexpected, but we defensively allow them to specify the validation period as invalidated,
```
minghinmatthewlam, line 295: i'm not sure we should be allowing other validation period statuses, I would expect and think we should check that the only valid statuses are pending removed or pending added

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
-            delete $._activeValidators[validator.nodeID];
+        if (validator.status == ValidatorStatus.PendingRemoved) {
             endStatus = ValidatorStatus.Completed;
         } else {
             endStatus = ValidatorStatus.Invalidated;
         }
+        // Remove the validator from the active validators mapping.
+        delete $._activeValidators[validator.nodeID];
 
+        // Update the validator status.
```
minghinmatthewlam, line 307: in the validation period end event, might want to include the new status

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
-        ) {
-            // Remove the validator from the active validators mapping.
-            delete $._activeValidators[validator.nodeID];
+        if (validator.status == ValidatorStatus.PendingRemoved) {
             endStatus = ValidatorStatus.Completed;
         } else {
             endStatus = ValidatorStatus.Invalidated;
         }
+        // Remove the validator from the active validators mapping.
+        delete $._activeValidators[validator.nodeID];
```
minghinmatthewlam, line 305: we have pending registered messages saved, do we need pending end messages too?

contracts/staking/ValidatorManager.sol
```diff
@@ -269,49 +269,43 @@ abstract contract ValidatorManager is
...
         // Get the Warp message.
         (WarpMessage memory warpMessage, bool valid) =
             WARP_MESSENGER.getVerifiedWarpMessage(messageIndex);
         require(valid, "ValidatorManager: Invalid warp message");
+        require(
+            warpMessage.sourceChainID == $._pChainBlockchainID,
+            "ValidatorManager: Invalid source chain ID"
+        );
+        require(
+            warpMessage.originSenderAddress == address(0),
```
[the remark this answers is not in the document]
cam-schultz: All Warp messages from the P-Chain are `AddressedCall` messages with the `sourceAddress` set to an empty byte array (i.e. the zero address): https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#proof-of-subnet-validator-set-change

I added a `_getPChainWarpMessage` helper in this [PR](https://github.com/ava-labs/teleporter/pull/504); I'll add the constant symbol there.

contracts/staking/ValidatorManager.sol
cam-schultz: Added in https://github.com/ava-labs/teleporter/pull/504

contracts/staking/ValidatorManager.sol
cam-schultz: Good call. Added the validaor status to the event

contracts/staking/ValidatorManager.sol
cam-schultz: Agreed. Simplified this by requiring onlly pending removed or pending added.

contracts/staking/ValidatorManager.sol
cam-schultz: Updated here and elsewhere in `contracts/staking`

contracts/staking/ValidatorManager.sol
cam-schultz: Simplified this by allowing only pending removed or pending added.

contracts/staking/interfaces/IValidatorManager.sol
cam-schultz: I think the Warp message format is an implementation detail that doesn't need to be mentioned in the interface. For consistency, it's also not mentioned in the `completeValidatorRegistration` description.

contracts/staking/ValidatorManager.sol
geoff-vball: Did we make a decision on what we wanted to do here? The churn-tracker PR also makes this function internal in order to access `ValidatorManager` state, but it's only to access a validator's weight. We might be okay adding a separate getter for the validator state.

contracts/staking/ValidatorManager.sol
```diff
@@ -269,57 +269,56 @@ abstract contract ValidatorManager is
...
-            (validationID,, weight) =
-                ValidatorMessages.unpackSetSubnetValidatorWeightMessage(warpMessage.payload);
-            require(weight == 0, "ValidatorManager: Weight must be 0");
-        } else {
-            bool validRegistration;
-            (validationID, validRegistration) =
-                ValidatorMessages.unpackSubnetValidatorRegistrationMessage(warpMessage.payload);
-            require(!validRegistration, "ValidatorManager: Registration still valid");
-        }
+        require(valid, "ValidatorManager: invalid warp message");
```
geoff-vball, line 279: nit: Since we have `validRegistration` below, can we rename `valid` here to `validWarpMessage` or something similar?

feuGeneA (commented):
Reminder to remove `SetSubnetValidatorWeight` from the README as well

contracts/staking/ValidatorManager.sol
cam-schultz: Going to leave this as is because it'll be encapsulated in the aforementioned [P-Chain helper](https://github.com/ava-labs/teleporter/pull/503#discussion_r1735284703) shortly

geoff-vball (approved):

PR #504, opened 2024-08-27
Delegation PoC

### LINKED ISSUE #441: Allow users to Delegate Stake
**Context and scope**
The staking contract should have an external function for users to delegate tokens for a validator to stake for them.

This will require sending a `SetSubnetValidatorWeightTx` message to the p-chain to update the validator's stake weight.

**Open questions**
Parameters on how rewards are shared.

cam-schultz:
## Why this should be merged
Partially fixes #441 

## How this works
- Adds initialize/resend/complete methods for adding and removing delegators
- Adds pack/unpack methods for `SubnetValidatorWeightUpdate` Warp messages
- Unit tests for the above
- Various cleanup items in the existing implementations (especially regarding nonce + weight management)

## How this was tested
CI, new unit tests

## How is this documented
N/A


contracts/staking/ERC20TokenStakingManager.sol
```diff
@@ -77,16 +77,37 @@ contract ERC20TokenStakingManager is
...
         uint64 weight = _processStake(stakeAmount);
         return _initializeValidatorRegistration(nodeID, weight, registrationExpiry, blsPublicKey);
     }
 
+    /**
+     * @notice Begins the delegator registration process. Locks the configured ERC20 in the contract as the delegated stake.
+     * @param validationID The ID of the validation period being delegated to.
+     * @param delegationAmount The amount to be delegated.
+     */
+    function initializeDelegatorRegistration(
```
minghinmatthewlam, line 102: did we want to offer delegation by default? Thought it was going to be an extension

contracts/staking/ERC20TokenStakingManager.sol
```diff
@@ -77,16 +77,37 @@ contract ERC20TokenStakingManager is
...
+        bytes32 validationID,
+        uint256 delegationAmount
+    ) external {
+        return _initializeDelegatorRegistration(
+            validationID, _msgSender(), valueToWeight(_lock(delegationAmount))
```
minghinmatthewlam, line 107: boh erc20 and native implementations call lock then convert to weight for stake amount, the internal `_initializeDelegatorRegistration` can just take in the stake amount as parameter

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        );
+        $._pendingRegisterDelegatorMessages[validationID][delegator] = setValidatorWeightPayload;
+        bytes32 messageID = WARP_MESSENGER.sendWarpMessage(setValidatorWeightPayload);
+
+        // Store the delegator information
+        $._delegatorStakes[validationID][delegator] = Delegator({
+            weight: weight,
+            startedAt: uint64(block.timestamp),
+            endedAt: 0,
+            status: ValidatorStatus.PendingAdded
```
minghinmatthewlam, line 178: delegations are similar to validations, but seems strange to use the validator status enum

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        require(
+            $._pendingRegisterDelegatorMessages[validationID][delegator].length > 0
+                && $._delegatorStakes[validationID][delegator].status == ValidatorStatus.PendingAdded,
+            "PoSValidatorManager: Delegator registration not pending"
+        );
+        delete $._pendingRegisterDelegatorMessages[validationID][delegator];
+
+        // Update the validator weight and delegator status
+        Validator memory validator = _getValidator(validationID);
+        require(
```
minghinmatthewlam, line 212: I don't think this check is valid if multiple delegations are pending for the same validation ID, and we receive the Warp message for the last delegation

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            delegator.status == ValidatorStatus.Active, "ValidatorManager: Delegator not active"
+        );
+        delegator.status = ValidatorStatus.PendingRemoved;
+        delegator.endedAt = uint64(block.timestamp);
+
+        $._delegatorStakes[validationID][_msgSender()] = delegator;
+
+        Validator memory validator = _getValidator(validationID);
+        require(validator.weight >= delegator.weight, "PoSValidatorManager: Invalid weight");
```
minghinmatthewlam, line 244: thinking if this should be `>`. Doesn't seem like a validator with zero weight would make sense.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+
+    function completeEndDelegation(uint32 messageIndex, address delegator) external {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        // Unpack the Warp message
+        WarpMessage memory warpMessage = _getPChainWarpMessage(messageIndex);
+        (bytes32 validationID, uint64 nonce, uint64 weight) =
+            ValidatorMessages.unpackSubnetValidatorWeightUpdateMessage(warpMessage.payload);
+        require(
+            $._pendingEndDelegatorMessages[validationID][delegator].length > 0
```
minghinmatthewlam, line 275: we perform these checks multiple times, shoudl use helper functions

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+                && $._delegatorStakes[validationID][delegator].status == ValidatorStatus.PendingRemoved,
+            "PoSValidatorManager: Delegator removal not pending"
+        );
+        delete $._pendingEndDelegatorMessages[validationID][delegator];
+
+        // Update the validator weight and delegator status
+        Validator memory validator = _getValidator(validationID);
+        require(
```
minghinmatthewlam, line 283: same here, don't think this will work for multiple delegation ends to the same validation ID

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            validator.weight - $._delegatorStakes[validationID][delegator].weight == weight,
+            "PoSValidatorManager: Invalid weight"
+        );
+        require(validator.messageNonce >= nonce, "PoSValidatorManager: Invalid nonce");
```
minghinmatthewlam, line 287: for nonce check we should make sure if we've seen a nonce greater than the received message, we don't double the effect, since the newer message would already account for older message

cam-schultz: Since the nonce is set by the validator manager, it should be safe to change this to `==`

contracts/staking/PoSValidatorManager.sol
cam-schultz: Good point. Since we have the nonce, we should be able to remove this check altogether.

contracts/staking/PoSValidatorManager.sol
cam-schultz: This will be a potentially worse UI if delegators are added at a fast rate, since only the most recent weight update message is eligible to be delivered.

cam-schultz: Actually, I think the implementation as is (with the removed weight check above) is sufficient. By checking `validator.messageNonce >= nonce`, we guarantee that we're only receiving Warp messages pertaining to `initialize` calls originating from the validator manager. The only state change in `completeInitializeDelegation` and `completeEndDelegation` use information provided in the Warp message, so replay protection is implicitly provided by the P-Chain.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +140,155 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        // Ensure the validation period is active
+        Validator memory validator = _getValidator(validationID);
+        require(
+            validator.status == ValidatorStatus.Active, "ValidatorManager: Validator not active"
+        );
+
+        // Ensure the delegator is not already registered
+        require(
```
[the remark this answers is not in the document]
cam-schultz: Let's defer this to a followup. 

I think the safest way to this would be to come up with a unique delegation ID that takes into account the validatorID, the delegator address, and maybe the starting timestamp for uniquenes. Then add a separate delegation owner field, rather than using the delegator address for that purpose. That would also be more general, since the address that adds the delegation would not need to be the same address that claims rewards.

contracts/staking/ERC20TokenStakingManager.sol
cam-schultz: Generally speaking, I prefer to make features optional rather than add complexity to the inheritance graph by adding extensions to support those features. Inheritance is best used as an abstraction tool to reduce shared code or provide polymorphism. 

Regardless, for now I'm going to keep this as part of the PoSValidatorManager for the PoC, and we can revisit inheritance later.

contracts/staking/PoSValidatorManager.sol
cam-schultz: *Sigh* this is a bit more complicated actually. If we only check `validator.messageNonce >= nonce`, then if a higher nonce is delivered before a lower nonce, then delivering the lower nonce will revert the weight change from the higher nonce. This is what you originally called out, but I needed to think through it a bit myself 😅 . I'll play around with it some more.

contracts/staking/ValidatorMessages.sol
```diff
@@ -367,6 +367,105 @@ library ValidatorMessages {
...
+        for (uint256 i; i < 32; ++i) {
+            res[i + 6] = bytes1(uint8(uint256(validationID >> (8 * (31 - i)))));
+        }
+        // Pack the nonce.
+        for (uint256 i; i < 8; ++i) {
+            res[i + 38] = bytes1(uint8(nonce >> (8 * (7 - i))));
+        }
+        // Pack the weight.
+        for (uint256 i; i < 8; ++i) {
+            res[i + 46] = bytes1(uint8(weight >> (8 * (7 - i))));
```
bernard-avalabs, line 412: I understand that we have always used this way of packing a bytes array. But it still looks error prone to me compared to an approach that doesn't require hard coding magic number indices for each field.

bernard-avalabs (commented):
Generally makes sense to me. Left a few comments.

contracts/staking/ValidatorMessages.sol
cam-schultz: Totally agreed. This very likely won't be the byte packing approach we go with, as it's more complex and less gas efficient than using, say ,`abi.encodePacked`. I went with this approach here for consistency, and plan to update all the packing utilities in one shot in a future change.

contracts/staking/PoSValidatorManager.sol
cam-schultz: I've implemented nonce management that updates the contract's stored validator weight on `initializeRegisterDelegation`/`initializeEndDelegation`. This allows us to elide any special handling on delivery order for the corresponding `complete` functions. This is safe because we don't need the validator's total weight for anything when we `completeEndDelegation`, since the delegation rewards are entirely determined by the delegator's weight.

contracts/staking/PoSValidatorManager.sol
cam-schultz: Added a distinct `DelegatorStatus` enum that removes `Invalidated` as an option.

contracts/staking/PoSValidatorManager.sol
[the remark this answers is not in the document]
iansuvak: > I'm not sure if this would make a big UX difference for delegators. Couldn't they just start a separate delegation with the additional stake in most cases?

In the current case they could but would have to use a separate address given this check. I do like the idea of the delegation owner field being tracked separately from the delegator address

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +146,170 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            ValidatorMessages.unpackSubnetValidatorWeightUpdateMessage(warpMessage.payload);
+        _checkPendingEndDelegatorMessage(validationID, delegator);
+        delete $._pendingEndDelegatorMessages[validationID][delegator];
+
+        // Update the validator weight and delegator status
+        Validator memory validator = _getValidator(validationID);
+        require(validator.messageNonce >= nonce, "PoSValidatorManager: invalid nonce");
+        $._delegatorStakes[validationID][delegator].status = DelegatorStatus.Completed;
+
+        emit DelegationEnded(validationID, delegator, nonce);
```
geoff-vball, line 289: We should be checking uptime, giving out rewards, and returning stake in here I think?

For uptime, I think anyone should be able to provide an uptime proof for a validator. Since we track the seconds of uptime, a proof can be provided basically at any point, and if the highest reported `uptimeSeconds` is at least 80% of the time from the validation period starting and the MIN(validation end, current time), then we can disburse rewards.

contracts/staking/ValidatorManager.sol
```diff
@@ -321,13 +306,20 @@ abstract contract ValidatorManager is
         emit ValidationPeriodEnded(validationID, validator.status);
     }
 
+    /**
+     * @notice Returns the validator's weight. This weight is not guaranteed to be known by the P-Chain
+     */
+    function getWeight(bytes32 validationID) external view returns (uint64) {
+        return _getValidator(validationID).weight;
+    }
```
geoff-vball, line 314: Should we require that this returns a non-zero value? Or leave it up to the caller?

contracts/staking/interfaces/IPoSValidatorManager.sol
```diff
@@ -24,6 +39,65 @@ interface IPoSValidatorManager is IValidatorManager {
...
+     * @notice Event emitted when a delegator registration is initiated
+     * @param validationID The ID of the validation period
+     * @param delegator The address of the delegator
+     * @param delegatorWeight The weight of the delegator
+     * @param validatorWeight The updated validator weight that is sent to the P-Chain
+     * @param nonce The message nonce used to update the validator weight
+     */
+    event DelegatorAdded(
+        bytes32 indexed validationID,
+        bytes32 indexed setWeightMessageID,
```
geoff-vball, line 52: This is missing in the natspec docs above

contracts/staking/interfaces/IPoSValidatorManager.sol
```diff
@@ -39,4 +113,43 @@ interface IPoSValidatorManager is IValidatorManager {
...
+     */
+    function initializeEndDelegation(bytes32 validationID) external;
+
+    /**
+     * @notice Resubmits a delegator end message to be sent to the P-Chain.
+     * Only necessary if the original message can't be delivered due to validator churn.
+     * @param validationID The ID of the validation period being ended.
+     * @param delegator The address of the delegator being removed.
+     */
+    function resendEndDelegation(bytes32 validationID, address delegator) external;
```
geoff-vball, line 146: Technically our resends for starting or ending a delegation would only need to resend the current validator total weight with the highest nonce we've used so far. We don't need to store any of these messages, we just need to send our current state.

geoff-vball (commented):
So far I've just gone over the implementations. I'll look over the tests next week :)

contracts/staking/ValidatorManager.sol
geoff-vball: If we don't want to require it here, we should leave a comment that a return value of 0 means the validation ID doesn't exist.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +146,170 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        WARP_MESSENGER.sendWarpMessage($._pendingRegisterDelegatorMessages[validationID][delegator]);
+    }
+
+    function completeDelegatorRegistration(uint32 messageIndex, address delegator) external {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        // Unpack the Warp message
+        WarpMessage memory warpMessage = _getPChainWarpMessage(messageIndex);
+        (bytes32 validationID, uint64 nonce,) =
+            ValidatorMessages.unpackSubnetValidatorWeightUpdateMessage(warpMessage.payload);
```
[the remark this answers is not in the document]
cam-schultz: Going to defer this to a followup, since it is an optimization. Definitely think we should support this though.

contracts/staking/interfaces/IPoSValidatorManager.sol
cam-schultz: I'm going to leave this PR storing the pending messages in the contract state, both for consistency, and because there's a potential gas tradeoff to reconstructing the Warp message when calling the `resend` methods. I'll create an issue to quantify that and proceed accordingly.

contracts/staking/ValidatorManager.sol
cam-schultz: Updated the comment to clarify the behavior for unregistered validation IDs

contracts/staking/PoSValidatorManager.sol
cam-schultz: I'm going to defer uptime/rewards handling since I think there are still some open design questions here.

cam-schultz: I implemented uptime handling here: https://github.com/ava-labs/teleporter/pull/520

Happy to merge that into this branch and include it in this PR if you prefer.

contracts/staking/interfaces/IPoSValidatorManager.sol
```diff
@@ -16,6 +24,15 @@ struct PoSValidatorManagerSettings {
     IRewardCalculator rewardCalculator;
 }
 
+struct Delegator {
+    uint64 weight;
+    uint64 startedAt;
+    uint64 endedAt;
+    uint64 startingNonce;
+    uint64 endingNonce;
+    DelegatorStatus status;
```
michaelkaplan13, line 33: Nit: To be consistent with the `Validator` struct, could we put `status` first?

contracts/staking/PoSValidatorManager.sol
```diff
@@ -17,10 +20,23 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        /// @notice The minimum amount of stake required to be a validator.
         uint256 _minimumStakeAmount;
+        /// @notice The maximum amount of stake allowed to be a validator.
         uint256 _maximumStakeAmount;
+        /// @notice The minimum amount of time a validator must be staked for.
         uint64 _minimumStakeDuration;
+        /// @notice The reward calculator for this validator manager.
         IRewardCalculator _rewardCalculator;
+        /// @notice Maps the validationID to a mapping of delegator address to delegator information.
+        mapping(bytes32 validationID => mapping(address delegator => Delegator)) _delegatorStakes;
```
michaelkaplan13, line 32: I think it may make more sense here to create a notion of a `delegationID` (similar to `validationID`) to key these maps on.

Theoretically, an address should be able to delegate to the same validation ID multiple times.

The `delegationID` could be a hash of the (`validationID`,`nonce`), where the nonce is the nonce used in the `SetSubnetValidatorWeight` message that added the delegator.

contracts/staking/interfaces/IPoSValidatorManager.sol
geoff-vball: I imagine the storage will cost more than the computation, and we should probably be optimizing for the happy path instead of the resend case. Happy to defer this to another ticket though

contracts/staking/PoSValidatorManager.sol
cam-schultz: This idea came up [previously](https://github.com/ava-labs/teleporter/pull/504#discussion_r1735220238). I agree that using a unique delegation ID rather than the address is the way to go, but going to defer that to a followup.

Using the nonce rather than a timestamp (as originally suggested in the linked conversation) seems like the way to go.

contracts/staking/interfaces/IPoSValidatorManager.sol
cam-schultz: Yup, don't want to make any assumptions about gas cost, so will defer that analysis and any implementation changes to a followup item.

bernard-avalabs (commented):
Generally makes sense to me. I assume there are no side effects from moving the weight update from completeEndDelegation to initializeEndDelegation.

contracts/staking/ERC20TokenStakingManager.sol
minghinmatthewlam: I think more so allowing Subnets to choose validator manager contract's that don't include delegation by default, and they can simply deploy the contract without delegation logic, as opposed to for reducing shared code.

cam-schultz: I agree that validators should have the option to not support delegation. There are many ways to do that, such as providing that toggle as a constructor argument, or leaving it up to the individual validator to specify when they register. I'll enumerate those options and defer to a followup.

minghinmatthewlam: sounds good I'm good with deferring. Likely would push back as a constructor argument for turn on/off, since that increases contract bytecode and deployment costs.

contracts/staking/PoSValidatorManager.sol
minghinmatthewlam: i like the (validationID, nonce) hash combo

contracts/staking/PoSValidatorManager.sol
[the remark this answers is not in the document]
minghinmatthewlam: good with deferring

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +144,196 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        // Ensure the validation period is active
+        Validator memory validator = _getValidator(validationID);
+        require(
+            validator.status == ValidatorStatus.Active, "ValidatorManager: validator not active"
+        );
+
+        // Ensure the delegator is not already registered
+        require(
+            $._delegatorStakes[validationID][delegator].weight == 0,
+            "ValidatorManager: delegator already registered"
```
minghinmatthewlam, line 165:
```suggestion
            "PoSValidatorManager: delegator already registered"
```

contracts/staking/ERC20TokenStakingManager.sol
geoff-vball: I don't think we need to worry about deployment costs, as these contracts will be deployed by subnet owners

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +144,196 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        // Ensure the delegator is not already registered
+        require(
+            $._delegatorStakes[validationID][delegator].weight == 0,
+            "PoSValidatorManager: delegator already registered"
+        );
+
+        _checkAndUpdateChurnTracker(weight);
+
+        // Update the validator weight
+        validator.weight += weight;
```
geoff-vball, line 171: I think we're going to need a new field to track the original weight of the validator. It'll be needed if the validator ends validation period to calculate their rewards, because right now we'd have to use the total weight with the delegations included.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,4 +144,196 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        _checkPendingRegisterDelegatorMessages(validationID, delegator);
+        delete $._pendingRegisterDelegatorMessages[validationID][delegator];
+
+        Validator memory validator = _getValidator(validationID);
+
+        // The received nonce should be no greater than the highest sent nonce
+        require(validator.messageNonce >= nonce, "PoSValidatorManager: invalid nonce");
+        // It should also be greater than or equal to the delegator's starting nonce
+        require(
+            $._delegatorStakes[validationID][delegator].startingNonce <= nonce,
```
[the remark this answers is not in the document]
minghinmatthewlam: nvm it's because P-chain is only willing to sign latest nonce weight update message. We keep a start nonce to check for registration completion, and ending nonce for ending completion.

minghinmatthewlam (approved):
LGTM

contracts/staking/PoSValidatorManager.sol
cam-schultz: I went ahead and added that here, so that it doesn't fall through the cracks.

minghinmatthewlam: I would advocate to only add it in when needed because future changes might not need it or want to implement differently, for example, whether `startTime` is only needed for PoS, or using a separate mechanism as a whole. Not a big deal though now that's already added.

Issue #488, opened 2024-08-08
Investigate PoA manager migration to PoS


minghinmatthewlam:
**Context and scope**
With ACP77 the new flow for PoA chains will be for the staking manager contracts to manage the validators of the chain. This should have similar logic to PoS staking manager, and already has an issue in #449 to scope out implementation for PoA contract. We should make sure chains are able to migrate from PoA -> PoS, and possibly vice versa.

**Discussion and alternatives**
Since the staking manager contracts are going to start with upgradability, the migration will likely include a contract upgrade. We need to ensure that there's no storage collision between the state variables of the PoA vs PoS contracts, which will only be obvious after seeing the state that needs to be saved for each implementation. 

Also for PoA there is likely an owner address, and thinking through how to replace for PoS.

**Open questions**
How important is flushing out PoS -> PoA?
Will PoA and PoS contracts have storage collision even with using namespaced storage?
What other factors to consider between differences of the implementations? Ex: Ownable, stake value -> weight calculation, state saved, etc.

PR #508, opened 2024-08-28
PoA -> PoS migration POC

minghinmatthewlam:
## Why this should be merged
Implements initial POC for #488 

## How this works
- uses `reinitializer(2)` on the PoS staking contracts to make sure after migration from PoA, the PoS contracts can reinitialize with its needed values like min staking amount, etc.
- setting both PoS contract to initialize 2 makes it so they can't by default upgrade to each other
- Also a higher initialize version for PoS contract since we by default don't support PoS->PoA.
- Adds an e2e test for testing out migration, and includes validator registration/ends from the different implementations

## How is this documented


tests/utils/staking.go
```diff
@@ -371,6 +371,132 @@ func completeValidatorRegistration(
 	return SendTransactionAndWaitForSuccess(context.Background(), subnet, signedRegistrationTx)
 }
 
+func InitializeAndCompleteNativeValidatorRegistration(
```
cam-schultz, line 374: For completeness, can we add `InitializeAndCompleteERC20ValidatorRegistration` and call that from `erc20_token_staking.go`?

contracts/staking/ValidatorManager.sol
```diff
@@ -204,6 +204,11 @@ abstract contract ValidatorManager is
         );
     }
 
+    function activeValidators(bytes32 nodeID) public view returns (bytes32) {
```
cam-schultz, line 207: Can we rename this to `getActiveValidatorID` or similar? The current name implies we return a collection of validators.

cam-schultz (commented):
Just a couple of nits, but generally looks great. Super cool that this particular workflow is so well supported.

contracts/staking/NativeTokenStakingManager.sol
```diff
@@ -27,7 +27,7 @@ contract NativeTokenStakingManager is
     }
 
     // solhint-disable ordering
-    function initialize(PoSValidatorManagerSettings calldata settings) external initializer {
+    function initialize(PoSValidatorManagerSettings calldata settings) external reinitializer(2) {
```
cam-schultz, line 30: If a PoA contract is upgraded separate from PoA->PoS migration, will this still work when we go to do the PoA->PoS migration?

contracts/staking/ValidatorManager.sol
minghinmatthewlam: i went with `activeValidators` to simulate the default getter function you'd have with normal public state variables

contracts/staking/NativeTokenStakingManager.sol
minghinmatthewlam: depends how the upgrade is done for the PoA before, but yes in most cases if it's just changing some of the logic of  the original contract.

contracts/staking/ValidatorManager.sol
cam-schultz: Yes, my only issue with that is that this getter doesn't return the `activeValidators` state variable, but instead returns a value from that mapping.

contracts/staking/NativeTokenStakingManager.sol
```diff
@@ -26,8 +26,13 @@ contract NativeTokenStakingManager is
...
     }
 
+    /**
+     * @notice Initialize the ERC20 token staking manager
+     * @dev Uses reinitializer(2) on the PoS staking contracts to make sure after migration from PoA, the PoS contracts can reinitialize with its needed values.
+     * @param settings Initial settings for the PoS validator manager
+     */
     // solhint-disable ordering
-    function initialize(PoSValidatorManagerSettings calldata settings) external initializer {
+    function initialize(PoSValidatorManagerSettings calldata settings) external reinitializer(2) {
```
cam-schultz, line 35: Where does the version `2` actually get set when performing an upgrade? I don't see it in `ProxyAdmin.sol`, which is where I would expect to see it.

I'm trying to understand how `reinitializer` works under the hood.

contracts/staking/ValidatorManager.sol
cam-schultz: Bumping this comment.

contracts/staking/NativeTokenStakingManager.sol
minghinmatthewlam: it's part of OZ's `Initializable` https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/Initializable.sol#L152

contracts/staking/ValidatorManager.sol
minghinmatthewlam: the default getter for a public mapping state varaible would act the same, the getter function returns the value associated with a passed in key, not the whole state variable.

cam-schultz: Ah, I did not realize that. Personal preference for names aside, sticking to established Solidity convention is the right call.

cam-schultz (approved):

PR #511, opened 2024-08-29
Churn tracker

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/480

## How this works
Ensures that after `churnTrackerStartTime`, only `maximumChurnPercentage` of stake is added/removed in `churnPeriodSeconds`.

I also renamed `minimumStakeAmount` and `maximumStakeAmount` to `minimumStakeWeight` and `maximumStakeWeight` to accurately reflect that we were dealing with converted weights, not token amounts.

## How this was tested
Unit tests

Delegation unit tests TODO

## How is this documented
Comments on fields


contracts/staking/ValidatorManager.sol
```diff
@@ -93,7 +95,17 @@ abstract contract ValidatorManager is
         ValidatorManagerStorage storage $ = _getValidatorManagerStorage();
         $._pChainBlockchainID = settings.pChainBlockchainID;
         $._subnetID = settings.subnetID;
-        $._maximumHourlyChurn = settings.maximumHourlyChurn;
+
+        require(
+            settings.maximumChurnPercentage <= 20,
+            "ValidatorManager: maximum churn percentage too high"
+        );
```
geoff-vball, line 102: We could also enforce maximum churn percentage > 0, but setting it to 0 to disable is useful at least for tests.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -21,10 +21,10 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
     // solhint-disable private-vars-leading-underscore
     /// @custom:storage-location erc7201:avalanche-icm.storage.PoSValidatorManager
     struct PoSValidatorManagerStorage {
-        /// @notice The minimum amount of stake required to be a validator.
-        uint256 _minimumStakeAmount;
-        /// @notice The maximum amount of stake allowed to be a validator.
-        uint256 _maximumStakeAmount;
+        /// @notice The minimum weight of stake required to be a validator.
```
minghinmatthewlam, line 24: I think the units for the PoS should be in staking amounts, since when users go to intiailize staking/delegation, it's with the stake amount units and we should keep consistent

contracts/staking/ValidatorManager.sol
minghinmatthewlam: 0 seems to be an invalid value to me

contracts/staking/ValidatorManager.sol
```diff
@@ -384,4 +371,59 @@ abstract contract ValidatorManager is
...
+
+    /**
+     * @dev Helper function to check if the stake weight to be added or removed would exceed the maximum stake churn
+     * rate for the past churn period. If the churn rate is exceeded, the function will revert. If the churn rate is
+     * not exceeded, the function will update the churn tracker with the new weight.
+     */
+    function _checkAndUpdateChurnTracker(uint64 weight, bool addition) private {
+        ValidatorManagerStorage storage $ = _getValidatorManagerStorage();
+
+        if ($._maximumChurnPercentage == 0) {
```
minghinmatthewlam, line 397: oh I actually understood initially for 0 to mean no churn can happen

contracts/staking/ValidatorManager.sol
iansuvak: Disabling churn tracker should only be useful in tests though right?

geoff-vball: Yeah, I guess with Stephen's comments about consensus failures, we should probably require it to be enabled. 

I think now that we're setting an initial validator set, tests should be easy enough to get to work around the churn tracker limits.

contracts/staking/ValidatorManager.sol
```diff
@@ -384,4 +371,59 @@ abstract contract ValidatorManager is
...
+            return;
+        }
+
+        uint256 currentTime = block.timestamp;
+        ValidatorChurnPeriod memory churnTracker = $._churnTracker;
+
+        if (
```
iansuvak, line 404: nit: Maybe we can add a comment to the branches of the conditional here, and maybe explain why the weight is always additive tot he `churnAmount` regardless of whether it's weight being added or removed. 

It is explained in the doc comment above so can drop it.

iansuvak (approved):
This LGTM. I think we can merge it and remove the `_maximumChurnPercentage == 0` case on the initial validator PR side

contracts/staking/PoSValidatorManager.sol
```diff
@@ -28,9 +28,9 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
     // solhint-disable private-vars-leading-underscore
     /// @custom:storage-location erc7201:avalanche-icm.storage.PoSValidatorManager
     struct PoSValidatorManagerStorage {
-        /// @notice The minimum amount of stake required to be a validator.
+        /// @notice The minimum weight of stake required to be a validator.
```
minghinmatthewlam, line 31:
```suggestion
        /// @notice The minimum amount of stake required to be a validator.
```

contracts/staking/PoSValidatorManager.sol
```diff
@@ -28,9 +28,9 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         uint256 _minimumStakeAmount;
-        /// @notice The maximum amount of stake allowed to be a validator.
+        /// @notice The maximum weight of stake allowed to be a validator.
```
minghinmatthewlam, line 33: same here

contracts/staking/PoSValidatorManager.sol
```diff
@@ -189,13 +189,14 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
                 && requirements.delegationFeeBips <= MAXIMUM_DELEGATION_FEE_BIPS,
             "PoSValidatorManager: invalid delegation fee"
         );
-        require(
-            stakeAmount >= $._minimumStakeAmount && stakeAmount <= $._maximumStakeAmount,
-            "PoSValidatorManager: invalid stake amount"
-        );
+
+        // Ensure the weight is within the valid range.
+        require(stakeAmount >= $._minimumStakeAmount, "PoSValidatorManager: stake amount too low");
```
minghinmatthewlam, line 194: thanks for breaking this up

contracts/staking/PoSValidatorManager.sol
```diff
@@ -343,6 +346,10 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         );
         uint64 nonce = _incrementAndGetNonce(validationID);
 
+        // Check that removing this delegator would not exceed the maximum churn rate.
+        // TODO this check won't be necessary for a delegator whose validator has already initialized ending their validation period.
+        _checkAndUpdateChurnTrackerRemoval(delegator.weight);
```
minghinmatthewlam, line 351: do you plan on handling this in this PR? I think we can check that the validator is PendingRemoved here, but not sure if I'm missing some other change we'd need.

contracts/staking/ValidatorManager.sol
```diff
@@ -92,7 +95,21 @@ abstract contract ValidatorManager is
...
         $._subnetID = settings.subnetID;
-        $._maximumHourlyChurn = settings.maximumHourlyChurn;
+
+        require(
+            settings.maximumChurnPercentage <= 20,
+            "ValidatorManager: maximum churn percentage too high"
+        );
+        require(
+            settings.maximumChurnPercentage > 0,
+            "ValidatorManager: maximum churn percentage cannot be zero"
```
minghinmatthewlam, line 105:
```suggestion
            "ValidatorManager: zero maximum churn percentage"
```

contracts/staking/ValidatorManager.sol
```diff
@@ -92,7 +95,21 @@ abstract contract ValidatorManager is
         ValidatorManagerStorage storage $ = _getValidatorManagerStorage();
         $._pChainBlockchainID = settings.pChainBlockchainID;
         $._subnetID = settings.subnetID;
-        $._maximumHourlyChurn = settings.maximumHourlyChurn;
+
+        require(
+            settings.maximumChurnPercentage <= 20,
```
minghinmatthewlam, line 100: can you use a constant here

contracts/staking/ValidatorManager.sol
```diff
@@ -106,11 +123,13 @@ abstract contract ValidatorManager is
...
-        // Ensure the registration expiry is in a valid range.
         require(
-            input.registrationExpiry > block.timestamp
-                && block.timestamp + 2 days > input.registrationExpiry,
-            "ValidatorManager: invalid registration expiry"
+            input.registrationExpiry > block.timestamp,
+            "ValidatorManager: registration expiry not in future"
+        );
+        require(
+            block.timestamp + 2 days > input.registrationExpiry,
```
minghinmatthewlam, line 131: not introduced in this PR, but would be good to have a constant for 2 days too

contracts/staking/ValidatorManager.sol
minghinmatthewlam: can the value still be zero, if not, is this needed?

minghinmatthewlam (approved):
LGTM

PR #514, opened 2024-08-29
[L-01] enable payable ValidatorSetSig targets

iansuvak:
## Why this should be merged
This will allow `ValidatorSetSig` contract to be used with `payable` target contract functions that require native tokens to be sent to them. 

## How this works
- Adds `receive()` as on option for funding the contract 
- Makes `executeCall()` `payable` to allow for funding the call directly.
- Adds a `value` field to the `ValidatorSetSigMessage` struct that will be signed by the validators and sends this much of the native token along with the payload to the `targetContractAddress`

This will use any funds available to the contract, either pre-received via `receive()` or directly supplied with the call to `executeCall()` function. Any remaining funds will remain on contract balance.

If there are insufficient funds the call will revert. 

## How this was tested
- Existing tests should pass
- New unit test was added testing sending value along with payload to a contract address. 

## How is this documented


michaelkaplan13:
Makes sense to me 👍 Just have to fix the E2E test flow looks like. I think the code scanning issue is likely a false positive.

michaelkaplan13 (approved):

contracts/governance/ValidatorSetSig.sol
```diff
@@ -113,7 +117,9 @@ contract ValidatorSetSig is ReentrancyGuard {
         // We don't need to protect against return bomb vectors below here since the caller is expected to have full control over the contract called.
         (bool success,) =
         // solhint-disable-next-line avoid-low-level-calls
-         validatorSetSigMessage.targetContractAddress.call(validatorSetSigMessage.payload);
+        validatorSetSigMessage.targetContractAddress.call{value: validatorSetSigMessage.value}(
```
minghinmatthewlam, line 120: do you know what error gets shown if `this` balance doesn't have enough funds?

minghinmatthewlam (approved):
LGTM

contracts/governance/ValidatorSetSig.sol
iansuvak: Internally  when testing it's `EvmError: OutOfFunds` but then  `bool success` is false so the actual revert message is `ValidatorSetSig: call failed`

geoff-vball (approved):

contracts/governance/ValidatorSetSig.sol
```diff
@@ -87,7 +89,9 @@ contract ValidatorSetSig is ReentrancyGuard {
         blockchainID = WARP_MESSENGER.getBlockchainID();
     }
 
-    function executeCall(uint32 messageIndex) external nonReentrant {
+    receive() external payable {}
+
+    function executeCall(uint32 messageIndex) external payable nonReentrant {
```
cam-schultz, line 94: For my own understanding, is `payable` here strictly necessary in order to specify a value in the below `call`? If this was not `payable`, would the behavior be that this function is callable by anyone (assuming the contract itself has the required funds, provided through `receive`)

cam-schultz (approved):
Just one clarifying question, but LGTM assuming my understanding is correct

contracts/governance/ValidatorSetSig.sol
iansuvak: It is not strictly necessary, as long as the contract has enough funds the `call` inside `executeCall` will be able to send them along to the target. It was made payable as a convenience to give two different ways of sending native tokens to the target contract.

bernard-avalabs (approved):

Issue #489, opened 2024-08-13
Unlock stake and issue rewards


cam-schultz:
a**Context and scope**
After completing a validation and calculating the staking rewards (#436), we should unlock those rewards such that they can be redeemed by the staker.

- [x] https://github.com/ava-labs/teleporter/pull/527
- [x] https://github.com/ava-labs/teleporter/pull/539
- [x] https://github.com/ava-labs/teleporter/pull/534
- [ ] look at removing the stale validationPeriods state if no longer needed

minghinmatthewlam:
after unlocking the staking rewards as part of `completeEndValidation` should also look at removing the stale `validationPeriods` state if no longer needed

PR #520, opened 2024-09-03
Track uptimes for delegation period

cam-schultz:
## Why this should be merged
Modifies `initializeEndDelegation` to accept an uptime proof from the P-Chain to be used when calculating rewards in `completeEndDelegation`. If no uptime proof is provided, the uptime is assumed to be zero. 

This PR does not implement delegation rewards calculation or unlocks. That should be done in the same fashion as https://github.com/ava-labs/teleporter/issues/489, or added as part of that work.

## How this works
See above

## How this was tested
CI. Skipped unit tests for now, since we are not doing anything interesting with the uptime proofs yet.

## How is this documented
N/A


contracts/staking/PoSValidatorManager.sol
```diff
@@ -92,34 +92,38 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        );
+
+        (bytes32 uptimeValidationID, uint64 uptime) =
+            ValidatorMessages.unpackValidationUptimeMessage(warpMessage.payload);
+        require(
+            validationID == uptimeValidationID, "PoSValidatorManager: invalid uptime validation ID"
+        );
+
+        $._validatorUptimes[validationID] = uptime;
+        emit ValidationUptimeUpdated(validationID, uptime);
```
geoff-vball, line 122: We should only update the validator's uptime if the uptime provided is greater than the previously provided uptime. This way delegators may not have to provide an uptime proof to get their rewards, provided the validator has had a sufficiently high uptime.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -241,7 +246,16 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
 
-    function initializeEndDelegation(bytes32 validationID) external {
+    function initializeEndDelegation(
+        bytes32 validationID,
+        bool includeUptimeProof,
+        uint32 messageIndex
+    ) external {
+        uint64 uptime;
+        if (includeUptimeProof) {
+            uptime = _updateUptime(validationID, messageIndex);
```
geoff-vball, line 256: We should read from state instead of using a return from `_updateUptime`. That way if the validator's uptime has been updated recently, anyone ending their delegation won't have to submit another uptime proof, provided that the latest reported uptime up until now hits the 80% threshold.

geoff-vball (commented):
I think we should also add an external `updateUptime` function that can be called by anyone to update a validator's uptime. 

I think we should also split the `initializeEndDelegation` and `initializeEndValidation` into two functions, one that expects the user to be reward eligible and one that doesn't. That way if a used calls the function expecting to be reward eligible, we can revert and not cause irreversible losses of rewards.

cam-schultz:
> I think we should also add an external updateUptime function that can be called by anyone to update a validator's uptime.

> I think we should also split the initializeEndDelegation and initializeEndValidation into two functions, one that expects the user to be reward eligible and one that doesn't. That way if a used calls the function expecting to be reward eligible, we can revert and not cause irreversible losses of rewards.

Discussed both of these points in a call, and agreed that they would be optimizations that would potentially trade gas efficiency for usability. Requiring an uptime proof to be submitted when intializing validation/delegation end in order to be eligible for rewards is simple to reason about, and cheap assuming that fetching and verifying an uptime proof is cheap. On the other hand, decoupling uptime proof delivery from delegation/validation end could save gas in some circumstances, but would require the user to first check if they're eligible for rewards under the latest known proof, and would also require that proof be stored in the contract state.

geoff-vball (approved):

minghinmatthewlam (approved):

contracts/staking/PoSValidatorManager.sol
```diff
@@ -92,34 +90,36 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        require(
+            warpMessage.sourceChainID == WARP_MESSENGER.getBlockchainID(),
+            "PoSValidatorManager: invalid source chain ID"
+        );
+        require(
+            warpMessage.originSenderAddress == address(0),
+            "PoSValidatorManager: invalid origin sender address"
+        );
+
+        (bytes32 uptimeValidationID, uint64 uptime) =
```
bernard-avalabs, line 114: Just for my own understanding, does uptime return the absolute amount of time that the validator was up, or a percent of time that it was up during the staking period? I'm just thinking of whether other information is necessary to calculate rewards.

bernard-avalabs (approved):

contracts/staking/PoSValidatorManager.sol
geoff-vball: It returns the absolute number of seconds the validator has been up

PR #521, opened 2024-09-04
Require non-zero delegation fee

### LINKED ISSUE #505: Delegation: Impose nonzero delegation fee
**Context and scope**
In order to properly incentive validation over exclusively delegation, validators should charge a nonzero fee of any delegation rewards.

The fee should be provided in `initializeValidatorRegistration`. The minimum fee should be set in the constructor, and should not be allowed to be zero.

iansuvak:
## Why this should be merged
Fixes #505

## How this works
Adds a new argument to the PoSValidatorManager initializer that sets a global minimum delegation fee rate and then enforces that any new validation respects it through a check in the initializer. 

It stores the rates in a separate `mapping` but we might consider adding it to the `Validator` struct. 

## How this was tested
Manually tested right now. I'm running into issues with  `vm.expectRevert()` presumably due to call nesting. Calls with invalid delegation fees that should revert fail both with and without `vm.expectRevert()`.

## How is this documented
Comments


tests/utils/staking.go
```diff
@@ -131,10 +131,11 @@ func DeployAndInitializeERC20TokenStakingManager(
...
 				MaximumHourlyChurn: 0,
 			},
-			MinimumStakeAmount:   big.NewInt(0).SetUint64(1e6),
-			MaximumStakeAmount:   big.NewInt(0).SetUint64(10e6),
-			MinimumStakeDuration: uint64(24 * time.Hour),
-			RewardCalculator:     common.Address{},
+			MinimumStakeAmount:       big.NewInt(0).SetUint64(1e6),
+			MaximumStakeAmount:       big.NewInt(0).SetUint64(10e6),
+			MinimumStakeDuration:     uint64(24 * time.Hour),
+			MinimumDelegationFeeRate: big.NewInt(0).SetUint64(100),
```
geoff-vball, line 137:
```suggestion
			MinimumStakeAmount:       big.NewInt(1e6),
			MaximumStakeAmount:       big.NewInt(10e6),
			MinimumStakeDuration:     uint64(24 * time.Hour),
			MinimumDelegationFeeRate: big.NewInt(100),
```

tests/utils/staking.go
```diff
@@ -213,6 +214,7 @@ func InitializeNativeValidatorRegistration(
 		opts,
 		nodeID,
 		uint64(time.Now().Add(24*time.Hour).Unix()),
+		big.NewInt(0).SetUint64(100),
```
geoff-vball, line 217:
```suggestion
		big.NewInt(100),
```

tests/utils/staking.go
```diff
@@ -252,6 +254,7 @@ func InitializeERC20ValidatorRegistration(
 		stakeAmount,
 		nodeID,
 		uint64(time.Now().Add(24*time.Hour).Unix()),
+		big.NewInt(0).SetUint64(100),
```
geoff-vball, line 257:
```suggestion
		big.NewInt(100),
```

geoff-vball:
Looks like Cam opened a ticket for disabling delegation. Not sure if you think it makes sense to implement here as well or want to defer to another PR https://github.com/ava-labs/teleporter/issues/522

minghinmatthewlam:
> Looks like Cam opened a ticket for allowing validators to disable delegation. Not sure if you think it makes sense to implement here as well or want to defer to another PR #522

This was a followup from a comment to separate delegation as an extension to allow for people to choose validator managers that don't have delegation built in by default. Can still move forward here, and move all the delegation out altogether

tests/utils/staking.go
iansuvak: In general I think using `big.NewInt(0).SetUint64()` is a safer practice than doing `big.NewInt()` directly since the latter relies on implicit conversion between types and particularly since scientific notation numbers are float64's and susceptible to precision loss issues.

That being said I agree that we should be consistent and I will update my additions to use the former pattern

contracts/staking/PoSValidatorManager.sol
```diff
@@ -38,6 +40,8 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
             _pendingEndDelegatorMessages;
         /// @notice Maps the validationID to the uptime of the validator.
         mapping(bytes32 validationID => uint64) _validatorUptimes;
+        /// @notice Maps the validationID to the delegation fee rate.
+        mapping(bytes32 validationID => uint256) _validatorDelegationFeeRates;
```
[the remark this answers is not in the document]
geoff-vball: I do like keeping the staking specific settings in the staking contract. We should think about if this is also the simplest solution given the PoA -> PoS upgrade potential, which I think it does? If we allow Validators to set their fee rate to 0 to disable delegation, then we could automatically get the property of PoA validators not being able to be delegated to, I think?

contracts/staking/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -22,6 +22,10 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
     address public constant DEFAULT_DELEGATOR_ADDRESS =
         address(0x1234123412341234123412341234123412341234);
 
+    // This is the rate that will be passed into the child contract `initializeValidatorRegistration` calls
+    // It is set here to avoid having to pass irrelevant initializers to the parent contract.
+    uint256 public delegationFeeRate;
```
iansuvak, line 27: I'm not set on this stateful solution being the cleanest way to pass in delegationFeeRate to child test contracts. Happy to hear alternatives.

tests/utils/staking.go
iansuvak: I agree that the only difference in practice here is  precision difference between int64 overflow and max uint64 which is not relevant. I will make the change

geoff-vball (approved):

contracts/staking/ERC20TokenStakingManager.sol
```diff
@@ -88,16 +88,21 @@ contract ERC20TokenStakingManager is
...
+     * @param delegationFeeRate The fee charged to delegators by the validator.
      * @param blsPublicKey The BLS public key of the validator.
      */
     function initializeValidatorRegistration(
         uint256 stakeAmount,
         bytes32 nodeID,
         uint64 registrationExpiry,
+        uint256 delegationFeeRate,
         bytes memory blsPublicKey
     ) external returns (bytes32 validationID) {
```
michaelkaplan13, line 100: The inconsistency wasn't introduced in this PR, but we've generally tried to be consistent in not using named returned values. This has been a minor item that has popped up in past audits. Could you switch to using non-named returns here?

contracts/staking/PoSValidatorManager.sol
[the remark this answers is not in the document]
iansuvak: I don't think that I love overloading the function of the fee rate, but this might be the cheapest/cleanest way of doing this. I will give it a try and consider alternatives.

PR #528, opened 2024-09-05
Delegation ID

### LINKED ISSUE #523: Delegation: Separate delegations from the delegator address
**Context and scope**
Delegations are currently identified by the address that calls `initializeRegisterDelegator`. A more general approach would be to compute a unique `delegationID`, which would enable a single address to delegate to the same validator multiple times.

Rather than the mapping `mapping(bytes32 validationID => mapping(address delegator => Delegator))`, we should use `mapping(bytes32 delegationID => Delegator)`, where `delegationID` is a hash of the `validationID`, the delegator's address, and the validation's nonce used to increase the validator's weight on delegator registration. The delegator's address would need to be moved to the `Delegator` struct to correctly distribute rewards. In addition to the functional use cases described above, this would remove a nested mapping (cheaper), and provide a globally unique delegation identifier, which may be useful in other contexts.

cam-schultz:
## Why this should be merged
Fixes #523 

## How this works
Converts the internal mappings to track delegations by a unique `delegationID` rather than the delegator's address. Updates the emitted events to reflect the inputs to the `delegationID` on registration, and the fields necessary to trace the delegation status and validator weight updates for the other methods.

## How this was tested
CI

## How is this documented
N/A


contracts/staking/PoSValidatorManager.sol
```diff
@@ -149,7 +147,7 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         bytes32 validationID,
         address delegator,
         uint256 delegationAmount
-    ) internal nonReentrant {
+    ) internal nonReentrant returns (bytes32) {
```
geoff-vball, line 150: Can we rename `delegator` to `delegatorAddress`? I was reading through the function and mistook it for a `Delegator` object.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -191,72 +192,73 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         emit DelegatorAdded({
             validationID: validationID,
             setWeightMessageID: messageID,
-            delegator: delegator,
+            delegationID: delegationID,
```
geoff-vball, line 195: We should include the delegator address in the event, so that the delegation ID can be related to an address

contracts/staking/PoSValidatorManager.sol
```diff
@@ -191,72 +192,73 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
             uptime = _getUptime(validationID, messageIndex);
         }
 
         // TODO: Calculate the delegator's reward, but do not unlock it
 
-        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
-
         // Ensure the delegator is active
-        Delegator memory delegator = $._delegatorStakes[validationID][_msgSender()];
+        Delegator memory delegator = $._delegatorStakes[delegationID];
```
geoff-vball, line 261: Since we're no longer using `_msgSender()`, we need to explicitly check that this function is being called by the address that owns the delegation.

geoff-vball (commented):
In general, I think we should make sure all the delegation functions are called by the right caller, and I think it would also be helpful to keep the delegator address in the events that you added the delegationID to.

cam-schultz:
> I think we should make sure all the delegation functions are called by the right caller

I added that check to `initializeEndDelegation`, but I see no reason to restrict delivery of the P-Chain acknowledgment in the corresponding `complete` methods. Note that we don't do so for the validation `complete` methods in `ValidatorManager` either.

geoff-vball:
> > I think we should make sure all the delegation functions are called by the right caller
> 
> I added that check to `initializeEndDelegation`, but I see no reason to restrict delivery of the P-Chain acknowledgment in the corresponding `complete` methods. Note that we don't do so for the validation `complete` methods in `ValidatorManager` either.

I think if we're withdrawing funds, that should only be able to be completed by the appropriate account. If we're just delivering a message and cleaning up state then anyone should be able to call

cam-schultz:
> I think if we're withdrawing funds, that should only be able to be completed by the appropriate account. If we're just delivering a message and cleaning up state then anyone should be able to call

100% agree. The flow I roughly have in mind is that `completeEndDelegation` would mark the funds as redeemable, but not transfer them. A separate method would be called (and only be callable by the delegation owner) to claim the funds. Either way, I think the question of how and when to mark funds available and disburse them is best left for a separate issue.

geoff-vball (approved):
LGTM 👍

contracts/staking/PoSValidatorManager.sol
```diff
@@ -29,13 +29,11 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
-        mapping(bytes32 validationID => mapping(address delegator => Delegator)) _delegatorStakes;
+        mapping(bytes32 delegationID => Delegator) _delegatorStakes;
         /// @notice Maps the validationID to a mapping of delegator address to pending register delegator messages.
-        mapping(bytes32 validationID => mapping(address delegator => bytes))
-            _pendingRegisterDelegatorMessages;
+        mapping(bytes32 delegationID => bytes) _pendingRegisterDelegatorMessages;
         /// @notice Maps the validationID to a mapping of delegator address to pending end delegator messages.
-        mapping(bytes32 validationID => mapping(address delegator => bytes))
-            _pendingEndDelegatorMessages;
+        mapping(bytes32 delegationID => bytes) _pendingEndDelegatorMessages;
```
michaelkaplan13, line 36: Can you update the comments for these mappings?

cam-schultz: Done

contracts/staking/PoSValidatorManager.sol
```diff
@@ -158,29 +156,32 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         require(
             validator.status == ValidatorStatus.Active, "PoSValidatorManager: validator not active"
         );
+        // Update the validator weight
+        uint64 newValidatorWeight = validator.weight + weight;
+        _setValidatorWeight(validationID, newValidatorWeight);
+
+        uint64 nonce = _getAndIncrementNonce(validationID);
+        bytes32 delegationID = sha256(abi.encodePacked(validationID, delegatorAddress, nonce));
```
michaelkaplan13, line 164: I think we've used keccack for hashing in all places we're able to since it's slightly less gas and arguably more secure (though don't think that matters in this case here).

Should we use keccack here also just for consistency though?

contracts/staking/PoSValidatorManager.sol
```diff
@@ -158,29 +156,32 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
 
-        // Ensure the delegator is not already registered
+        // Ensure the delegationID is not already registered
         require(
-            $._delegatorStakes[validationID][delegator].weight == 0,
-            "PoSValidatorManager: delegator already registered"
+            $._delegatorStakes[delegationID].weight == 0,
```
michaelkaplan13, line 168: I think this check can be removed since `delegationID` is guaranteed to be unique for every call right?

contracts/staking/PoSValidatorManager.sol
cam-schultz: I chose `sha256` to be consistent with `validationID`, but don't feel strongly either way. Since the `delegationID` isn't used outside of the validator manager contracts, I think gas cost should be the primary consideration.

contracts/staking/PoSValidatorManager.sol
cam-schultz: True, nice catch

PR #534, opened 2024-09-06
Staking rewards

feuGeneA:
## Why this should be merged
for #489 

## How this was tested
via the unit tests herein


contracts/staking/PoSValidatorManager.sol
```diff
@@ -98,6 +98,14 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
 
     function completeEndValidation(uint32 messageIndex) external {
         Validator memory validator = _completeEndValidation(messageIndex);
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        uint256 rewardAmount = $._rewardCalculator.calculateReward(
```
michaelkaplan13, line 103: Should this reward be calculated above in `initializeEndValidation`, as the TODO there suggests?

I think the reason it would need to be calculated in `intializeEndValidation` is because the uptime proof is included there, which should be used in calculating the rewards. I think we'd either need to store the uptime submitted to be used here, or calculate and store the reward during the call to `initializeEndValidation`.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -356,7 +364,11 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
 
         Delegator memory delegator = $._delegatorStakes[delegationID];
         _unlock(delegator.weight, delegator.owner);
-        // TODO: issue rewards
+
+        uint256 rewardAmount = $._rewardCalculator.calculateReward(
```
michaelkaplan13, line 368: Same here, regarding the TODO on line 280

contracts/staking/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -372,6 +387,23 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
         posValidatorManager.resendEndDelegation(delegationID);
     }
 
+    function testResendEndValidation() public override {
```
feuGeneA, line 390: This `override` was brought to this level (from `ValidatorManagerTests`) just for the inclusion of `includeUptime: true`.

contracts/staking/tests/ValidatorManagerTests.t.sol
```diff
@@ -127,7 +127,7 @@ abstract contract ValidatorManagerTest is Test {
         });
     }
 
-    function testInitializeEndValidation() public {
+    function testInitializeEndValidation() public virtual {
```
feuGeneA, line 130: Made virtual so it can be overridden in the PoS test file, in order to induce uptime logic.

contracts/staking/tests/ValidatorManagerTests.t.sol
```diff
@@ -143,7 +143,7 @@ abstract contract ValidatorManagerTest is Test {
         // TODO: implement
     }
 
-    function testResendEndValidation() public {
+    function testResendEndValidation() public virtual {
```
feuGeneA, line 146: Made virtual so it can be overridden in the PoS test file, in order to induce uptime logic.

contracts/staking/PoSValidatorManager.sol
feuGeneA: that makes sense to me, thanks for calling it out. I had seen that TODO but had been unsure about it, since the inputs needed for the calculator (start and end times in particular) were readily available at the point where the actual disbursement needs to happen (`completeEndValidation`). But now I see the discussion of `uptime` in the README, and it makes sense to me that the proven uptime is the basis of the rewards, so I re-worked the calculator's inputs to take an uptime rather than a start and an end, and used the proven uptime as that input to the calculation, in https://github.com/ava-labs/teleporter/pull/534/commits/57a519e728da4431a644ebd148f588bad396ad9e

contracts/staking/PoSValidatorManager.sol
feuGeneA: as discussed in https://github.com/ava-labs/teleporter/pull/534#discussion_r1754707932 ,
addressed in https://github.com/ava-labs/teleporter/pull/534/commits/57a519e728da4431a644ebd148f588bad396ad9e

contracts/staking/PoSValidatorManager.sol
```diff
@@ -88,16 +92,26 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         }
-        // TODO: Calculate the reward for the validator, but do not unlock it
 
-        _initializeEndValidation(validationID);
+        Validator memory validator = _initializeEndValidation(validationID);
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        $._pendingValidatorRewards[validator.owner] =
+            $._rewardCalculator.calculateReward(validator.weight, uptime, 0, 0);
     }
```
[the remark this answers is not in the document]
feuGeneA: fixed in https://github.com/ava-labs/teleporter/pull/534/commits/f26ecbe8cb75b19247396aa0c083b48ab5f2f02c

contracts/mocks/ExampleERC20.sol
```diff
@@ -30,4 +31,11 @@ contract ExampleERC20 is ERC20Burnable {
 
         _mint(msg.sender, amount);
     }
+
+    function mint(address account, uint256 amount) external {
+        // Can only mint 10 at a time.
```
geoff-vball, line 36: Is this comment correct? If _MAX_MINT is `1e16` that would mean decimals is `15`?

contracts/staking/ExampleRewardCalculator.sol
```diff
@@ -21,11 +21,10 @@ contract ExampleRewardCalculator is IRewardCalculator {
...
     function calculateReward(
         uint256 stakeAmount,
-        uint64 startTime,
-        uint64 endTime,
+        uint64 uptime,
         uint256, // initialSupply
         uint256 // endSupply
     ) external view returns (uint256) {
-        return (stakeAmount * rewardBasisPoints * (endTime - startTime)) / SECONDS_IN_YEAR / 1000;
+        return (stakeAmount * rewardBasisPoints * uptime) / SECONDS_IN_YEAR / 1000;
```
geoff-vball, line 28: I don't think we want this change in the example. On the p-chain today, rewards are given for the whole staking period as long as an uptime threshold of 80% has been reached. So the calculator should give the full reward. The contract itself can figure out if the validator gets all or none of the rewards calculated.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -35,6 +35,10 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         mapping(bytes32 delegationID => bytes) _pendingRegisterDelegatorMessages;
         /// @notice Maps the delegationID to pending end delegator messages.
         mapping(bytes32 delegationID => bytes) _pendingEndDelegatorMessages;
+        /// @notice Maps the delegationID to its pending staking rewards.
+        mapping(bytes32 delegationID => uint256) _pendingDelegatorRewards;
+        /// @notice Maps the validator owner address to its pending staking rewards.
+        mapping(address validatorOwner => uint256) _pendingValidatorRewards;
```
geoff-vball, line 41: I think this might be better named as `unclaimedXRewards`. "Pending" makes me think that this should calculate how many rewards a staker would get if they exited now. I think we could make this one map with the key being the address that can claim the rewards.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -88,16 +92,26 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            uptime = _getUptime(validationID, messageIndex);
         }
-        // TODO: Calculate the reward for the validator, but do not unlock it
 
-        _initializeEndValidation(validationID);
+        Validator memory validator = _initializeEndValidation(validationID);
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        $._pendingValidatorRewards[validator.owner] +=
+            $._rewardCalculator.calculateReward(validator.weight, uptime, 0, 0);
```
geoff-vball, line 104: this needs to be calculated with `validator.startingWeight`, since `weight` includes all the weight of the delegations, which should not be given to the validator.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -88,16 +92,26 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
 
     function completeEndValidation(uint32 messageIndex) external {
         Validator memory validator = _completeEndValidation(messageIndex);
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        _reward(validator.owner, $._pendingValidatorRewards[validator.owner]);
+        delete $._pendingValidatorRewards[validator.owner];
+
         _unlock(validator.startingWeight, validator.owner);
```
geoff-vball, line 115: We should be consistent with the parameter ordering of `_reward` and `_unlock`

contracts/staking/PoSValidatorManager.sol
```diff
@@ -264,15 +278,16 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        Delegator memory delegator = $._delegatorStakes[delegationID];
+
         uint64 uptime;
         if (includeUptimeProof) {
             uptime = _getUptime(validationID, messageIndex);
         }
-
-        // TODO: Calculate the delegator's reward, but do not unlock it
+        $._pendingDelegatorRewards[delegationID] =
+            $._rewardCalculator.calculateReward(delegator.weight, uptime, 0, 0);
```
geoff-vball, line 288: We definitely shouldn't be passing the validator's uptime into the reward calculator. Currently if a delegator staked for one second, they would get rewards for the entire validator's uptime.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -356,7 +371,9 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
 
         Delegator memory delegator = $._delegatorStakes[delegationID];
         _unlock(delegator.weight, delegator.owner);
-        // TODO: issue rewards
+
+        _reward(delegator.owner, $._pendingDelegatorRewards[delegationID]);
+        delete $._pendingDelegatorRewards[delegationID];
```
geoff-vball, line 376: We should keep the ordering of these two operations consistent with the validator function.

contracts/staking/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -557,25 +601,36 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
...
             blsPublicKey: DEFAULT_BLS_PUBLIC_KEY,
             registrationTimestamp: DEFAULT_REGISTRATION_TIMESTAMP,
-            completionTimestamp: DEFAULT_COMPLETION_TIMESTAMP
+            completionTimestamp: uint64(registrationExpiry),
+            includeUptime: true
         });
 
         uint256 balanceBefore = _getStakeAssetBalance(address(this));
 
+        uint256 expectedReward = DEFAULT_WEIGHT * DEFAULT_REWARD_RATE / 365 - 1;
```
geoff-vball, line 622: expected rewards shouldn't be based on weight, but on the value staked which is currently a 12 order of magnitude difference

contracts/staking/PoSValidatorManager.sol
minghinmatthewlam: i would go with `redeemableRewards`

contracts/staking/PoSValidatorManager.sol
```diff
@@ -88,16 +92,26 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        $._pendingValidatorRewards[validator.owner] +=
+            $._rewardCalculator.calculateReward(validator.weight, uptime, 0, 0);
     }
 
     function completeEndValidation(uint32 messageIndex) external {
         Validator memory validator = _completeEndValidation(messageIndex);
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
```
minghinmatthewlam, line 110: when we need to access the `PoSValidatorManagerStorage` can you add it as the first line in function

contracts/staking/PoSValidatorManager.sol
minghinmatthewlam: we could just put the delete in the `_reward` function too

minghinmatthewlam: nvm this is for both PoS so should leave here

contracts/staking/ExampleRewardCalculator.sol
```diff
@@ -26,6 +26,6 @@ contract ExampleRewardCalculator is IRewardCalculator {
         uint256, // initialSupply
         uint256 // endSupply
     ) external view returns (uint256) {
-        return (stakeAmount * rewardBasisPoints * (endTime - startTime)) / SECONDS_IN_YEAR / 1000;
+        return (stakeAmount * rewardBasisPoints * (endTime - startTime)) / SECONDS_IN_YEAR / 10000;
```
geoff-vball, line 29: This was a mistake I originally made. Basis points are 1/100 of a percent :)

contracts/staking/PoSValidatorManager.sol
```diff
@@ -129,25 +139,44 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         require(
-            block.timestamp
+            validator.endedAt
                 >= validator.startedAt + $._validatorRequirements[validationID].minStakeDuration,
             "PoSValidatorManager: minimum stake duration not met"
         );
 
         if (includeUptimeProof) {
-            _getUptime(validationID, messageIndex);
+            uint64 uptime = _getUptime(validationID, messageIndex);
```
minghinmatthewlam, line 153: can you add comments here explaining the value from uptime, and what the check is checking

contracts/staking/PoSValidatorManager.sol
```diff
@@ -129,25 +139,44 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
-        _initializeEndValidation(validationID);
     }
 
     function completeEndValidation(uint32 messageIndex) external {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
         Validator memory validator = _completeEndValidation(messageIndex);
-        _unlock(validator.startingWeight, validator.owner);
+
+        _reward(validator.owner, $._redeemableValidatorRewards[validator.owner]);
```
minghinmatthewlam, line 176: this should be check to be PoS Validator, if it's a PoA after migration just does nothign else. Also calling since these functions make external calls for `token.safeTransfer` or mint, we need to watch out for reentrancy. I should have it added already in this PR though https://github.com/ava-labs/teleporter/pull/547/files. 

Can you change the ordering to adhere to check-effects-interaction pattern? Would cause an extra read of reading the state variable redeemable rewards to a local variable, deleting state, then making the external calls, but think it's worth for security and readability.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -350,13 +375,22 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         delegator.status = DelegatorStatus.PendingRemoved;
-        delegator.endedAt = uint64(block.timestamp);
+        delegator.endedAt = currentTime;
         delegator.endingNonce = nonce;
 
         $._delegatorStakes[delegationID] = delegator;
 
-        Validator memory validator = getValidator(validationID);
-        require(validator.weight > delegator.weight, "PoSValidatorManager: Invalid weight");
+        if (includeUptimeProof) {
```
minghinmatthewlam, line 383: seems similar to that of validator end, any helpers that can be used?

contracts/staking/PoSValidatorManager.sol
```diff
@@ -431,26 +467,26 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
             "PoSValidatorManager: delegation not pending added"
         );
 
         // Update the delegator status
         $._delegatorStakes[delegationID].status = DelegatorStatus.Completed;
 
-        Delegator memory delegator = $._delegatorStakes[delegationID];
-        _unlock(delegator.weight, delegator.owner);
-        // TODO: issue rewards
+        _reward(delegator.owner, $._redeemableDelegatorRewards[delegationID]);
```
minghinmatthewlam, line 483: same as above for validation end

minghinmatthewlam (commented):
mostly LGTM, just some small comments

bernard-avalabs (commented):
Generally makes sense to me. I think Matt pointed out a missing check for PoA vs. PoS that should be fixed. There is also some repeated code between the validator and delegator reward check, but it might not be worth abstracting that out into a helper since it is not used anywhere else.

contracts/staking/PoSValidatorManager.sol
geoff-vball: So, I tested out adding a check for `validator.owner == address(0)` here, but without your other changes to move the `owner` to be PoS-only, it doesn't currently do anything for the migration case.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -60,6 +68,8 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
 
     uint8 public constant MAXIMUM_STAKE_MULTIPLIER_LIMIT = 10;
 
+    uint8 public constant UPTIME_REWARDS_THRESHOLD_PERCENTAGE = 80;
```
michaelkaplan13, line 71: Lets move this to the example rewards calculator, and add the `uptimeSeconds` to the calculate rewards interface.

That allows for different rewards calculators to have different uptime requirements, or even to pro-rated rewards based on uptime.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -129,25 +139,44 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         );
 
         if (includeUptimeProof) {
-            _getUptime(validationID, messageIndex);
+            uint64 uptime = _getUptime(validationID, messageIndex);
+            if (
+                uptime * 100
+                    >= (validator.endedAt - validator.startedAt) * UPTIME_REWARDS_THRESHOLD_PERCENTAGE
+            ) {
+                $._redeemableValidatorRewards[validator.owner] += $
```
michaelkaplan13, line 158: [Not for this PR - can be after the initial audit]

Primary network validations allow for specifying a different rewards address than the address that created the validation. Would be good to support that in the contracts as well.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -50,6 +50,14 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         /// @notice Maps the delegationID to the delegator information.
         mapping(bytes32 delegationID => Delegator) _delegatorStakes;
+        /// @notice Maps the delegationID to pending register delegator messages.
+        mapping(bytes32 delegationID => bytes) _pendingRegisterDelegatorMessages;
+        /// @notice Maps the delegationID to pending end delegator messages.
+        mapping(bytes32 delegationID => bytes) _pendingEndDelegatorMessages;
+        /// @notice Maps the delegationID to its pending staking rewards.
+        mapping(bytes32 delegationID => uint256) _redeemableDelegatorRewards;
+        /// @notice Maps the validator owner address to its pending staking rewards.
+        mapping(address validatorOwner => uint256) _redeemableValidatorRewards;
```
michaelkaplan13, line 60: Should we key this mapping on `validationID` rather than `validatorOwner`. That may be more intuitive since a single address could have multiple active validation periods at any point in time.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -350,13 +375,22 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         // the complete step, even if the delivered nonce is greater than the nonce used to
         // initialize the removal.
         delegator.status = DelegatorStatus.PendingRemoved;
-        delegator.endedAt = uint64(block.timestamp);
+        delegator.endedAt = currentTime;
```
michaelkaplan13, line 378: Do we handle the case when the validator has already previously been ended correctly here?

In that case, we shouldn't send any message to the P-chain, and the delegator should only be reward for the time period up until the validator end time.

contracts/staking/PoSValidatorManager.sol
geoff-vball: https://github.com/ava-labs/teleporter/issues/555

contracts/staking/interfaces/IRewardCalculator.sol
```diff
@@ -8,8 +8,10 @@ pragma solidity 0.8.25;
 interface IRewardCalculator {
     function calculateReward(
         uint256 stakeAmount,
-        uint64 startTime,
-        uint64 endTime,
+        uint64 validatorStartTime,
```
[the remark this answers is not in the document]
geoff-vball: Yeah, exactly. Otherwise we can't calculate the uptime percentage.

contracts/staking/PoSValidatorManager.sol
geoff-vball: Accounted for in this PR https://github.com/ava-labs/teleporter/pull/554

contracts/staking/PoSValidatorManager.sol
geoff-vball: Not as far as I can think...

contracts/staking/PoSValidatorManager.sol
```diff
@@ -129,25 +135,52 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            return;
+        }
+
+        if (validator.status == ValidatorStatus.Completed) {
+            uint256 rewards = $._redeemableValidatorRewards[validationID];
+            delete $._redeemableValidatorRewards[validationID];
+            _reward(validator.owner, rewards);
+        }
+
+        // We unlock the state whether the validation period is completed or invalidated.
```
cam-schultz, line 182:
```suggestion
        // We unlock the stake whether the validation period is completed or invalidated.
```

contracts/staking/PoSValidatorManager.sol
```diff
@@ -350,13 +379,32 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         $._delegatorStakes[delegationID] = delegator;
 
-        Validator memory validator = getValidator(validationID);
-        require(validator.weight > delegator.weight, "PoSValidatorManager: Invalid weight");
+        uint64 validatorUptimeSeconds;
+        if (
+            validator.status == ValidatorStatus.PendingRemoved
+                || validator.status == ValidatorStatus.Completed
+        ) {
+            // If the validation period has already ended, we have saved the uptime.
```
cam-schultz, line 392:
```suggestion
            // If the validation period has already ended, we have saved the uptime.
            // Further, it is impossible to retrieve an uptime proof for an already ended validation,
            // so there's no need to check any uptime proof provided in this function call.
```

contracts/staking/PoSValidatorManager.sol
```diff
@@ -350,13 +379,32 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            validatorStartTime: validator.startedAt,
+            stakingStartTime: delegator.startedAt,
+            stakingEndTime: delegator.endedAt,
+            uptimeSeconds: validatorUptimeSeconds,
+            initialSupply: 0,
+            endSupply: 0
+        });
 
         // Check that removing this delegator would not exceed the maximum churn rate.
         // We only need to check this is the validator is still active. If the validator ends its validation
```
cam-schultz, line 410:
```suggestion
        // We only need to check this if the validator is still active. If the validator ends its validation
```

contracts/staking/PoSValidatorManager.sol
```diff
@@ -129,25 +135,52 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
         );
 
         if (includeUptimeProof) {
-            _getUptime(validationID, messageIndex);
+            // Uptime proofs include the absolute number of seconds the validator has been active.
+            uint64 uptimeSeconds = _getUptime(validationID, messageIndex);
+            // Save this value for use by this validator's delegators.
+            $._completedValidationUptimeSeconds[validationID] = uptimeSeconds;
+
+            $._redeemableValidatorRewards[validationID] += $._rewardCalculator.calculateReward({
```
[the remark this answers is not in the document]
geoff-vball: Yes. Validator's delegation fees will be credited here

contracts/staking/interfaces/IRewardCalculator.sol
```diff
@@ -8,8 +8,10 @@ pragma solidity 0.8.25;
...
+        uint64 stakingStartTime,
+        uint64 stakingEndTime,
+        uint64 uptimeSeconds,
         uint256 initialSupply,
         uint256 endSupply
```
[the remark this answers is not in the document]
geoff-vball: The current p-chain staking takes it into account https://docs.avax.network/nodes/validate/how-to-stake

cam-schultz (approved):

contracts/staking/PoSValidatorManager.sol
```diff
@@ -56,6 +56,12 @@ abstract contract PoSValidatorManager is
         mapping(bytes32 validationID => PoSValidatorRequirements) _validatorRequirements;
         /// @notice Maps the delegationID to the delegator information.
         mapping(bytes32 delegationID => Delegator) _delegatorStakes;
+        /// @notice Maps the delegationID to its pending staking rewards.
+        mapping(bytes32 delegationID => uint256) _redeemableDelegatorRewards;
+        /// @notice Maps the validator owner address to its pending staking rewards.
```
minghinmatthewlam, line 61: can you update this comment for the correct key value

contracts/staking/PoSValidatorManager.sol
```diff
@@ -136,26 +142,53 @@ abstract contract PoSValidatorManager is
...
         );
 
         if (includeUptimeProof) {
-            _getUptime(validationID, messageIndex);
+            // Uptime proofs include the absolute number of seconds the validator has been active.
+            uint64 uptimeSeconds = _getUptime(validationID, messageIndex);
+            // Save this value for use by this validator's delegators.
+            $._completedValidationUptimeSeconds[validationID] = uptimeSeconds;
+
+            $._redeemableValidatorRewards[validationID] += $._rewardCalculator.calculateReward({
```
minghinmatthewlam, line 161: think we should check for PoS validator here too, and never call the reward manager if it's not a PoS validator, can return early and save gas

contracts/mocks/ExampleERC20.sol
```diff
@@ -13,12 +13,13 @@ import {
...
 
-contract ExampleERC20 is ERC20Burnable {
+contract ExampleERC20 is ERC20Burnable, IERC20Mintable {
     string private constant _TOKEN_NAME = "Mock Token";
     string private constant _TOKEN_SYMBOL = "EXMP";
 
-    uint256 private constant _MAX_MINT = 1e16;
+    uint256 private constant _MAX_MINT = 1e19;
 
     constructor() ERC20(_TOKEN_NAME, _TOKEN_SYMBOL) {
```
minghinmatthewlam, line 24: thinking if this example ERC20 should only give access to mint to an ownable address. Even for an example seems like a pretty common check that would exist.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -136,26 +142,53 @@ abstract contract PoSValidatorManager is
...
         (bytes32 validationID, Validator memory validator) = _completeEndValidation(messageIndex);
-        _unlock(validator.startingWeight, $._validatorRequirements[validationID].owner);
+
+        // Return now if this was originally a PoA validator that was later migrated to this PoS manager
+        if (!_isPoSValidator(validationID)) {
+            return;
+        }
+
+        address owner = $._validatorRequirements[validationID].owner;
+        if (validator.status == ValidatorStatus.Completed) {
```
michaelkaplan13, line 184: In what case here would `validator.status` not be `Completed`? Could we add it as a comment

michaelkaplan13: Ah, I see below it's if the validator wasn't properly registered on the P-Chain prior to the message `expiry`. I'm wondering if we could remove the `Invalidated` status and just go off the `startTime` being zero in the contract, but not a priority right now.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -355,19 +381,36 @@ abstract contract PoSValidatorManager is
         // initialize the removal.
         delegator.status = DelegatorStatus.PendingRemoved;
 
+        uint64 validatorUptimeSeconds;
         if (validator.status == ValidatorStatus.Active) {
+            if (includeUptimeProof) {
```
michaelkaplan13, line 386: Nit: Can we add a @notice comment that `includeUptimeProof` can always be `false` if the validation that the `delegationID` delegated to has already ended since the uptime proof will have already been provided?

contracts/staking/PoSValidatorManager.sol
```diff
@@ -56,6 +56,12 @@ abstract contract PoSValidatorManager is
         mapping(bytes32 validationID => PoSValidatorRequirements) _validatorRequirements;
         /// @notice Maps the delegationID to the delegator information.
         mapping(bytes32 delegationID => Delegator) _delegatorStakes;
+        /// @notice Maps the delegationID to its pending staking rewards.
+        mapping(bytes32 delegationID => uint256) _redeemableDelegatorRewards;
+        /// @notice Maps the validator owner address to its pending staking rewards.
+        mapping(bytes32 validationID => uint256) _redeemableValidatorRewards;
+        /// @notice Saves the uptime of a pending completed or completed validation period so that delegators can collect rewards.
+        mapping(bytes32 validationID => uint64) _completedValidationUptimeSeconds;
```
michaelkaplan13, line 64: Are we able to clean up the entries in this map when the final delegation to a completed validation is also completed? Not sure if we have a counter on the number of delegations for a given validation or not as is.

contracts/mocks/ExampleERC20.sol
geoff-vball: I'm not quite sure what you mean

contracts/staking/PoSValidatorManager.sol
```diff
@@ -425,29 +470,29 @@ abstract contract PoSValidatorManager is
...
         // Update the delegator status
         $._delegatorStakes[delegationID].status = DelegatorStatus.Completed;
 
-        Delegator memory delegator = $._delegatorStakes[delegationID];
-        _unlock(delegator.weight, delegator.owner);
-        // TODO: issue rewards
+        _reward(delegator.owner, $._redeemableDelegatorRewards[delegationID]);
+        delete $._redeemableDelegatorRewards[delegationID];
+        _unlock(delegator.owner, weightToValue(delegator.weight));
+        // TODO can we remove the delegation from _delegatorStakes here?
```
michaelkaplan13: It seems like the answer here is yes. Once a delegationID's status is `Completed`, it shouldn't be able to be modified at all in the future right?

contracts/staking/PoSValidatorManager.sol
```diff
@@ -136,26 +142,53 @@ abstract contract PoSValidatorManager is
...
-        _initializeEndValidation(validationID);
     }
 
     function completeEndValidation(uint32 messageIndex) external {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
         (bytes32 validationID, Validator memory validator) = _completeEndValidation(messageIndex);
-        _unlock(validator.startingWeight, $._validatorRequirements[validationID].owner);
+
+        // Return now if this was originally a PoA validator that was later migrated to this PoS manager
```
iansuvak, line 201: nit: This can also be an initialValidator provided to PoSManager. In that case it is a PoA validator but there was no migration to PoS.

contracts/staking/PoSValidatorManager.sol
geoff-vball: Added to the doc in `IPoSValidatorManager.sol`

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,32 +136,83 @@ abstract contract PoSValidatorManager is
...
 
+    function claimDelegationFees(bytes32 validationID) external {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        Validator memory validator = getValidator(validationID);
+
+        require(
+            validator.status == ValidatorStatus.Completed,
+            "PoSValidatorManager: validation period not completed"
+        );
```
michaelkaplan13, line 147: Lets keep this requirement for now, but worth calling out it's not technically needed. It would be safe for validators to claim their delegation fees prior to the end of their validation period (it may have tax-related implications for people various jurisdictions though).

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,32 +136,83 @@ abstract contract PoSValidatorManager is
...
         bool includeUptimeProof,
         uint32 messageIndex
     ) external {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
-        // Check that minimum stake duration has passed
-        Validator memory validator = getValidator(validationID);
+
+        Validator memory validator = _initializeEndValidation(validationID);
+
+        if (!_isPoSValidator(validationID)) {
```
michaelkaplan13, line 167: Can we add a comment here explaining why non-PoS validators exist and why they are handled differently?

contracts/staking/PoSValidatorManager.sol
```diff
@@ -425,29 +501,36 @@ abstract contract PoSValidatorManager is
...
         // end delegation has been initialized, we need to make sure that this function is only
         // callable after that has been done.
         require(
-            $._delegatorStakes[delegationID].status == DelegatorStatus.PendingRemoved,
+            delegator.status == DelegatorStatus.PendingRemoved,
             "PoSValidatorManager: delegation not pending added"
         );
 
-        // Update the delegator status
-        $._delegatorStakes[delegationID].status = DelegatorStatus.Completed;
```
geoff-vball, line 442: I realized we don't actually need a `DelegatorStatus.Completed` since we can clear the delegator's state when we get to this stage. I can remove it from the enum if others thinks that makes sense

contracts/staking/PoSValidatorManager.sol
geoff-vball: Yeah, we talked about this on a call yesterday. If we want validators to be able to collect their fees before their validation period ends, we'd have to create a separate map for fees and rewards, because putting them in the same map would allow the validator to collect their rewards in the `PendingRemoved` state.

So we went for this simpler but more restrictive option for now.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,32 +136,84 @@ abstract contract PoSValidatorManager is
...
+
     function initializeEndValidation(
         bytes32 validationID,
         bool includeUptimeProof,
         uint32 messageIndex
     ) external {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
-        // Check that minimum stake duration has passed
-        Validator memory validator = getValidator(validationID);
+
```
michaelkaplan13, line 164: It looks like we're missing requirement that only the owner of the validation is able to call `initializeEndValidation` for PoS validators.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -425,29 +501,36 @@ abstract contract PoSValidatorManager is
...
+        // Allocate the delegation fees to the validator.
+        $._redeemableValidatorRewards[validationID] += validatorFees;
+
+        // Reward the remaining tokens to the delegator.
+        _reward(delegator.owner, rewards - validatorFees);
+
+        // Unlock the delegator's stake.
+        _unlock(delegator.owner, weightToValue(delegator.weight));
 
         emit DelegationEnded(delegationID, validationID, nonce);
```
michaelkaplan13, line 529: Should we include the delegator rewards and the delegator fee amount in this event?

contracts/staking/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -22,17 +23,25 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
...
+    uint16 public constant DEFAULT_DELEGATION_FEE_BIPS = 150;
     uint8 public constant DEFAULT_MAXIMUM_STAKE_MULTIPLIER = 4;
+    uint256 public constant SECONDS_IN_YEAR = 31536000;
 
     PoSValidatorManager public posValidatorManager;
+    IRewardCalculator public rewardCalculator;
 
     event ValidationUptimeUpdated(bytes32 indexed validationID, uint64 uptime);
 
+    event GeoffEvent(uint256 num);
```
minghinmatthewlam, line 43: todo: remove this

contracts/staking/PoSValidatorManager.sol
```diff
@@ -130,32 +136,91 @@ abstract contract PoSValidatorManager is
...
+
+        // Non-PoS validators are required to boostrap the network, but are not eligible for rewards.
+        if (!_isPoSValidator(validationID)) {
+            return;
+        }
+
+        // PoS validations can only be ended by their owners.
         require(
-            block.timestamp
+            $._validatorRequirements[validationID].owner == _msgSender(),
```
minghinmatthewlam, line 174: good catch on this, can you add unit tests for this?

geoff-vball: Added

michaelkaplan13 (approved):

PR #535, opened 2024-09-06
Update subnet-evm/avalanchego and remove hardcoded timeouts from tmpnet

iansuvak:
## Why this should be merged
Latest changes make tmpnet a bit slower to startup. To update to the newest versions and have the E2E tests pass we need to extend timeouts. Since this repo is used downstream in the `awm-relayer` repo as well I made `NewLocalNetwork` accept a context and let the caller handle setting timeouts. 
## How this was tested
CI
## How is this documented
N/A


geoff-vball (approved):

tests/local/network.go
```diff
@@ -128,9 +126,9 @@ func NewLocalNetwork(
 	avalancheGoBuildPath, ok := os.LookupEnv("AVALANCHEGO_BUILD_PATH")
 	Expect(ok).Should(Equal(true))
 
-	ctxWithTimeout, cancelBootstrap := context.WithTimeout(ctx, timeout)
+	ctx, cancelBootstrap := context.WithCancel(ctx)
```
feuGeneA, line 129: Not related to your change here (and I probably wrote this code myself 😅 ) but shouldn't the `defer` come _before_ the context is passed into `BootstrapNewNetwork`?

feuGeneA (approved):

tests/local/network.go
iansuvak: Indeed -- thanks for noticing

geoff-vball (approved):

Issue #525, opened 2024-09-05
Delegation: Unify resend methods


cam-schultz:
**Context and scope**
Following the pattern in `ValidatorManager` for re-sending validator registrations/removals, `PoSValidatorManager` has the methods `resendDelegatorRegistration` and `resendEndDelegation`. Instead, we could combine this into a single `sendCurrentValidatorWeight` that constructs a new Warp message containing the latest validator weight and nonce tracked in `ValidatorManager`.

This is safe to do because the methods `completeDelegatorRegistration` and `completeEndDelegation` will accept Warp messages from the P-Chain acknowledging the weight change with any nonce greater than or equal to the nonce used in the corresponding `initialize*` methods.

PR #536, opened 2024-09-06
Unify resend methods

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/525

## How this works
Uses the current state to resend messages, instead of caching them.

## How this was tested
Existing tests

## How is this documented
N/A


contracts/staking/PoSValidatorManager.sol
```diff
@@ -312,10 +301,27 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         });
     }
 
-    function resendEndDelegation(bytes32 delegationID) external {
-        _checkPendingEndDelegatorMessage(delegationID);
+    function resendUpdateDelegation(bytes32 delegationID) external {
```
michaelkaplan13, line 304: Can we add a comment here explaining why it's safe to always resend the latest nonce and weight even it it's not the nonce and weight that were originally sent for the given message?

michaelkaplan13 (approved):

iansuvak (approved):

Issue #439, opened 2024-07-22
Consider parameterizing staking dynamics


cam-schultz:
**Context and scope**
We should consider parameterizing min/max staking duration, min/max staking amount, and allowed validator churn rate in the Staking Manager contracts.

minghinmatthewlam:
Min staking duration is added in separate issue, max staking duration not needed
Min staking amount included, max staking amount include in this issue
Allowed churn rate is in separate issue
Min delegation should be included, max delegation is proportional to validator stake amount (up to x amount of validator stake amount).

PR #544, opened 2024-09-10
parameterize validator manager values

### LINKED ISSUE #507: Delegation: Extensions to specify min validator duration
**Context and scope**
To improve the UX for delegators, we should add an extension or optional feature to `PoSValidatorManager` that supports specifying the minimum staking duration. This will allow potential delegators to select validators based on how actively they wish to maintain their delegation. For instance, a delegator that wants to "set and forget" may choose a validator with a relatively long minimum validation duration. 

- `initializeEndValidation` should not be able to be called until after the minimum duration has elapsed.

**Open questions**
A related idea is to add a method for validators to extend the minimum validation duration without having to end their validation period and start a new one.

minghinmatthewlam:
## Why this should be merged
Fixes #507 and #439. 

## How this works
For all PoS validators:
- minimum delegation fee
- maximum stake multiplier, multiplied with maximum stake amount. This value denotes maximum stake amount that be attributable to a validator, including initial stake amount and all delegations

Individual PoS validator options:
- min staking duration promised by the validator, the validator cannot `initializeEndValidation` until that time has passed.
- delegation fee the validator needs for delegations

## How this was tested
unit tests

## How is this documented
comments


contracts/validatormanager/PoSValidatorManager.sol
```diff
@@ -73,12 +92,23 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        uint64 maximumStakeMultiplier,
         IRewardCalculator rewardCalculator
     ) internal onlyInitializing {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        require(minimumDelegationFee > 0, "PoSValidatorManager: zero delegation fee");
+        require(
+            minimumStakeAmount <= maximumStakeAmount,
+            "PoSValidatorManager: invalid stake amount range"
+        );
+        require(maximumStakeMultiplier > 0, "PoSValidatorManager: zero maximum stake multiplier");
```
michaelkaplan13, line 105: Should we support 0 as a valid value here? That would effectively disable any delegation, which could be a reasonable use case. 

I think we should enforce some maximum as well. If someone really wants to go above the maximum, it's easy to change, but otherwise could result in them unintentionally messing up the value used in the constructor. Just by gut feeling, I'd say a reasonable maximum `maximumStakeMultipier` to enforce would be 10.

geoff-vball:
@minghinmatthewlam Can you move the folder renaming to it's own PR? With additional changes on top, it's going to make merge conflicts really tough

contracts/staking/PoSValidatorManager.sol
```diff
@@ -27,8 +34,20 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         uint256 _maximumStakeAmount;
         /// @notice The minimum amount of time a validator must be staked for.
         uint64 _minimumStakeDuration;
+        /// @notice The minimum delegation fee required to delegate to a validator.
+        uint256 _minimumDelegationFee;
```
bernard-avalabs, line 38: Delegation fee on the p-chain is a percentage of the staking award. Is minimumDelegationFee also a percentage, or is it a fixed fee?

bernard-avalabs (commented):
I left a few questions about how we calculate maximum stake weight. Perhaps I'm misunderstanding something, but it appears to be different from how the maximums are calculated on the p-chain.

contracts/staking/PoSValidatorManager.sol
minghinmatthewlam: updated to take in bips, meant to be a percentage

contracts/staking/PoSValidatorManager.sol
```diff
@@ -164,12 +226,16 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
         );
         // Update the validator weight
         uint64 newValidatorWeight = validator.weight + weight;
+        require(
+            weightToValue(newValidatorWeight) <= $._maximumStakeAmount * $._maximumStakeMultiplier,
```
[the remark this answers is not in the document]
minghinmatthewlam: same as above

contracts/staking/PoSValidatorManager.sol
```diff
@@ -61,25 +85,41 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
     function __POS_Validator_Manager_init_unchained(
         uint256 minimumStakeAmount,
         uint256 maximumStakeAmount,
         uint64 minimumStakeDuration,
+        uint256 minimumDelegationFeeBips,
+        uint8 maximumStakeMultiplier,
         IRewardCalculator rewardCalculator
     ) internal onlyInitializing {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        require(minimumDelegationFeeBips > 0, "PoSValidatorManager: zero delegation fee");
```
iansuvak, line 108: we should also require that `minimumDelegationFeeBips < MAXIMUM_DELEGATION_FEE_BIPS` otherwise you can initialize a manager that can't add any validators

minghinmatthewlam: good point, added

contracts/staking/PoSValidatorManager.sol
```diff
@@ -39,6 +58,11 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
     bytes32 private constant _POS_VALIDATOR_MANAGER_STORAGE_LOCATION =
         0x4317713f7ecbdddd4bc99e95d903adedaa883b2e7c2551610bd13e2c7e473d00;
 
+    uint8 public constant MAXIMUM_STAKE_MULTIPLIER_LIMIT = 10;
+
+    // TODO: revisit maximum, currently set to 100%
```
geoff-vball, line 63: I think 100% makes sense. A validator would set this if they only want to delegate to themselves

contracts/staking/PoSValidatorManager.sol
```diff
@@ -57,25 +81,45 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            "PoSValidatorManager: invalid delegation fee"
+        );
+        require(
+            minimumStakeAmount <= maximumStakeAmount,
+            "PoSValidatorManager: invalid stake amount range"
+        );
+        require(
+            maximumStakeMultiplier <= MAXIMUM_STAKE_MULTIPLIER_LIMIT && maximumStakeMultiplier > 0,
+            "PoSValidatorManager: invalid maximum stake multiplier"
+        );
```
geoff-vball, line 116: Can we split this into two separate require statements? I find it much easier to debug if I know which condition is failing.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -120,18 +173,34 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+        ValidatorRegistrationInput calldata registrationInput,
+        PoSValidatorRequirements calldata requirements,
+        uint256 stakeAmount
+    ) internal virtual returns (bytes32) {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        // Validate and save the validator requirements
+        require(
+            requirements.minStakeDuration >= $._minimumStakeDuration,
+            "PoSValidatorManager: invalid min stake duration"
+        );
```
geoff-vball, line 186: I don't know if we should have to check this. I think the validator settings it's own `minStakeDuration` to 0 should be valid if they don't care, we will just still enforce `_minimumStakeDuration` anyway.

geoff-vball (commented):
Mostly LGTM, just a few small questions

contracts/staking/PoSValidatorManager.sol
minghinmatthewlam: I think validators should know the `minStakeDuration` they're required before their funds get locked. Wouldn't want cases where validators dont know of the min stake duration then get caught off guard when initializing end, and probably perform badly if they aren't happy with the stake duration needed.

contracts/validatormanager/PoSValidatorManager.sol
minghinmatthewlam: added maximum, since the multiplier is applied to inital stake amount the value of 1 already disables any further delegation, so 0 is not needed.

iansuvak (approved):

Issue #542, opened 2024-09-10
validate initial validator set from ConvertSubnetTx


minghinmatthewlam:
**Context and scope**
Update to ACP77, will have `ConvertSubnetTx` specify an initial set of PAYG validators, that can only be removed from the validator manager. The idea is for validator manager contract to accept an initial validator set that was specified in `ConvertSubnetTx`, and the contract would not be able to initialize other validators until this process is done.

**Discussion and alternatives**
For verification, there might be a new P-Chain message type that includes all these new validators. The validator manager contract should get this verified Warp message, verify the message, and parse for the initial validator set. Then the contract needs to compute new validation IDs to associate these validators, which can look like `sha256(convertSubnetTxID,validatorIndex)`. 

**Open questions**
Will there be a separate function to initialize this validator set, or can it be done through contract construction in the `initialize` functions? The caveat being migration from PoA->PoS, we need to make sure that the `reinitializer` of the PoS contracts can not specify another initial validator set that was already specified as part of PoA contract initially.

meaghanfitzgerald:
will `initializeValidatorSet`  in https://github.com/ava-labs/teleporter/pull/547 need to be called manually by the creator of the L1 or will the initial validator set be added to the set tracked by the ValidatorManager via the contructor?

minghinmatthewlam:
> will `initializeValidatorSet` in #547 need to be called manually by the creator of the L1 or will the initial validator set be added to the set tracked by the ValidatorManager via the contructor?

will ned to be called separately. At genesis we can still pass the parameters required for `initialize`, but at genesis we do not have the subnet conversion tx ID, which is found after `ConvertSubnetTx`.

cam-schultz:
Completed by https://github.com/ava-labs/teleporter/pull/547

PR #547, opened 2024-09-12
Initial validator sets

### LINKED ISSUE #543: initial validator set ending and rewards
**Context and scope**
Relates to #542 , we need special handling for the initial validator set, in terms of who can initialize and complete validation for these validators, and making sure no rewards can be given to them.

For PoA, the owner should be able to initialize end validation for the validator set, and likely anyone can complete end validation similar to other validators.

For PoS, there's no original sender that initialized validator registration, so should anyone be able to initialize end validation here? Does it make sense to have a minimum time before these validators can be ended?

minghinmatthewlam:
## Why this should be merged
Fixes #543 and #542 

## How this works
- Adds a separate `initializeValidatorSet` call which expects a Warp message that contains a `SubnetConversionMessage` in its payload, and then verifies the tx ID against the hash of the bytes for the passed in `SubnetConversionData` with the initial validators.
- Restricts initial valiator registration until initial validator set is set, and implicitly has checked for other calls to not be callable until `initializedValidatorSet`
- Removes `owner` from the base struct of `Validator`, any PoA validator either added from initial validator set, or from a PoA validator manager contract can be intiialized end by anyone after migration to PoS. 
- For PoA, only owner can call intiialize registration/end for any validators, fixes a bug where after owner change the new owner doesn't have access.
- PoS validator manager contract keeps track of its added validations, so it knows which validations are PoS and eligibile for rewards, etc. 
- Reentrancy guard is moved to only PoS, because PoA doesn't have external calls to lock/unlock stake, only calls to Warp.

## How this was tested
TODO

## How is this documented
comments


contracts/staking/interfaces/IPoSValidatorManager.sol
```diff
@@ -39,6 +39,7 @@ struct Delegator {
 }
 
 struct PoSValidatorRequirements {
+    address owner;
```
geoff-vball, line 42: If we're adding `owner` here, we should probably name this something other than `requirements`

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,104 @@ abstract contract ValidatorManager is
...
+        uint256 length = encodedConversion.length;
+        for (uint256 i; i < length; i++) {
+            InitialValidator memory initialValidator = subnetConversionData.initialValidators[i];
+            encodedConversion = abi.encodePacked(
+                encodedConversion,
+                initialValidator.nodeID,
+                initialValidator.weight,
+                initialValidator.blsPublicKey
+            );
+        }
```
geoff-vball, line 134: This doesn't match up with the serialization in https://github.com/avalanche-foundation/ACPs/pull/146/files. I think the following should work fine?
```suggestion
        bytes memory encodedConversion = abi.encodePacked(subnetConversionData);
```

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,104 @@ abstract contract ValidatorManager is
...
+            address(bytes20(subnetConversionData.validatorManagerAddress)) == address(this),
+            "ValidatorManager: invalid validator manager address"
+        );
+        require(
+            subnetConversionData.initialValidators.length > 0,
+            "ValidatorManager: no initial validators"
+        );
+
+        // Insert the initial validators into the storage.
+        for (uint256 i; i < length; i++) {
```
geoff-vball, line 159: This `length` is not at all the right length to be using here

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,104 @@ abstract contract ValidatorManager is
...
+                status: ValidatorStatus.PendingAdded,
+                nodeID: initialValidator.nodeID,
+                startingWeight: initialValidator.weight,
+                messageNonce: 0,
+                weight: initialValidator.weight,
+                startedAt: 0, // The validation period only starts once the registration is acknowledged.
+                endedAt: 0
+            });
+            // Increment the nonce since these validations are completed.
+            _incrementAndGetNonce(validationID);
```
geoff-vball, line 185: No need to increment the nonce here

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,104 @@ abstract contract ValidatorManager is
...
+                "ValidatorManager: node ID already active"
+            );
+            require(
+                initialValidator.blsPublicKey.length == 48,
+                "ValidatorManager: invalid blsPublicKey length"
+            );
+
+            $._activeValidators[nodeID] = validationID;
+            $._validationPeriods[validationID] = Validator({
+                status: ValidatorStatus.PendingAdded,
```
geoff-vball, line 176: This should be `ValidatorStatus.Active`

geoff-vball (commented):
I know this is still WIP but wanted to leave feedback as early as possible

contracts/staking/ValidatorManager.sol
minghinmatthewlam: as in how it was set right? Updated to be `subnetConversionData.initialValidators.length`

contracts/staking/ValidatorManager.sol
minghinmatthewlam: can't pass a struct to encode packed, which part doesn't match up?

contracts/staking/ValidatorManager.sol
geoff-vball: Yeah, that's right

contracts/staking/ValidatorManager.sol
geoff-vball: Ahhh okay... So my understanding (could totally be wrong) is that you need to pack an entire array at once, because the serialization is going to prepend the serialized array with its length, which is needed to deserialize correctly. You'll see that in the ACP, each array has a static 4 bytes in size in addition to the size taken by the fields, which I assume is for the length.

minghinmatthewlam: ah yes you might be right, need to double check on the additional 4 bytes here

contracts/staking/ValidatorMessages.sol
```diff
@@ -19,26 +19,27 @@ library ValidatorMessages {
     // The P-Chain uses a hardcoded codecID of 0 for all messages.
     uint16 internal constant CODEC_ID = 0;
 
+    // The P-Chain signs a SubnetConversion message that is used to verify for the Subnet's initial validators.
+    uint32 internal constant SUBNET_CONVERSION_MESSAGE_TYPE_ID = 5;
```
iansuvak, line 23:
```suggestion
    uint32 internal constant SUBNET_CONVERSION_MESSAGE_TYPE_ID = 0;
```

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -41,7 +41,7 @@ struct ValidatorManagerSettings {
 struct SubnetConversionData {
     bytes32 convertSubnetTxID;
     bytes32 blockchainID;
-    bytes validatorManagerAddress;
+    address validatorManagerAddress;
```
iansuvak, line 44: Should we leave this as bytes as that's what spec defines it as in terms of encoding? It's not fixed 20 bytes but 4 bytes for len + `len(validationManagerAddress)` That would make any automated packing/unpacking easier and then can convert to address at usage time.

contracts/staking/ValidatorManager.sol
```diff
@@ -110,56 +110,33 @@ abstract contract ValidatorManager is Initializable, ContextUpgradeable, IValida
...
+        require(
+            numInitialValidators < type(uint32).max, "ValidatorManager: too many initial validators"
+        );
 
-        // Insert the initial validators into the storage.
+        // Verify that the sha256 hash of the Subnet conversion data matches with the Warp message's subnetConversionID.
+        bytes memory encodedConversion = abi.encodePacked(
+            subnetConversionData.convertSubnetTxID,
+            subnetConversionData.blockchainID,
+            uint32(20),
```
iansuvak, line 134: ah I see that you are hardcoding length here instead of doing the bytes packing, I'm not sure what is cleaner long term

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -99,6 +117,17 @@ interface IValidatorManager {
      */
     event ValidationPeriodEnded(bytes32 indexed validationID, ValidatorStatus indexed status);
 
+    /**
+     * @notice Verifies and sets the initial validator set for the chain through a P-Chain
+     * SubnetConversionMessage.
+     * @param subnetConversionData The subnet conversion message data used to recompute and verify against the subnetConversionID.
+     * @param messsageIndex The storage slot index that contains the SubnetConversionMessage Warp message.
```
michaelkaplan13, line 124:
```suggestion
     * @param messsageIndex The index that contains the SubnetConversionMessage Warp message in the transaction.
```

Nit: Messages are encoded in the transactions access list, not storage

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,105 @@ abstract contract ValidatorManager is
...
+
+        // Verify that the sha256 hash of the Subnet conversion data matches with the Warp message's subnetConversionID.
+        bytes memory encodedConversion = abi.encodePacked(
+            subnetConversionData.convertSubnetTxID,
+            subnetConversionData.blockchainID,
+            uint32(20),
+            subnetConversionData.validatorManagerAddress,
+            uint32(numInitialValidators)
+        );
+        for (uint256 i; i < numInitialValidators; i++) {
```
michaelkaplan13, line 138:
```suggestion
        for (uint256 i; i < numInitialValidators; ++i) {
```

Nit: Pre-increment is generally slightly more gas efficient for for loops since it results in one few store operation. Not sure if we've gone with post-increment in the past for any specific reason though.

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,105 @@ abstract contract ValidatorManager is
...
+                $._activeValidators[nodeID] == bytes32(0),
+                "ValidatorManager: node ID already active"
+            );
+            require(
+                initialValidator.blsPublicKey.length == 48,
+                "ValidatorManager: invalid blsPublicKey length"
+            );
+
+            // Continue to encode the initial validators.
+            encodedConversion = abi.encodePacked(
```
michaelkaplan13, line 152: I think this results in a copy of the full `encodedConversion` bytes up until this point. It'd probably be a bit cleaner and more efficient to pre-allocate an array of the necessary size upfront once, and then assign into it as we go along.

contracts/staking/ValidatorManager.sol
```diff
@@ -93,6 +90,105 @@ abstract contract ValidatorManager is
...
+                encodedConversion,
+                initialValidator.nodeID,
+                initialValidator.weight,
+                initialValidator.blsPublicKey
+            );
+
+            // Validation ID of the initial validators is the sha256 hash of the
+            // convert Subnet tx ID and the index of the initial validator.
+            bytes32 validationID =
+                sha256(abi.encodePacked(subnetConversionData.convertSubnetTxID, i));
```
michaelkaplan13, line 162: We'll need to make sure `i` here is represented as 4 bytes (uint32) in the pre-image. I'm not sure if `encodePacked` would handle that properly if we use a `uint32` instead of ` uint256`, or if we have to pack it manually.

contracts/staking/ValidatorManager.sol
```diff
@@ -278,7 +376,10 @@ abstract contract ValidatorManager is
      * {registrationExpiry} being reached.
      * @return The Validator instance representing the completed validation period
      */
-    function _completeEndValidation(uint32 messageIndex) internal returns (Validator memory) {
+    function _completeEndValidation(uint32 messageIndex)
+        internal
+        returns (bytes32, Validator memory)
```
michaelkaplan13, line 381: We should update the `@return` comment above

contracts/staking/ValidatorManager.sol
iansuvak: Using `uint32` as iterator is sufficient. I made the change in #553

contracts/staking/ValidatorManager.sol
minghinmatthewlam: updated

contracts/staking/ValidatorManager.sol
minghinmatthewlam: ye makes sense, I think we just forgot at times previously

contracts/staking/ERC20TokenStakingManager.sol
```diff
@@ -107,7 +107,7 @@ contract ERC20TokenStakingManager is
     function initializeDelegatorRegistration(
         bytes32 validationID,
         uint256 delegationAmount
-    ) external returns (bytes32) {
+    ) external nonReentrant returns (bytes32) {
```
cam-schultz, line 110: For my own understanding, is there an advantage to moving `nonReentrant` here rather than in `_initializeDelegatorRegistration`?

contracts/staking/ValidatorMessages.sol
```diff
@@ -19,28 +19,81 @@ library ValidatorMessages {
     // The P-Chain uses a hardcoded codecID of 0 for all messages.
     uint16 internal constant CODEC_ID = 0;
 
+    // The P-Chain signs a SubnetConversion message that is used to verify for the Subnet's initial validators.
```
cam-schultz, line 22:
```suggestion
    // The P-Chain signs a SubnetConversion message that is used to verify the Subnet's initial validators.
```
grammar nit

contracts/staking/ValidatorMessages.sol
```diff
@@ -19,28 +19,81 @@ library ValidatorMessages {
...
 
     // The Subnet will self-sign a ValidationUptimeMessage to be provided when a validator is initiating
     // the end of their validation period.
-    uint32 internal constant VALIDATION_UPTIME_MESSAGE_TYPE_ID = 4;
+    uint32 internal constant VALIDATION_UPTIME_MESSAGE_TYPE_ID = 5;
 
     // TODO: The implementation of these packing and unpacking functions is neither tested nor optimized at all.
     // Full test coverage should be provided, and the implementation should be optimized for gas efficiency.
 
+    function packSubnetConversionMessage(bytes32 subnetConversionID)
```
cam-schultz, line 47: Can you add natspec comments to these methods to match the others? The message specs are especially useful.

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -39,6 +38,19 @@ struct ValidatorManagerSettings {
     uint8 maximumChurnPercentage;
 }
 
+struct SubnetConversionData {
+    bytes32 convertSubnetTxID;
+    bytes32 blockchainID;
```
cam-schultz, line 43: Can you change this to `validatorManagerBlockchainID` to make it more explicit?

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -39,6 +38,19 @@ struct ValidatorManagerSettings {
...
+    address validatorManagerAddress;
+    InitialValidator[] initialValidators;
+}
+
+struct InitialValidator {
```
cam-schultz, line 48: I think we should have the name here match the spec. The spec [update](https://github.com/avalanche-foundation/ACPs/pull/146/files) calls this `ValidatorData`. Between the two, I prefer `InitialValidator`, unless there's another scenario in which `ValidatorData` would be used outside of initialization. @michaelkaplan13

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -101,6 +117,17 @@ interface IValidatorManager {
      */
     event ValidationPeriodEnded(bytes32 indexed validationID, ValidatorStatus indexed status);
 
+    /**
+     * @notice Verifies and sets the initial validator set for the chain through a P-Chain
+     * SubnetConversionMessage.
+     * @param subnetConversionData The subnet conversion message data used to recompute and verify against the subnetConversionID.
+     * @param messsageIndex The index that contains the SubnetConversionMessage Warp message in the transaction.
```
cam-schultz, line 124:
```suggestion
     * @param messsageIndex The index that contains the SubnetConversionMessage Warp message containing the subnetConversionID to be verified against the provided {subnetConversionData}
```
Small suggestion to link the `subnetConversionData` param with the Warp message retreived from `messageIndex`.

contracts/staking/ValidatorManager.sol
```diff
@@ -108,10 +105,108 @@ abstract contract ValidatorManager is
...
+            subnetConversionData.blockchainID,
+            uint32(20),
+            subnetConversionData.validatorManagerAddress,
+            uint32(numInitialValidators)
+        );
+        uint256 totalWeight;
+        for (uint32 i; i < numInitialValidators; ++i) {
+            InitialValidator memory initialValidator = subnetConversionData.initialValidators[i];
+            bytes32 nodeID = initialValidator.nodeID;
+            require(nodeID != bytes32(0), "ValidatorManager: invalid node ID");
```
cam-schultz, line 156: Is this check strictly necessary? IMO node ID validity should be checked on the P-Chain side, so if the P-Chain is willing to sign it we should accept it. A zero node ID would cause the hash comparison to fail.

cam-schultz: Expanding on this, one could argue _all_ the validity checks on `subnetConversionData` are unnecessary since the `subnetConversionID` from the Warp message already encodes validated data (by the P-Chain). If the caller supplies invalid `subnetConversionData` here, the hash comparison will fail regardless, and the result would be the same as if we checks the input's validity.

contracts/staking/ValidatorManager.sol
```diff
@@ -122,7 +217,7 @@ abstract contract ValidatorManager is
     function _initializeValidatorRegistration(
```
cam-schultz, line 217: Can you add a comment here explaining why this is the only validator registration/delisting function that needs the `initializedValidatorSet` modifier? I know you explained why in comments in those functions, but I think it would be helpful to have that more visible too.

cam-schultz (commented):
Generally makes sense to me. I left a handful of clarifying questions and cleanup suggestions.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -453,4 +471,9 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
 
         emit DelegationEnded(delegationID, validationID, nonce);
     }
+
+    function _isPoSValidator(bytes32 validationID) internal view returns (bool) {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        return $._validatorRequirements[validationID].owner != address(0);
```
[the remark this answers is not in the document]
geoff-vball: Yep

contracts/staking/ERC20TokenStakingManager.sol
geoff-vball: The idiomatic pattern is to place `nonReentrant` on all external functions. That way a contract can never be called back into by unknown code in an external contract, but can still call its own internal functions without worry.

cam-schultz: Makes sense, thanks for the explanation.

bernard-avalabs (commented):
Generally makes sense to me.

contracts/staking/ValidatorManager.sol
iansuvak: I agree and would be fine removing all of these checks.

contracts/staking/ValidatorManager.sol
```diff
@@ -108,10 +105,108 @@ abstract contract ValidatorManager is
...
+            "ValidatorManager: invalid validator manager address"
+        );
+        require(
+            subnetConversionData.initialValidators.length > 0,
+            "ValidatorManager: no initial validators"
+        );
+        uint256 numInitialValidators = subnetConversionData.initialValidators.length;
+        require(
+            numInitialValidators < type(uint32).max, "ValidatorManager: too many initial validators"
+        );
```
geoff-vball, line 142: Do we really have to check that we haven't included over 4 billion validators?

cam-schultz: IMO this can be removed along with the other checks mentioned [here](https://github.com/ava-labs/teleporter/pull/547#discussion_r1764103545), as long as we assume the P-Chain will enforce sane initial validator set size limits

iansuvak (approved):

PR #550, opened 2024-09-13
min stake duration fix

iansuvak:
## Why this should be merged

This fixes issues with minstakeduration not elapsing by forcing a block production between registration and ending by sending a loopback native transfer tx.

## How this was tested

ERC20 and Native e2e tests should now pass. 

## How is this documented


minghinmatthewlam (approved):
thanks for fix

PR #553, opened 2024-09-17
Initial validators ending e2e

iansuvak:
## Why this should be merged

Provides a pattern for calling `initializeValidatiorSet` as part of E2E testing. 

This will probably interact with the initial churn tracker stuff and we should make sure that the initial validators we are adding have reasonable weight. 

TODO:
- [x] Native 
- [x] ERC20
- [x] PoA->PoS case

## How this works
- References new AvaGo commit defining a new P-chain signable message 
- Adds manual packing for any subnetconversionData struct using reflect
- Adds helper functions in utils for PoA

## How this was tested
All E2E tests should pass except for 
Poa -> PoS conversion which just passes the step until PoS

All tests right now just add a single validator, we should add a test focusing on adding multiple and removing them. 
## How is this documented


geoff-vball (approved):
LGTM aside from Matt's comments

tests/flows/staking/poa_to_pos.go
```diff
@@ -185,16 +197,7 @@ func PoAMigrationToPoS(network interfaces.LocalNetwork) {
 		stakeAmount,
 	)
 
-	// Attempt to delist previous PoA validator with wrong owner and check that it fails
```
iansuvak, line 188: Anyone can delist PoA validators after conversion now so this test no longer applies

contracts/staking/ValidatorManager.sol
```diff
@@ -150,9 +149,8 @@ abstract contract ValidatorManager is Initializable, ContextUpgradeable, IValida
             subnetConversionData.validatorManagerAddress,
             uint32(numInitialValidators)
         );
-
         uint256 totalWeight;
-        for (uint256 i; i < numInitialValidators; ++i) {
+        for (uint32 i; i < numInitialValidators; i++) {
```
minghinmatthewlam, line 153: think this was by accident for the i++ change

minghinmatthewlam (approved):
LGTM small increment variable change

PR #554, opened 2024-09-17
Account for validation ending before delegation

geoff-vball:
## Why this should be merged
Accounts on the delegation side for the validation ending before the delegation.

## How this works
`initializeEndDelegation` has different behaviour depending on if the validationID for a given delegation is still active or not.

Also splits `DelegatorRemovalInitialized` event into two events, now with `ValidatorWeightUpdate` only being emitted if the validation period is active.

## How this was tested
TODO

## How is this documented
TODO


contracts/staking/PoSValidatorManager.sol
```diff
@@ -337,52 +337,57 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
...
+            _setValidatorWeight(validationID, newValidatorWeight);
+
+            // Submit the message to the Warp precompile.
+            bytes32 messageID = WARP_MESSENGER.sendWarpMessage(
+                ValidatorMessages.packSetSubnetValidatorWeightMessage(
+                    validationID, delegator.endingNonce, newValidatorWeight
+                )
+            );
+
+            emit ValidatorWeightUpdate({
```
michaelkaplan13, line 374: Would it make sense to move this emission to inside the `_setValidatorWeight` implementation? It may require removing the nonce/messageID fields, but feels a bit less error prone (as is, I think we're forgetting to emit it in the `initializeDelegatorRegistration` case).

geoff-vball: I've moved move of the weigh change logic to `_setValidatorWeight`, including checking the churn tracker. The only other time the churn tracker needs to be called is on validator registration.

contracts/staking/PoSValidatorManager.sol
```diff
@@ -337,52 +325,36 @@ abstract contract PoSValidatorManager is IPoSValidatorManager, ValidatorManager
 
         // Ensure the delegator is active
         Delegator memory delegator = $._delegatorStakes[delegationID];
+        Validator memory validator = getValidator(validationID);
```
michaelkaplan13, line 328: If the validation period has already ended, there won't be any uptime message able to be provided in the call to `initializeEndDelegation`, and the uptime submitted when the validator was ended should be used to calculate the delegators rewards.

contracts/staking/interfaces/IValidatorManager.sol
```diff
@@ -32,6 +32,20 @@ struct ValidatorChurnPeriod {
     uint64 churnAmount;
 }
 
+/**
+ * @notice Event emitted when delegator removal is initiated
```
cam-schultz, line 36: Can you make the `@notice` comment more generic? It applies to both delegation registration and removal, but IMO we don't need to mention delegation at all here.

contracts/staking/ValidatorManager.sol
```diff
@@ -375,46 +369,63 @@ abstract contract ValidatorManager is
...
+        return (nonce, messageID);
     }
 
     /**
      * @dev Helper function to check if the stake weight to be added or removed would exceed the maximum stake churn
      * rate for the past churn period. If the churn rate is exceeded, the function will revert. If the churn rate is
      * not exceeded, the function will update the churn tracker with the new weight.
      */
-    function _checkAndUpdateChurnTracker(uint64 weight, bool addition) private {
+    function _checkAndUpdateChurnTracker(uint64 newWeight, uint64 oldWeight) private {
```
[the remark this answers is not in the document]
geoff-vball: Yeah, renamed

cam-schultz (approved):

PR #561, opened 2024-09-19
Permissionless uptime submission

### LINKED ISSUE #557: Validators calling initializeEndValidation are required to submit an uptime proof
**Context and scope**
We have decided that validators must provide *some* valid uptime proof in order to end their validation. This is because other validators will stop tracking uptime for a validator that has called `initializeEndValidation`. To stop a validator from accidentally or maliciously not providing an uptime proof, we will require one. This proof will not be required to hit any sort of threshold for uptime, because even validators with low uptime should be able to exit the system.

**Open questions**
This may break current testing flows and require a larger refactor.

cam-schultz:
## Why this should be merged
Fixes #557 by implementing the spec in this [comment](https://github.com/ava-labs/teleporter/issues/557#issuecomment-2358725245).

~~TODO: `ERC20TokenStakingManager` exceeds the 24kB contract size with default optimization, causing the e2e tests to fail. This needs to be addressed before this PR is mergeable.~~ Addressed by https://github.com/ava-labs/teleporter/pull/566

## How this works
- Adds the function `submitUptimeProof` that allows anybody to submit an uptime proof for a validation. The uptime is stored in the contract if it is greater than the current greatest uptime.
- Splits `initializeEndValidation` into `initializeEndValidation` and `forceInitializeEndValidation`. The former reverts if the current uptime is not sufficient for rewards; the latter does not revert in this case.

## How this was tested
- Added unit tests
- CI

## How is this documented
N/A


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -175,22 +186,60 @@ abstract contract PoSValidatorManager is
...
+            stakingEndTime: validator.endedAt,
+            uptimeSeconds: uptimeSeconds,
+            initialSupply: 0,
+            endSupply: 0
+        });
+        require(reward > 0, "PoSValidatorManager: validation ineligble for rewards");
+        $._redeemableValidatorRewards[validationID] += reward;
+    }
+
+    function forceInitializeEndValidation(
```
geoff-vball, line 210: We could combine this and `initializeEndValidation` by making an internal function that returns `reward`, and then checking it in one but not the other.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -235,6 +286,15 @@ abstract contract PoSValidatorManager is
...
         );
 
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+        if (uptime > $._validationUptimeSeconds[validationID]) {
+            $._validationUptimeSeconds[validationID] = uptime;
+        } else {
+            uptime = $._validationUptimeSeconds[validationID];
+        }
+
+        emit UptimeUpdated(validationID, uptime);
```
geoff-vball, line 296: We should maybe only emit this in the `uptime > $._validationUptimeSeconds[validationID]` case?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -59,8 +59,8 @@ abstract contract PoSValidatorManager is
         mapping(bytes32 delegationID => uint256) _redeemableDelegatorRewards;
         /// @notice Maps the validation ID to its pending staking rewards.
         mapping(bytes32 validationID => uint256) _redeemableValidatorRewards;
-        /// @notice Saves the uptime of a pending completed or completed validation period so that delegators can collect rewards.
-        mapping(bytes32 validationID => uint64) _completedValidationUptimeSeconds;
+        /// @notice Saves the largest known uptime of an active, pending completed, or completed validation period so that delegators can collect rewards.
+        mapping(bytes32 validationID => uint64) _validationUptimeSeconds;
```
geoff-vball, line 63: I think we should move this into `_posValidatorInfo`

tests/utils/validator_manager.go
```diff
@@ -905,15 +916,26 @@ func InitializeEndERC20Validation(
 	subnet interfaces.SubnetTestInfo,
 	stakingManager *erc20tokenstakingmanager.ERC20TokenStakingManager,
 	validationID ids.ID,
+	force bool,
```
geoff-vball, line 919: For a function this short, I have a slight preference for splitting it into two different functions to call the different underlying functions. This makes it more clear what the caller is doing instead of having a list of booleans passed into a function.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -165,13 +176,32 @@ abstract contract PoSValidatorManager is
...
+        uint32 messageIndex
+    ) internal returns (uint256) {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
 
         Validator memory validator = _initializeEndValidation(validationID);
 
         // Non-PoS validators are required to boostrap the network, but are not eligible for rewards.
         if (!_isPoSValidator(validationID)) {
-            return;
+            revert ValidatorNotPoS();
```
geoff-vball, line 204: I'm pretty sure this leaves no route for the initial validators to be removed from the validator set. The flow was the call to `_initializeEndValidation` above, and then returning so none of the rewards calculations or anything were dealt with.

cam-schultz: Good catch. This was causing the E2E tests to fail. Fixed now.

tests/utils/validator_manager.go
cam-schultz: Agreed.

geoff-vball (approved):
One nit, otherwise LGTM

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -165,13 +174,31 @@ abstract contract PoSValidatorManager is
         bool includeUptimeProof,
         uint32 messageIndex
     ) external {
+        if (!_initializeEndPoSValidation(validationID, includeUptimeProof, messageIndex)) {
+            revert ValidatorIneligibleForRewards();
+        }
+    }
+
+    function forceInitializeEndValidation(
```
minghinmatthewlam, line 182: can youy add natspec comments

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -165,13 +174,31 @@ abstract contract PoSValidatorManager is
...
+        uint32 messageIndex
+    ) internal returns (bool) {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
 
         Validator memory validator = _initializeEndValidation(validationID);
 
         // Non-PoS validators are required to boostrap the network, but are not eligible for rewards.
         if (!_isPoSValidator(validationID)) {
-            return;
+            return true;
```
minghinmatthewlam, line 201: why do we return `true` here if below it returns for `reward > 0`? For non PoS validator shouldn't it return 0 rewards

minghinmatthewlam: can we add unit tests to cover these cases, for example a non PoSvalidator that is claiming delegation fees

contracts/validator-manager/PoSValidatorManager.sol
cam-schultz: Natspec comments are included in the interface definition.

minghinmatthewlam: generally still have a {See Interface-method} comment for functions of an interface

contracts/validator-manager/PoSValidatorManager.sol
cam-schultz: This function returns `false` if 1) it's a PoS validator and 2) it's not able to claim rewards based on the submitted uptime. I switched this function to return a bool rather than the rewards, since in the call to `initializeEndValidation`, `0` rewards should revert for PoS validators, but not PoA.

cam-schultz: Unit testing this is not straightforward, since it requires a PoA -> PoS migration. I think this is better left for the e2e tests for now. In any case, I'm going to defer.

contracts/validator-manager/PoSValidatorManager.sol
cam-schultz: Wow we're very inconsistent with natspec on implementation functions. I'm going to defer this to a generalized cleanup ticket.

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: We should be able to hit `!_isPoSValidator()` cases using the initial validator set

iansuvak: Yeah to hit this case you should be able to do:
1. setup pos manager with 1 initial validator
2. add a new pos validator
3. remove the initial validator

cam-schultz: Thanks all. I've added a unit test that removes an initial validator to hit this case.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -165,13 +174,31 @@ abstract contract PoSValidatorManager is
...
+    }
+
+    function forceInitializeEndValidation(
+        bytes32 validationID,
+        bool includeUptimeProof,
+        uint32 messageIndex
+    ) external {
+        _initializeEndPoSValidation(validationID, includeUptimeProof, messageIndex);
+    }
+
```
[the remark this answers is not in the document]
cam-schultz: Added a comment here.

michaelkaplan13 (approved):

tests/utils/validator_manager.go
```diff
@@ -929,14 +929,34 @@ func InitializeEndNativeValidation(
...
+		0,
+	)
+	Expect(err).Should(BeNil())
+	return WaitForTransactionSuccess(ctx, subnet, tx.Hash())
+}
+
 func InitializeEndERC20Validation(
 	ctx context.Context,
-	senderKey *ecdsa.PrivateKey,
+	sendingKey *ecdsa.PrivateKey,
```
iansuvak, line 953: Can we change these back to `senderKey` for the standardization. This change as made in https://github.com/ava-labs/teleporter/pull/576

PR #563, opened 2024-09-19
Staking rename

iansuvak:
## Why this should be merged
Previous PR didn't automatically update base and I accidentally merged into staking-rewards instead of staking-contract. 

## How is this documented


geoff-vball (approved):
👍

michaelkaplan13 (approved):

PR #565, opened 2024-09-19
Small fixes pre-audit

geoff-vball:
## Why this should be merged
Fixes a bug where the map for `nodeID` -> `ValidationID` was only populated once a validation period was active, and deleted when the validation period was `Completed` or `Invalidated`. 
Someone could have registered two validation periods with the same NodeID, if one came back as active first, and then the second came back as `invalid`, this would delete the `nodeID` -> `ValidationID` entry in the map. 
I chose to add the NodeID to the map when the validator is registered, that way any NodeID can only ever be registered once at a time.

Separates some `requires` cases, and eliminates some checks that are not possible to hit.

Renames `PoSValidatorRequirements` to `PoSValidatorInfo` since `owner` is not a requirement. We can also put the cached uptime into this struct in a future PR.

For delegations, adds checks that the unpacked validation ID matches the delegator's validation ID.

Fixes the case where a Validator exits while a Delegator is `PendingAdded`. The Delegator's start time would be greater than it's end time, causing an underflow in the reward calculation, causing a revert in the completion step.

Adds a new external method `endDelegationCompletedValidator` to account for the case where a validator has fully exited and the Delegator will not be able to get a `SubnetValidatorWeightUpdateMessage` signed by the p-chain because the validator's state will have been removed.

## How is this documented


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -520,6 +543,18 @@ abstract contract PoSValidatorManager is
...
+    }
+
+    function _completeEndDelegation(bytes32 delegationID) internal {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        Delegator memory delegator = $._delegatorStakes[delegationID];
+        bytes32 validationID = delegator.validationID;
+
+        // Once this function completes, the delegation is completed so we can clear it from state now.
+        delete $._delegatorStakes[delegationID];
```
cam-schultz, line 556: Can we move this to the end of the function, just ahead of event emission? IMO it would improve readability by making it clearer that the above `Delegator memory` variable is unaffected by the delete.

geoff-vball: This is to prevent possible reentrancy. We have reentrancy guards, but I think this is good practice to delete the balance/state before calling out to an external function.

cam-schultz: Makes sense, thanks

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -388,10 +422,41 @@ abstract contract PoSValidatorManager is
...
+        bytes32 validationID = delegator.validationID;
+        Validator memory validator = getValidator(validationID);
+
+        // Ensure the validation is completed
+        require(
+            validator.status == ValidatorStatus.Completed,
+            "PoSValidatorManager: validation not completed"
+        );
+
+        require(
```
michaelkaplan13, line 438: In the case where the validator is ended after the call to `initializeAddDelegator` but before `completeAddDelegator`, is the expected flow for the delegator user to still call `completeAddDelegator`?

That may not be possible if the validator has ended, since the P-Chain won't be willing to sign the `SubnetValidatorWeightUpdated` messages any longer. We may want to allow the delegator status to be `PendingAdded` here also to support this case.

contracts/validator-manager/interfaces/IPoSValidatorManager.sol
```diff
@@ -131,6 +129,14 @@ interface IPoSValidatorManager is IValidatorManager {
      */
     function completeDelegatorRegistration(uint32 messageIndex, bytes32 delegationID) external;
 
+    /**
+     * @notice Removes a delegator from a completed validation period. The delegator can be in either the active or pending removed
+     * state. No uptime proof is required in this case, because it will have been provided by the validator upon their exit.
+     * Note that this function can only be called by any address to clean up the delegation.
```
michaelkaplan13, line 135: "can only be called by any address" seems a bit contradictory 😅 

Think there is arguments to be made that either it should only be able to be called by the owner of the delegation, or that it should be callable by anyone, but lets clarify either way.

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: Yeah, they could exit through `completeDelegatorRegistration` without attaching a warp message, but also letting them exit through this function is a little more clear.

michaelkaplan13: Ah, @cam-schultz pointed out that in `completeDelegatorRegistration`, the warp message isn't used at all in the case that the validator has already been completed.

michaelkaplan13 (approved):

iansuvak (approved):

PR #572, opened 2024-09-24
delegation readme + state diagram

cam-schultz:
## Why this should be merged
Adds high level documentation describing delegation, as well as a state transition diagram.

## How is this documented


geoff-vball (approved):
Some of the function calls change in https://github.com/ava-labs/teleporter/pull/573 if we think those changes are worthwhile. We should remember to update this afterwards.

iansuvak (approved):
🔥

PR #573, opened 2024-09-24
Delegator refactor

geoff-vball:
## Why this should be merged
Fixes state transitions to comply with https://github.com/ava-labs/teleporter/pull/572

## How this works
Removes `endDelegationCompletedValidator` in favour of using the original function calls. If the validator is in the `Completed` state, then any call to the function appropriate for the state the delegator is in will move the delegator all the way to the completed state.

## How this was tested
TODO

## How is this documented
https://github.com/ava-labs/teleporter/pull/572


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -357,112 +357,62 @@ abstract contract PoSValidatorManager is
...
-
-            emit DelegatorRemovalInitialized({
-                delegationID: delegationID,
-                validationID: delegator.validationID,
-                endTime: validator.endedAt
-            });
-
-            return;
-        }
+            delegatorStartTime = validator.endedAt;
```
michaelkaplan13, line 381: Could you explain this case again?

It seems incorrect to me to not require a valid message here. If a delegator starts to register, and then the validator starts to exit, we should still require proof that the delegator's message has been delivered to the P-Chain in order to mark them as `active` right?

I guess marking them as `active` doesn't have an effect either way since their end time is locked in and is no later than their start time, but it may be cleaner and more intuitive to have these two cases handled exactly the same, and then handle the possibility of `start time > end time` when removing the delegator

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -482,11 +432,15 @@ abstract contract PoSValidatorManager is
         if (delegator.owner != _msgSender()) {
             revert InvalidAddress();
         }
+        // Delegators may not initiate removal when the validator's state is pending.
+        if (validator.status == ValidatorStatus.PendingRemoved) {
```
michaelkaplan13, line 436: Could we make this a bit more defensive and check `if (validator.status != ValidatorStatus.Active)`?

michaelkaplan13: Ah nvm I see we handle the `Completed` case also. Can we move the check a single if/else if/else block though? 

ie.
```
if (status == active){
    ...
} else if (status == completed) {
    ...
} else {
   revert
}

File: contracts/validator-manager/PoSValidatorManager.sol (hunk 1 above)
Comment (geoff-vball) [reply contracts/validator-manager/PoSValidatorManager.sol:381]: Hmm, yeah I think I like the handling of `start time > end time` separately.

File: contracts/validator-manager/PoSValidatorManager.sol (hunk 2 above)
Comment (geoff-vball) [reply contracts/validator-manager/PoSValidatorManager.sol]: I refactored this a bunch in **https://github.com/ava-labs/teleporter/pull/573/commits/9dbb1faa8ebbf54013a7f92f876aa4bc3b3afa30**, let me know what you think.

File: contracts/validator-manager/PoSValidatorManager.sol [hunk 3]
```diff
@@ -73,6 +73,8 @@ abstract contract PoSValidatorManager is
 
     uint16 public constant MAXIMUM_DELEGATION_FEE_BIPS = 10000;
 
+    event GeoffEvent(uint64 val);
```
minghinmatthewlam, line 76: this can be removed

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -486,46 +426,64 @@ abstract contract PoSValidatorManager is
...
+            revert InvalidValidatorStatus();
+        }
+
+        emit GeoffEvent(delegationEndTime);
+        emit GeoffEvent(delegator.startedAt);
+
+        // Only give rewards in the case that the delegation started before the validator exited.
+        if (delegationEndTime <= delegator.startedAt) {
+            return;
         }
```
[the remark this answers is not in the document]
geoff-vball: Yeah, I was debating whether it should be in the calculator or the validator manager... I do think having the check in the calculator probably makes a little more sense.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -357,58 +357,36 @@ abstract contract PoSValidatorManager is
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
 
         Delegator memory delegator = $._delegatorStakes[delegationID];
+        bytes32 validationID = delegator.validationID;
         Validator memory validator = getValidator(delegator.validationID);
```
michaelkaplan13, line 361:
```suggestion
        bytes32 validationID = delegator.validationID;
        Validator memory validator = getValidator(validationID);
```

Would be fine with removing the local variable either way TBH

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -486,46 +424,61 @@ abstract contract PoSValidatorManager is
...
             validatorStartTime: validator.startedAt,
             stakingStartTime: delegator.startedAt,
             stakingEndTime: delegationEndTime,
-            uptimeSeconds: validatorUptimeSeconds,
+            uptimeSeconds: $._completedValidationUptimeSeconds[validationID],
             initialSupply: 0,
             endSupply: 0
         });
-
-        $._delegatorStakes[delegationID] = delegator;
```
michaelkaplan13, line 522: Are we properly saving the end time for the delegator now still? I don't see where if so

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -562,30 +515,32 @@ abstract contract PoSValidatorManager is
...
-        // complete delisting for an earlier delegation. This is necessary because the P-Chain is only willing
-        // to sign the latest weight update.
-        if (delegator.endingNonce > nonce) {
-            revert InvalidNonce();
-        }
-
         // Ensure the delegator is pending removed. Since anybody can call this function once
         // end delegation has been initialized, we need to make sure that this function is only
         // callable after that has been done.
         if (delegator.status != DelegatorStatus.PendingRemoved) {
```
michaelkaplan13, line 521: In a case where the validator ends before the delegator, is `initializeEndDelegation` intended to be the only function allowed to be called to complete the delegation? If so, can we make that clear in some of the documentation/comments?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -486,46 +424,61 @@ abstract contract PoSValidatorManager is
...
-            // Further, it is impossible to retrieve an uptime proof for an already ended validation,
-            // so there's no need to check any uptime proof provided in this function call.
-            validatorUptimeSeconds = $._completedValidationUptimeSeconds[validationID];
+            revert InvalidValidatorStatus();
+        }
+
+        emit DelegatorRemovalInitialized({delegationID: delegationID, validationID: validationID});
+    }
+
+    function _calculateDelegationReward(bytes32 delegationID) private {
```
michaelkaplan13, line 449: This could just be my personal preference, so feel free to take it or leave it: I think it'd be a bit cleaner to have the `_calculate...` methods be `pure` or `view`, and then leave it to the caller to make any resulting state updates. That way, the `delegator` struct could be passed directly to the helper to avoid the extra look up, and it is clear that caller is responsible for saving any changes.

minghinmatthewlam: would be my pref too just based on naming of func

geoff-vball: Yeah, this is a good call :)

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: We're not, it's currently unnecessary. We calculate it in the `initializeEndDelegation` step, and then immediately use it to calculate rewards.

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: We could leave an escape in `completeEndDelegation` if the validator is `Active` and the validator is `Completed`, but I fear that would make the code a little circular.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -486,46 +424,63 @@ abstract contract PoSValidatorManager is
...
         } else {
-            // If the validation period has already ended, we have saved the uptime.
-            // Further, it is impossible to retrieve an uptime proof for an already ended validation,
-            // so there's no need to check any uptime proof provided in this function call.
-            validatorUptimeSeconds = $._completedValidationUptimeSeconds[validationID];
+            revert InvalidValidatorStatus();
+        }
 
-            delegator.endingNonce = validator.messageNonce;
+        emit DelegatorRemovalInitialized({delegationID: delegationID, validationID: validationID});
```
bernard-avalabs, line 448: nit: perhaps move this to the end of the "if" block since both the "else if" and "else" blocks either return or revert and doesn't get here.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -486,46 +424,63 @@ abstract contract PoSValidatorManager is
...
+            return 0;
+        }
+
+        return $._rewardCalculator.calculateReward({
             stakeAmount: weightToValue(delegator.weight),
             validatorStartTime: validator.startedAt,
             stakingStartTime: delegator.startedAt,
             stakingEndTime: delegationEndTime,
-            uptimeSeconds: validatorUptimeSeconds,
+            uptimeSeconds: $._completedValidationUptimeSeconds[delegator.validationID],
```
bernard-avalabs, line 480: It looks like "$._completedValidationUptimeSeconds[delegator.validationID]" is only updated in initializeEndDelegation if includeUptimeProof is true. Am I correct to assume that the delegation reward should be 0 if an uptime proof was never provided for this validationID?

bernard-avalabs (commented):
Generally makes sense to me. Left a few minor comments mostly to help with my own understanding.

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: So this has changed a bit with Cam's PR [here](https://github.com/ava-labs/teleporter/pull/561). We cache the uptime proof provided, so this could have a value whether the delegator provides an uptime proof or not.

michaelkaplan13 (approved):
I didn't check the unit tests too closely, but went through each of the state transitions in https://github.com/ava-labs/teleporter/blob/validator-manager-doc-improvements/contracts/validator-manager/StateTransition.md and confirmed they match this implementation. 👍

iansuvak (approved):

PR #574, opened 2024-09-25
Unit test cleanup

geoff-vball:
## Why this should be merged
Refactors a lot of our testing code. Simplifies creating the first validator and delegator for tests that should start out with them.

## How this was tested
Existing tests

## How is this documented

### DIFF
--- a/contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
+++ b/contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
@@ -151,27 +151,15 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
     }
 
     function testInvalidInitializeEndTime() public {
-        bytes32 validationID = _setUpCompleteValidatorRegistration({
-            nodeID: DEFAULT_NODE_ID,
-            subnetID: DEFAULT_SUBNET_ID,
-            weight: DEFAULT_WEIGHT,
-            registrationExpiry: DEFAULT_EXPIRY,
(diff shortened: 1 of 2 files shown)


contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -882,104 +691,189 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
...
+            delegationID,
+            completeRegistrationTimestamp,
+            expectedValidatorWeight,
+            expectedNonce
+        );
         return delegationID;
     }
 
-    function _setUpInitializeEndDelegation(
+    function _initializeEndDelegationValidatorActiveWithChecks(
```
cam-schultz, line 738: Is there a specific use case you have in mind that would be better served by having separate `WithChecks` functions? It looks like the functions without the checks are not called directly anywhere else.

cam-schultz (approved):
LGTM. Just one question about the necessity of separating out some of the helpers.

minghinmatthewlam (approved):

iansuvak (approved):

contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
geoff-vball: If you expect the code to revert somewhere inside of `_initializeEndDelegationValidatorActive`, you can call `vm.expectRevert` before calling it. If you used `_initializeEndDelegationValidatorActiveWithChecks` in the same situation, your test would probably fail because it failed to emit an event that it was expecting, due to the early revert.

PR #578, opened 2024-09-27
revert to using abi.encodePacked for packSubnetConversionData

iansuvak:
## Why this should be merged
`abi.encodePacked` implementation is cleaner than manual byte packing and is more performant according to the experiment done in #570
## How this works
Replaces manual byte packing with `abi.encodePacked` in a for loop
## How this was tested

- Current test pass indicating that the hash didn't change on the default input.
- #570 includes more extensive testing of two implementations with variable length of `numInitialValidators`
- Unfortunately since this is happening in setUp of the tests, existing tests don't show the change in gas usage other than the test cases deploying the contracts due to smaller bytecode. 

## How is this documented


contracts/validator-manager/ValidatorMessages.sol
```diff
@@ -146,48 +146,21 @@ library ValidatorMessages {
...
-            res[i + 68] = addrBytes[i];
-        }
-
-        // Pack the initial validators length
-        uint32 ivLength = uint32(subnetConversionData.initialValidators.length);
-        for (uint256 i; i < 4; ++i) {
-            res[i + 88] = bytes1(uint8(ivLength >> (8 * (3 - i))));
-        }
-
+        // Hardcoded 20 is for length of the managerAddress
```
minghinmatthewlam, line 149: can you add a comment about why encode packed is used instead of preallocated array

cam-schultz (approved):
LGTM. Agreed with Matt's comment that we should summarize the experiement in a comment.

contracts/validator-manager/ValidatorMessages.sol
iansuvak: Hmm I can but we already use encodePacked anyway and results of my experiment show it to just be empirically true not absolutely true.

cam-schultz (approved):

contracts/validator-manager/ValidatorMessages.sol
minghinmatthewlam: yup that's fine, just a note that we at least took a look at the other approach previously, so we know when we look back

minghinmatthewlam (approved):

PR #579, opened 2024-09-27
Custom error parameters

### LINKED ISSUE #568: Parameterize Validator Manager Custom Errors
**Context and scope**
https://github.com/ava-labs/teleporter/pull/566 added custom errors to the validator manager contracts, primarily as a way to reduce the contract size. We should circle back and add parameters where appropriate to make the error messages more meaningful.

- We should evaluate the context in which each error is called, and aim to make each error case unique within a given context.
- We will need to update the unit tests to catch the parameterized errors. `forge-std` 1.9.2 does not yet include `expectPartialRevert`, which allows for name based pattern matching on custom errors, without requiring explicit parameter matches. The most robust approach would be to match parameter values anyway, though there may be some scenarios in which that's overkill.

**Open questions**
Questions that are still being discussed.

minghinmatthewlam:
## Why this should be merged
Fixes #568 

## How this works
Adds parameters to the custom errors, and breaks down generic errors to be more specific.

## How this was tested
CI

## How is this documented


contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -617,6 +657,13 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
         assertEq(v3, 1e27);
     }
 
+    function testPoSValidatorManagerStorageSlot() public view {
```
iansuvak, line 660: was the change to storage locations just to make these tests possible?

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -62,18 +62,24 @@ abstract contract ValidatorManager is Initializable, ContextUpgradeable, IValida
...
-    error NodeAlreadyRegistered();
-    error InvalidSubnetConversionID();
-    error InvalidInput();
-    error InvalidExpiry();
+    error InvalidValidatorManagerAddress(address validatorManagerAddress);
+    error InvalidWarpOriginSenderAddress(address senderAddress);
+    error InvalidValidatorManagerBlockchainID(bytes32 blockchainID);
+    error InvalidWarpSourceChainID(bytes32 sourceChainID);
+    error InvalidRegistrationExpiry(uint64 registrationExpiry);
     error InvalidInitializationStatus();
```
iansuvak, line 70: tiny nit feel free to ignore but it doesn't seem unreasonable to me to also give this a `bool` parameter. to distinguish between validator already initialized or not

iansuvak (approved):
LGTM both of my comments are non-blocking questions

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -66,22 +66,24 @@ abstract contract PoSValidatorManager is
 
     // keccak256(abi.encode(uint256(keccak256("avalanche-icm.storage.PoSValidatorManager")) - 1)) & ~bytes32(uint256(0xff));
     // TODO: Unit test for storage slot and update slot
-    bytes32 private constant _POS_VALIDATOR_MANAGER_STORAGE_LOCATION =
+    bytes32 public constant POS_VALIDATOR_MANAGER_STORAGE_LOCATION =
```
cam-schultz, line 69: Why did you change this to be public? Since we can't enforce access within the struct definition, my strong preference would be to make the storage private and expose public getters and setters as needed.

cam-schultz (commented):
What's the size difference on `ERC20TokenStakingManager` and `NativeTokenStakingManager` with this change? The more specific errors are definitely more readable, but re-using errors reduces the contract size much more than using a custom error only once.

iansuvak:
> What's the size difference on `ERC20TokenStakingManager` and `NativeTokenStakingManager` with this change? The more specific errors are definitely more readable, but re-using errors reduces the contract size much more than using a custom error only once.
It increases the sizes but seems relatively marginal for now. I would consider making them less specific if we start running up against size limits again
image

cam-schultz:
> It increases the sizes but seems relatively marginal for now. I would consider making them less specific if we start running up against size limits again

Thanks for sharing. Agreed, this is an acceptable increase for now IMO.

contracts/validator-manager/PoSValidatorManager.sol
minghinmatthewlam: made them public to be used in unit test for storage slots. These variables are constants so do not need to worry about derived contracts accidentally accessing and overriding the value.

cam-schultz: Fair point 👍

cam-schultz (approved):

contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
minghinmatthewlam: yup

contracts/validator-manager/ValidatorManager.sol
minghinmatthewlam: I debated that, but seeing as invalid is only thrown when the bool paramter is false, seemed not needed. Not opposed to it either.

PR #581, opened 2024-10-01
Merge ictt

minghinmatthewlam:
## Why this should be merged
Merges [ICTT](https://github.com/ava-labs/avalanche-interchain-token-transfer) into this repo.

## How this works
uses `git subtree` to merge ICTT into this repo while maintaining the git history of the ICTT repo.
- moves ICTT into a subdirectory under `contracts` similar to other subdirectories like `teleporter` etc.
- Removes duplicate files that are no longer needed like duplicate `scripts`, `abi-bindings`, etc.
- Combines the ICTT `utils` folder into the existing `utilities` folder
  - `SafeERC20TransferFrom` had different implementations between ICTT and teleporter, combined into one library.
  - Rename to `ICMInitializable`
  - SendReentrancyGuardUpgradeable to show that library is compatible with upgradeable contracts
- Combine ICTT Readme's, and removes not needed setup repo instructions
- Move ICTT e2e tests to under `root/tests/flows`.
- Unit tests structure is similar to rest of `teleporter` repo
- Update e2e test setup to be compatible with ICTT e2e tests

## How this was tested
CI, all unit and e2e tests passing

## How is this documented
New combined readmes

TODO: update audit readme based on new decided structure


contracts/ictt/README.md
```diff
@@ -0,0 +1,94 @@
+# Avalanche Interchain Token Transfer (ICTT)
```
iansuvak, line 1: We should also move the teleporter readme and add a top level readme explaining the overall repo structure and types of contracts it contains but we can enter a separate issue for that

iansuvak (commented):
Generally the layout looks good to me.

It's cool that you imported the full commit history with git subtree but I'm not sure whether that's cleaner or to just maintain the old ictt repo as an archive and have a single merge commit here point to it.

foundry.toml
```diff
@@ -6,7 +6,8 @@ solc_version = '0.8.25'
 evm_version = 'shanghai'
 test = 'tests'
 bytecode_hash = "none"
-auto_detect_remappings = false
+# TODO: this fails to compile when set to false
+auto_detect_remappings = true
```
geoff-vball, line 10: We should do a quick investigation on this. It might be because ICTT no longer has multiple copies of some submodules

tests/local/e2e_test.go
```diff
@@ -24,6 +24,14 @@ const (
 	upgradabilityLabel       = "upgradability"
 	utilsLabel               = "utils"
 	validatorSetSigLabel     = "ValidatorSetSig"
+
```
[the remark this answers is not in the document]
geoff-vball: We should also consider splitting up our E2E tests. The run on this PR was 11min. We can split the separate product tests into parallel E2E tests, and we could even cause tests for each product to run only when the contracts for that product have been changed.

geoff-vball (approved):

minghinmatthewlam:
there should be a decision on `audit` folder structure, right now there's a root audits folder, as well as subdirectory under ictt.

cam-schultz:
> there should be a decision on `audit` folder structure, right now there's a root audits folder, as well as subdirectory under ictt.

I agree we should move all audits to the top level

geoff-vball:
@cam-schultz I opened a ticket here to consolidate the audits somewhere when we get the Validator Manager audit back https://github.com/ava-labs/teleporter/issues/593

cam-schultz (approved):

iansuvak (approved):

tests/local/ictt/ictt_suite_test.go
```diff
@@ -0,0 +1,171 @@
...
+	if os.Getenv("RUN_E2E") == "" {
+		t.Skip("Environment variable RUN_E2E not set; skipping E2E tests")
+	}
+
+	RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "Teleporter e2e test")
+}
+
+// Define the Teleporter before and after suite functions.
+var _ = ginkgo.BeforeEach(func() {
```
cam-schultz, line 48: This was `BeforeSuite` and `AfterSuite` before. Is there a reason to set up a new network for each flow?

michaelkaplan13 (approved):

PR #585, opened 2024-10-02
Acp 77 updates

cam-schultz:
## Why this should be merged
Updates the Validator Manager contracts to match the updated ACP-77 spec: https://github.com/avalanche-foundation/ACPs/pull/149/files

Note: Until an Avalanchego commit that includes the updated message specifications and corresponding signature handlers is available, e2e tests are not expected to pass.

Review focus:
- `contracts/validator-manager`

## How this works
- Updates the message specs in `MessageSpec.md` and `ValidatorMessages.sol`
- Updates the internal structs to match the ordering of the variables in the updated spec

## How this was tested
CI

## How is this documented


contracts/validator-manager/ValidatorMessages.sol
```diff
@@ -230,65 +267,169 @@ library ValidatorMessages {
...
+        ValidationPeriod memory validation;
 
         // Unpack the codec ID
-        uint16 codecID;
-        for (uint256 i; i < 2; ++i) {
-            codecID |= uint16(uint8(input[i])) << uint16((8 * (1 - i)));
-        }
-        if (codecID != CODEC_ID) {
-            revert InvalidCodecID();
+        {
```
cam-schultz, line 274: Local variables are put into inner scopes to avoid max stack depth errors.

iansuvak: can we add a comment for this?

contracts/validator-manager/ValidatorMessages.sol
```diff
@@ -162,11 +165,15 @@ library ValidatorMessages {
         // was tested against pre-allocating the array and doing manual byte by byte packing and
         // it was found to be more gas efficient.
         for (uint256 i = 0; i < subnetConversionData.initialValidators.length; i++) {
+            if (subnetConversionData.initialValidators[i].blsPublicKey.length != 48) {
```
iansuvak, line 168: I know that we are doing this in other packing methods as well but I'm not convinced that we want to check this here.

With current usage this is not going to catch any user errors since the hash that we will verify against the hash of packed bytes is signed by the p-chain. 

I doubt that this would change the efficiency of the compiled bytecode but I would prefer to be careful with keeping the logic here as minimal as possible to make sure that the gas efficiency comment above this stays true. If we do want to check it still I would slightly prefer to do it in the `ValidatorManager.sol` where we loop through `validators` already.

Feel free to drop if I'm being paranoid 😅

contracts/validator-manager/ValidatorMessages.sol
```diff
@@ -202,19 +224,34 @@ library ValidatorMessages {
...
-            revert InvalidMessageLength();
+            revert InvalidBLSPublicKey();
         }
 
         // solhint-disable-next-line func-named-parameters
         bytes memory res = abi.encodePacked(
             CODEC_ID,
             REGISTER_SUBNET_VALIDATOR_MESSAGE_TYPE_ID,
             validationPeriod.subnetID,
+            uint32(validationPeriod.nodeID.length),
```
iansuvak, line 235: Nothing we can do here but the fact that now we have messages with multiple dynamic fields means we can't use the current version of generator in #482 here any longer 😢

geoff-vball (commented):
Overall LGTM :)

contracts/validator-manager/ValidatorMessages.sol
cam-schultz: Agreed. We removed input validation on the `SubnetConversionData` parameter to `initializeValidatorSet` for the same reason. I've made this change here.

contracts/validator-manager/ValidatorMessages.sol
iansuvak: This is actually false, we can still do it but it's more involved

contracts/validator-manager/MessageSpec.md
```diff
@@ -1,45 +1,115 @@
...
+- Description: Confirms conversion to a Permissionless Subnet on the P-Chain
+
+- Signed by: P-Chain
+
+- Consumed by: Validator Manager contract
+
+Specification:
+
+```
+SubnetConversionMessage
```
michaelkaplan13, line 14: Doesn't matter at all for this PR, but worth noting that markdown supports tables that provide auto-formatting in the future. It makes it easy to modify in the future without having to manually correct all of the spacing.

contracts/validator-manager/MessageSpec.md
```diff
@@ -101,27 +159,27 @@ Specification:
                           +----------+
 ```
 
-### `SubnetValidatorWeightUpdateMessage`
+### `ValidationUptimeMessage`
+
+- Description: Provides a Validator's uptime for calculating staking rewards
 
-Description: Acknowledges a Validator weight update
+- Signed by: Subnet
```
michaelkaplan13, line 166: Not for this PR, but just a heads up that we'll be updating the terminology from subnet -> L1 soon, including for these message names and transaction types.

Subnets will refer to pre-Etna Subnets, where their validator set must also validate the primary network, making it a subset of the primary network validators.

L1s will refer to post-Etna sovereign networks, whose validators sets can be completely independent of the primary network.

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -236,7 +233,7 @@ abstract contract ValidatorManager is Initializable, ContextUpgradeable, IValida
         if (input.blsPublicKey.length != BLS_PUBLIC_KEY_LENGTH) {
             revert InvalidBLSKeyLength(input.blsPublicKey.length);
         }
-        if (input.nodeID == bytes32(0)) {
+        if (input.nodeID.length == 0) {
```
michaelkaplan13, line 236: As of Etna, the P-Chain will only support 20 byte node IDs, but it's planned to add support for other node IDs in the future (particularly 32-byte Ed25519 node IDs).

I've gone back and forth on if we should require 20 byte node IDs here and track them as `bytes20`, or if we should leave it more open to support those future upgrades by default.

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -251,9 +248,11 @@ abstract contract ValidatorManager is Initializable, ContextUpgradeable, IValida
             ValidatorMessages.ValidationPeriod({
                 subnetID: $._subnetID,
                 nodeID: input.nodeID,
-                weight: weight,
                 blsPublicKey: input.blsPublicKey,
-                registrationExpiry: input.registrationExpiry
+                remainingBalanceOwner: input.remainingBalanceOwner,
```
michaelkaplan13, line 252: Should we validate the `PChainOwners` here? The requirements are:
- If threshold is 0, addresses must be empty.
- threshold <= len(addresses)
- addresses must be sorted in ascending order

bernard-avalabs (approved):
LGTM. Just one suggestion about unpacking/packing for a future PR.

PR #591, opened 2024-10-09
Post-Etna ERC20 validator e2e

cam-schultz:
## Why this should be merged
Implements the full validator registration flow post Etna for ERC20 PoSValidatorManagers

Review focus:
- `tests/flows/validator-manager/erc20_token_staking.go`

## How this works
Adds support for issuing P-Chain tx's. Converts the subnet to permissionless as part of the test flow.
This PR is NOT expected to pass CI. The main objective of this PR is to add the functionality needed to convert all tests to post-Etna validator management via the ValidatorManager contract. A following PR will implement that refactor.

## How is this documented

PR #592, opened 2024-10-10
E2e refactor separate suites

cam-schultz:
## Why this should be merged
Following https://github.com/ava-labs/teleporter/pull/591, updates the remaining validator manager e2e tests to use the full validator registration flow with the P-Chain.

## How this works
- Modifies native token staking, erc20 token staking, and PoA -> PoS migration e2e tests to send P-Chain txs and consume Warp messages signed by the P-Chain
- Separates the top-level e2e suite into separate suites for `teleporter`, `governance`, and `validator-manager`. Each of these suites run in their own CI jobs.
- Tears down the tmpnet network in between each `validator-manager` suite test since they modify the underlying network topology
- Consolidates the `validator-manager` tests to reduce the overall e2e running time
  - `erc20_delegation.go` removed; delegation logic moved to `erc20_token_staking.go`
  - `native_delegation.go` removed; delegation logic moved to `native_token_staking.go`
  - `poa_validator_manager.go` removed; PoA owner checks moved to `poa_to_pos.go`

## How this was tested
CI

## How is this documented
N/A


scripts/local/e2e_test.sh
```diff
@@ -9,6 +9,53 @@ TELEPORTER_PATH=$(
   cd ../.. && pwd
 )
 
+function printHelp() {
+    echo "Usage: ./scripts/local/e2e_test.sh [--component component]"
+    echo ""
+    printUsage
+}
+
+valid_components="teleporter governance validator-manager"
```
geoff-vball, line 18: If we just test for the existence of `./tests/local//$component/$component.test`, then we don't have to maintain this list.

tests/local/teleporter/teleporter_suite_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if os.Getenv("RUN_E2E") == "" {
+		t.Skip("Environment variable RUN_E2E not set; skipping E2E tests")
+	}
+
+	RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "Teleporter e2e test")
+}
+
+// Define the Teleporter before and after suite functions.
+var _ = ginkgo.BeforeSuite(func() {
```
geoff-vball, line 43: Is this setup the same for all the suites? Can we extract this function out somewhere common?

cam-schultz: The only difference is that `governance` and `validator-manager` suites omit setting up Teleporter. I'm open to suggestions, but IMO the set of network setup functions are sufficiently modular to cover our test cases.

cam-schultz: Sorry - got my wires crossed. This [WIP branch](https://github.com/ava-labs/teleporter/compare/e2e-refactor-separate-suites...e2e-refactor-cleanup?expand=1) makes the change I described. With it, the `governance` and `validator-manager` suites have no dependency on the Teleporter contracts

scripts/local/e2e_test.sh
```diff
@@ -15,8 +15,6 @@ function printHelp() {
     printUsage
 }
 
-valid_components="teleporter governance validator-manager"
-
```
geoff-vball: We still reference `valid_components` on line 21. We should probably just `ls ./tests/local/` to give a list of valid components.

cam-schultz: I just removed the variable containing valid components altogether to simplify things.

geoff-vball (approved):

scripts/local/e2e_test.sh
```diff
@@ -9,6 +9,49 @@ TELEPORTER_PATH=$(
...
+function printHelp() {
+    echo "Usage: ./scripts/local/e2e_test.sh [--component component]"
+    echo ""
+    printUsage
+}
+
+function printUsage() {
+    cat << EOF
+Arguments:
+    --component <component>          Component test suite to run. Must be a valid directory in tests/local/
```
iansuvak, line 21: Love the change to be able to only test the component that changed but should we still have the ability to run all of them easily?

I'm thinking it might be worthwhile locally and maybe on merge to master which would be a good sanity check and be rarer than each PR commit push

cam-schultz: For CI, we should run the suites in parallel on separate runners in all scenarios for expediancy and isolation. The local use case makes sense though. I've modified the script to run all components by default if no argument is provided.

PR #602, opened 2024-10-15
Validators can remove Delegators

### LINKED ISSUE #569: Allow Validators to remove Delegators
**Context and scope**
With the current approach to churn tracking, it's possible for a validation to not be able to exit if it's total weight (initial stake + delegations) exceeds the limit for one-shot changes. Assuming that the churn limits are symmetric for additions and removals, a validator can register with no problem, but not be able to exit if it has delegations.

To work around this, validators should have the ability to end delegations, in addition to the delegation owner. Validators already have the ability to end delegations by ending their validation. Further, validators are incentivized to keep delegators around in order to collect their delegation fees, so validators shouldn't be booting delegators without cause.

Validators should not be able to remove delegators until `minStakeTime` has passed.

**Open questions**
One restriction we could impose would be to have the validator cease earning rewards as soon as it removes a delegator. This would more directly support the use case of exiting a validation (with delegators) without violating churn limits. This may not be worth the complexity, however, since validators are incentivized to keep delegators around for as long as possible to maximize their delegation fee.

cam-schultz:
## Why this should be merged
Fixes #569 

## How this works
Allows the validation owner to call `initializeEndDelegation` as long as the minimum stake duration has elapsed.

## How this was tested
CI, added unit tests

## How is this documented
Updates the README with Validator and Delegator owner role descriptions.


iansuvak (approved):
LGTM we might want to add another utility function that's callable by validation owner only to end all valid delegations but that's out of scope for this certainly and might be too much of a complication for the minor UX improvement

geoff-vball (approved):
👍

geoff-vball:
@iansuvak That would also require keeping a list... Which we might want to do, but adds a lot of complication and overhead.

PR #604, opened 2024-10-17
External packing lib

cam-schultz:
## Why this should be merged
Marks the methods in `ValidatorMessages.sol` as external to save on contract size for the validator managers. See failing test example: https://github.com/ava-labs/teleporter/actions/runs/11375625546/job/31646613603?pr=603

There were a handful of test framework changes that needed to be made to properly integrate external libraries.

## How this works
- ~Changing the packing functions to external breaks the unit tests since the `vm` calls are directed to `ValidatorMessages` instead of the validator manager under test. As a (hacky) workaround, I added an internal version of `ValidatorMessages` to `contracts/validator-messages/tests/utils`.~ Unit tests refactored to use the external library.

- Integrating an external library also broke our foundry + abigen toolchain. `forge build` outputs artifacts that differ from the `solc --combined-json` output that abigen expects. The workaround here was to compile each bound contract directly using `solc`, and pass the output combined JSON directly to abigen. The different toolchain resulted in significant changes to the generated bindings.

- To make the foundry project remappings compatible with `solc`, I had to remove the `openzeppelin-contracts` submodule so that the nested remapping in `openzeppelin-contracts-upgradeable` to _its_ `openzeppelin-contracts` submodule works properly (otherwise we get multiple definition errors)

- There seems to be an issue fully cleaning up tmpnet instances within a program run. This causes subsequent validator manager tests to fail, even though we're tearing down the network in between runs. Reassigning the `LocalNetworkInstance` variable does not work, nor does scoping the `network` instance to each individual flow. The workaround I've identified is to run each validator-manager flow serially in their own processes.

## How this was tested
CI. Updated ABI bindings checker job.

## How is this documented
N/A


geoff-vball (approved):

scripts/local/e2e_test.sh
```diff
@@ -92,6 +92,20 @@ for component in $(echo $components | tr ',' ' '); do
...
+        for flow in "ERC20" "Native" "PoA"; do
+            echo "Running $flow flow"
+            RUN_E2E=true ./tests/local/$component/$component.test \
+            --ginkgo.vv \
+            --ginkgo.label-filter=${GINKGO_LABEL_FILTER:-""} \
+            --ginkgo.trace \
+            --ginkgo.focus=$flow
+        done
+        continue
+    fi
```
geoff-vball, line 107: Would it make sense to split these into 3 separate jobs for CI?

cam-schultz: Let's keep it as one job to keep this workaround encapsulated in this script.

scripts/abi_bindings.sh
```diff
@@ -64,26 +64,66 @@ function convertToLower() {
     fi
 }
 
+remove_matching_string() {
```
michaelkaplan13, line 67: Could you add a small doc comment for this bash function here?

scripts/local/e2e_test.sh
michaelkaplan13: Were we able to pin down what the issue was with running multiple flows in the same process here?

.github/workflows/abi_bindings_checker.yml
```diff
@@ -27,6 +27,12 @@ jobs:
       - name: Install Foundry
         run: ./scripts/install_foundry.sh
 
+      - name: Install solc
+        run: |
+          wget https://github.com/ethereum/solidity/releases/download/v0.8.25/solc-static-linux
```
michaelkaplan13, line 32: Could we move the solidity version used to a more visible place? Or ideally keep it defined in a single place between both `solc` and `forge`

geoff-vball (commented):
LGTM once Michael's comments are addressed.

.github/workflows/abi_bindings_checker.yml
cam-schultz: Addressed in https://github.com/ava-labs/teleporter/pull/626

scripts/abi_bindings.sh
cam-schultz: Addressed in https://github.com/ava-labs/teleporter/pull/626

cam-schultz:
Closing in favor of https://github.com/ava-labs/teleporter/pull/626

Issue #598, opened 2024-10-14
Verify initial validator weights are removable


cam-schultz:
**Context and scope**
As part of permissionless L1 initialization, an initial validator set is provided in the call to `[initializeValidatorSet](https://github.com/ava-labs/teleporter/blob/validator-manager/contracts/validator-manager/ValidatorManager.sol#L143)`. The weights of these validators are checked against the signed Warp message from the P-Chain attesting that the subnet was converted.

Not all initial validator weights allow the validator set to be changed without violated the churn tracking limits. The maximum one-shot churn percentage is [capped](https://github.com/ava-labs/teleporter/blob/validator-manager/contracts/validator-manager/ValidatorManager.sol#L62) at 20% of the L1's current weight. For example, if the initial validator set consisted of a single validator with weight 3, then it would be impossible to add or remove a validator, since a weight change of 1 (the minimum amount) would be a 33% change.

There's nothing we can do on the contract side to ensure the `ConvertSubnetTx` specifies valid weights, but validation on the contract level would at least cause an early failure, rather than when trying to change the validator set, at which point it may no longer be feasible to tear down the L1 and start again.

**Discussion and alternatives**
We should at the very least validate that the L1's total initial weight is at least 5, so that validators can be added after the fact. We should also disallow removing validators if it would cause the total weight to dip below 5.

**Open questions**
Should we implement this validation check in `ConvertSubnetTx` on the P-Chain side as well?

PR #606, opened 2024-10-18
Check weights

geoff-vball:
## Why this should be merged
Closes https://github.com/ava-labs/teleporter/issues/598

## How this works
Checks that the total weight of all validators cannot go below the point where a validator cannot be added with weight at least 1.

## How this was tested
Unit tests.

## How is this documented
N/A


cam-schultz (approved):
LGTM. Double checked the inequalities.

PR #617, opened 2024-10-18
M-02 Churn Tracker Can Be Abused to Stall Validator Manager

geoff-vball:
## Why this should be merged
Prevents churn tracker abuse.

## How this works
The global minimum stake duration for the PoS Validator Manager cannot be less than `_churnPeriodSeconds`.

Delegators must now stake for at least `_minimumStakeDuration`

## How this was tested
Unit tests

## How is this documented


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -491,22 +495,28 @@ abstract contract PoSValidatorManager is
...
-            } else {
+            if ($._posValidatorInfo[validationID].owner != _msgSender()) {
                 revert UnauthorizedOwner(_msgSender());
             }
+
+            if (
+                block.timestamp
+                    < validator.startedAt + $._posValidatorInfo[validationID].minStakeDuration
+            ) {
+                revert MinStakeDurationNotPassed(uint64(block.timestamp));
```
cam-schultz, line 510: I think this check would make more sense in `initializeDelegatorRegistration`, that way the validator's weight is guaranteed not to change until the churn period has elapsed.

geoff-vball: This check is for Validator's that want to end a delegation attached to them. They can only end the delegation after their `minStakeDuration` has elapsed. I only refactored the nesting for these checks, I didn't change any functionality here.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -141,6 +141,10 @@ abstract contract PoSValidatorManager is
         {
             revert InvalidStakeMultiplier(maximumStakeMultiplier);
         }
+        // Minimum stake duration should be at least one churn period in order to prevent churn tracker abuse.
+        if (minimumStakeDuration < _getChurnPeriodSeconds()) {
```
bernard-avalabs, line 145: Does our expected churn period for most deployments align well with our expected minimum stake duration? For example, if minimunStakeDuration is set to the current minimum (2 weeks), is a churn period of 2 weeks or less a reasonable setting?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -491,22 +495,28 @@ abstract contract PoSValidatorManager is
...
+                block.timestamp
+                    < validator.startedAt + $._posValidatorInfo[validationID].minStakeDuration
+            ) {
+                revert MinStakeDurationNotPassed(uint64(block.timestamp));
+            }
         }
 
         if (validator.status == ValidatorStatus.Active) {
+            // Check that minimum stake duration has passed.
+            if (block.timestamp < delegator.startedAt + $._minimumStakeDuration) {
```
bernard-avalabs, line 516: For my own understanding, we only check that the minimuStakeDuration has passed when ending a delegation if the validator is still active. We don't perform this check if it is not since we still want the delegator to receive their reward in that case. I agree that this is the right semantic, but couldn't an attacker still stall the validator manager if they only delegate (potentially a large amount) to validators with staking periods that are about to end?

bernard-avalabs (commented):
Generally makes sense to me. Left a few clarification questions.

contracts/validator-manager/PoSValidatorManager.sol
cam-schultz: Whoops, thanks for clarifying.

cam-schultz (approved):

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: I think a reasonable setting for the churn period is probably on the order of 1-10 hours. Some users might opt for a very low minimum stake duration, so this is to sanity check against that.

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: Staking periods won't have defined ends, but they could end any time after their minimum stake duration has elapsed. A delegator could potentially collude with a validator to get them to end immediately, but for any reasonable `_minimumStakeDuration`,  this won't be repeatable. Especially since delegators are capped at 4-5x the validator's stake.

PR #621, opened 2024-10-21
M-01 token conversion

cam-schultz:
## Why this should be merged
Parameterizes weight to value conversion factor.

## How this works
- Adds an initialization parameter to PoSValidatorManager to specify the weight to value conversion factor. This was previously hardcoded to 1e12.
- Disallows values that round down to 0 weight.

## How this was tested
CI, new unit tests

## How is this documented
Documents gotchas when selecting the conversion factor


contracts/validator-manager/README.md
```diff
@@ -58,6 +58,11 @@ Proof-of-Authority Validator management is provided via `PoAValidatorManager`, w
 
 Proof-of-Stake Validator management is provided by the abstract contract `PoSValidatorManager`, which has two concrete implementations: `NativeTokenStakingManager` and `ERC20TokenStakingManager`. In addition to basic Validator management provided in `ValidatorManager`, `PoSValidatorManager` supports uptime-based Validation rewards, as well as Delegation to a Validator. This [state transition diagram](./StateTransition.md) illustrates the relationship between Validators and Delegators.
 
+> [!NOTE]
+> The `weightToValueFactor` fields of the `PoSValidatorManagerSettings` passed to `PoSValidatorManager`'s `initialize` function sets the factor used to convert between the value transferred to the contract as stake, and the weight that the Validator is registered with on the P-Chain. This involves integer division, which may result in loss of precision. When selecting `weightToValueFactor`, it's important to make the following considerations:
```
michaelkaplan13, line 62:
```suggestion
> The `weightToValueFactor` fields of the `PoSValidatorManagerSettings` passed to `PoSValidatorManager`'s `initialize` function sets the factor used to convert between the weight that the Validator is registered with on the P-Chain, and the value transferred to the contract as stake. This involves integer division, which may result in loss of precision. When selecting `weightToValueFactor`, it's important to make the following considerations:
```

We may want to note here that `weight` on the P-Chain is represented as a `uint64`, so the `weightToValueFactor` should be set with that in mind to ensure that the weight value will not overflow. We may even want to add a `require` statement when registering a validator that the weight is `<= UINT64_MAX`.

michaelkaplan13 (commented):
Makes sense to me, just one suggestion

contracts/validator-manager/README.md
cam-schultz: Good call. Added documentation, a check against `UINT64_MAX`, and a test.

contracts/validator-manager/README.md
```diff
@@ -58,6 +58,12 @@ Proof-of-Authority Validator management is provided via `PoAValidatorManager`, w
 
 Proof-of-Stake Validator management is provided by the abstract contract `PoSValidatorManager`, which has two concrete implementations: `NativeTokenStakingManager` and `ERC20TokenStakingManager`. In addition to basic Validator management provided in `ValidatorManager`, `PoSValidatorManager` supports uptime-based Validation rewards, as well as Delegation to a Validator. This [state transition diagram](./StateTransition.md) illustrates the relationship between Validators and Delegators.
 
+> [!NOTE]
```
bernard-avalabs, line 61: Suggestion: It may be helpful to include an example that illustrates the first 2 considerations.

bernard-avalabs (approved):
LGTM. Adding an example in the README about possible issues with conversion might help clarify any possible misunderstandings.

michaelkaplan13 (approved):

contracts/validator-manager/README.md
cam-schultz: Added examples

PR #624, opened 2024-10-24
Unit test coverage

geoff-vball:
## Why this should be merged
Provides full coverage for all Validator Manager contracts except `ValidatorManager.sol` (WIP).

There are a couple checks that are currently unreachable, but I think are still good to have for when code changes.

Covecov doesn't seem to be able to handle one of the `if, elif, else` statements.

## How this was tested
Screenshot 2024-10-24 at 10 14 31 AM

## How is this documented
N/A


richardpringle (commented):
Not requesting changes (necessarily), so leaving as a "comment" review. 

If you don't want to make change, you can just re-request a review and I'll approve, but please respond to that one question first! Just making sure you wanted to do that.

contracts/validator-manager/ValidatorMessages.sol
```diff
@@ -152,7 +152,7 @@ library ValidatorMessages {
      * @param conversionData The struct representing data to pack into the message.
      * @return The packed message.
      */
-    function packConversionData(ConversionData calldata conversionData)
+    function packConversionData(ConversionData memory conversionData)
```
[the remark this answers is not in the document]
geoff-vball: So, another really frustrating thing about solidity... `calldata` can only originate from an external function. You can pass it down as far as you want as `calldata`, but if you're already in a function, you can't generate new `calldata`. So in order to test this in `testPackL1ConversionData()`, we can only create a `ConversionData memory`, which we can't pass to the function call if the signature requires calldata.

There might be a workaround to this? If there's a way to call an external function with arbitrary params? I can't seem to figure it out though 😅

contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -40,6 +47,30 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
     PoSValidatorManager public posValidatorManager;
     IRewardCalculator public rewardCalculator;
 
+    PoSValidatorManagerSettings public defaultPoSSettings = PoSValidatorManagerSettings({
```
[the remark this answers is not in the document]
geoff-vball: Changed :)

richardpringle (approved):

contracts/validator-manager/tests/ERC20TokenStakingManagerTests.t.sol
```diff
@@ -38,29 +33,31 @@ contract ERC20TokenStakingManagerTest is PoSValidatorManagerTest {
...
-                minimumDelegationFeeBips: 0,
-                maximumStakeMultiplier: DEFAULT_MAXIMUM_STAKE_MULTIPLIER,
-                weightToValueFactor: DEFAULT_WEIGHT_TO_VALUE_FACTOR,
-                rewardCalculator: IRewardCalculator(address(0)),
-                uptimeBlockchainID: DEFAULT_SOURCE_BLOCKCHAIN_ID
-            }),
-            token
-        );
+
+        PoSValidatorManagerSettings memory defaultPoSSettings = _defaultPoSSettings();
```
iansuvak, line 58: I like this minimization of repetition and it's also way easier to see what the value/case being tested is. 👍

contracts/validator-manager/tests/ValidatorMessagesTests.t.sol
```diff
@@ -23,13 +25,261 @@ contract ValidatorMessagesTest is Test {
...
+                registrationExpiry: DEFAULT_EXPIRY,
+                blsPublicKey: invalidBLSKey,
+                remainingBalanceOwner: DEFAULT_P_CHAIN_OWNER,
+                disableOwner: DEFAULT_P_CHAIN_OWNER,
+                weight: DEFAULT_WEIGHT
+            })
+        );
+    }
+
+    function testRegisterSubnetValidatorMessageInvalidInputLength() public {
```
iansuvak, line 92: Should we change the name here? It looks like this test checks 3 different invalid cases and not just `InvalidInputLength`

contracts/validator-manager/tests/ValidatorMessagesTests.t.sol
```diff
@@ -23,13 +25,261 @@ contract ValidatorMessagesTest is Test {
...
+
+        // Invalid message type
+        bytes memory invalidPacked3 = _getPackedRegisterL1ValidatorMessage();
+        invalidPacked3[5] = 0x00;
+        vm.expectRevert(ValidatorMessages.InvalidMessageType.selector);
+        ValidatorMessages.unpackRegisterL1ValidatorMessage(invalidPacked3);
+    }
+
+    function testRegisterSubnetValidatorMessageInvalidCodecID() public {
+        (, bytes memory packed) = ValidatorMessages.packRegisterL1ValidatorMessage(
```
iansuvak, line 122: Do this test case and `testRegisterSubnetValidatorMessageInvalidTypeID` below  re-test functionality tested above? I agree that it would make sense to remove it from the above in that case to have a single test for single error state but we should be able to use `_getPackedRegisterL1ValidatorMessage()` as well

contracts/validator-manager/tests/ValidatorMessagesTests.t.sol
geoff-vball: Yeah I think the two tests below got copied into this test accidentally on a merge. All fixed now.

iansuvak (approved):

iansuvak (approved):

PR #625, opened 2024-10-24
Check for churn period seconds to pass when delegator ends

geoff-vball:
## Why this should be merged
Makes delegators wait at least `churnPeriodSeconds` to reclaim their stake in the case that their validator ended early. If the validator has not ended, the delegator must still wait `minStakeDuration`, where `minStakeDuration >= churnPeriodSeconds`. This prevents the delegator from potentially staking multiple times in the same churn period.

## How this was tested
Unit tests

## How is this documented
Comments


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -691,6 +697,11 @@ abstract contract PoSValidatorManager is
             revert InvalidDelegatorStatus(delegator.status);
         }
 
+        // Check that minimum stake duration has passed.
+        if (block.timestamp < delegator.startedAt + $._minimumStakeDuration) {
```
bernard-avalabs, line 701: Is this check necessary given that we know the delegator is in PendingRemoved state?

bernard-avalabs (commented):
Looks good to me. Just a quick question about the extra check in completeEndDelegation.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -599,6 +599,12 @@ abstract contract PoSValidatorManager is
             });
             return (reward > 0);
         } else if (validator.status == ValidatorStatus.Completed) {
+            // To prevent churn tracker abuse, check that one full churn period has passed,
+            // so a delegator may not stake twice in the same churn period.
+            if (block.timestamp < delegator.startedAt + _getChurnPeriodSeconds()) {
+                revert MinStakeDurationNotPassed(uint64(block.timestamp));
+            }
```
michaelkaplan13, line 606: I think this check should actually be moved into `_completeEndDelegation` itself. Currently I think there is still a corner case where `_completeEndDelegation` could be called from `completeDelegatorRegistration` prior to a churn period elapsing if the validator completed prior the delegation registration being completed.

michaelkaplan13: Otherwise, we could also just add the similar check in that case, but feels more robust and less error prone to include it in a single place within `_completeEndDelegation` (possibly with the downside of doing a redundant check in certain cases).

geoff-vball: Ahh yep, you're right

cam-schultz (approved):

michaelkaplan13 (approved):

PR #631, opened 2024-10-29
Prorated rewards v2

cam-schultz:
## Why this should be merged
Alternative to https://github.com/ava-labs/teleporter/pull/603 that implements the approach [suggested](https://github.com/ava-labs/teleporter/pull/603#issuecomment-2444246895) by @geoff-vball .

Rather than normalizing submitted uptimes to determine reward eligibility for a given claim period, we instead check that the submitted uptime is sufficient for the entire validation period to date. We then calculate the total reward for the validation period to date, and subtract any previously claimed rewards.

## How this works

This works because the uptime threshold calculation is linear: if period A (beginning from the start of the validation) has sufficient uptime relative to the start of the valdiation, and period B has sufficient uptime relative to the start of the validation, then period B also has sufficient uptime relative to the end of period A.

## How this was tested
CI

## How is this documented
TODO


bernard-avalabs:
Without a maximum staking period, we may run into an undesirable situation where a validator can accumulate a large amount of uptime, allowing the operator to turn off the validator for an extended period of time while still earning rewards. This is similar to allowing unused vacation days to be rolled over to the next year, which may be undesirable (to the company) if it can be done indefinitely.

Updating the effective validation start time after each reward claim such that it is at most X number of days in the past would cap the amount of saved uptime.

geoff-vball:
@bernard-avalabs I do think the problem of "accumulating too many vacation days" is a real issue, but I don't think tying any kind of penalty to reward redemption is the solution. In your example, if the validator has 100% uptime for 100 days, but doesn't claim any rewards early, they can still go offline for the full 25 days. 

I think the only solution we can currently implement at the contract level would be to set a maximum validation time. There's a fundamental issue with how validators track uptime for other validators (which may be worth the tradeoff in terms of efficiency etc) that stops us from restricting the amount of "vacation time" a validator can accrue, assuming that the rewards are based on an 80% uptime cliff.

contracts/validator-manager/interfaces/IPoSValidatorManager.sol
```diff
@@ -156,6 +159,16 @@ interface IPoSValidatorManager is IValidatorManager {
         uint32 messageIndex
     ) external;
 
+    /**
+     * @notice Withdraws pro-rated rewards for an active validation to the owner of the validator.
+     * Uses the uptime proof provided in the Warp message to determine rewards eligibility. Reverts if the uptime is not eligible for rewards.
+     * Rewards are calculated from the last time this function was called, or the beginning of the
+     * validation, whichever is later.
```
geoff-vball, line 166: I don't think this part is true for this implementation.

bernard-avalabs:
> In your example, if the validator has 100% uptime for 100 days, but doesn't claim any rewards early, they can still go offline for the full 25 days.

That's a good point. We can get around that by performing automatic redemption after a max period.  So in my previous example, if the automatic redemption period is 100 days, then in the event that the operator turns off their validator immediately after day 100 (assuming perfect uptime) and tries to redeem at day 125, then they would only receive rewards for the first 100 days since that is the first redemption period. They did not meet the uptime requirement for the second period. Alternatively, if they tried to redeem at 112.5 days, they would get the reward for the entire duration.

I understand this gets a bit complicated even to explain. But a max staking period would be not allow us to provide continuous staking to validator operators.

PR #637, opened 2024-11-01
Uptime e2e

cam-schultz:
## Why this should be merged
Integrates https://github.com/ava-labs/subnet-evm/pull/1377 to enable constructing signed uptime proofs. The uptime proofs are submitted in the call to `initializeEndValidation` and `forceInitializeEndValidation` for PoS validators in the e2e tests. Uptime proofs are not supplied for validators that were registered as PoA, since they are not eligible for rewards.

Also changes the requirement that the uptime proof be signed by the same chain that the validator manager is deployed on, to instead requiring that it be signed by an `uptimeBlockchainID` that is supplied at initialization. This chain must be validated by the L1 managed by the validator manager. There is no requirement that the validator manager be deployed on the chain whose validator set it is managing, so the existing check is incorrect.

## How this works
The helpers that call `initializeEndValidation` first construct an uptime proof using the helper `ConstructUptimeProofMessage`, signed by the L1.

## How this was tested
CI

## How is this documented
N/A


geoff-vball (dismissed):

geoff-vball (approved):

bernard-avalabs (approved):
LGTM

PR #639, opened 2024-11-07
Simplify local network struct

cam-schultz:
## Why this should be merged
Cleans up the `LocalNetwork` struct to act as an extension to `tmpnet`.

## How this works
- Directly composes `LocalNetwork` with `tmpnet.Network`, rather than owning an instance. This automatically exposes all `tmpnet.Network` methods, allowing for greater flexibility.
- Moves Teleporter-specific methods from `LocalNetwork` to `TeleporterTestInfo`
- Removes a handful of redundant Teleporter utils methods, instead making them `TeleporterTestInfo` methods.
- Construct the primary network and subnet `SubnetTestInfo` on demand, removing the need to consider stale state when modifying the network.

## How this was tested
CI

## How is this documented
N/A


tests/network/network.go
```diff
@@ -41,16 +33,12 @@ import (
 
 // Implements Network, pointing to the network setup in local_network_setup.go
 type LocalNetwork struct {
-	primaryNetworkInfo *interfaces.SubnetTestInfo
-	subnetsInfo        map[ids.ID]*interfaces.SubnetTestInfo
+	tmpnet.Network
```
iansuvak, line 36: Thanks for making this change! Love it.

iansuvak (approved):
Primarily went through the changes in the `network.go` and they LGTM.

geoff-vball (approved):

PR #640, opened 2024-11-07
Add Validator manager audit

geoff-vball:
## Why this should be merged
Adds Validator Manager audit. 

Moves ICTT audit into root audit folder.

## How is this documented


contracts/ictt/audits/README.md
```diff
@@ -1,8 +0,0 @@
-# Audits
```
iansuvak, line 1: Do we no longer want to have a README that points to the audits (updated of course)

geoff-vball: I think it would probably be a good idea... I'm just trying to think of the best way to do it. Should we only include line items for the most recent audit for any given contract? For instance, it doesn't feel helpful to list the Least Authority `teleporter` audit from July 2023, since this is not we want anyone to deploy. But even the most recent audit for `teleporter` isn't done on the exact commit that we tagged `v1.0.0`. I'm having trouble figuring out the best way to communicate best practices here 😅

bernard-avalabs: It would be nice to keep a historical accounting of all the audits. Perhaps it can be broken down into two sections: Latest audit, and historical audits with text to make it clear that the old contracts should not be deployed.

meaghanfitzgerald: I would keep all audit reports in the repository, pointing readers to the relevant, most up-to-date audits. Advise that old audits may now be obsolete, and legacy code superseded by new versions are **not** recommended for production deployments.

iansuvak (approved):
Confirmed that rendered markdown links correctly to the audits!

richardpringle (approved):

PR #641, opened 2024-11-09
Tests use permissionless l1

cam-schultz:
## Why this should be merged
Adds support for Etna SoVs in `LocalNetwork`. Allows for full separation of primary network and SoV validator sets. Also includes many semi-related cleanup items.

## How this works
The meat of this change involves moving the subnet conversion steps previously done in the validator manager test flows to `LocalNetwork`.
- `LocalNetwork.NewNetwork` specifies only a single bootstrap node per subnet. 
- `LocalNetwork.ConvertSubnet` converts a subnet by:
  1. Deploying a `ValidatorManager` instance (the concrete instance is specfied as an argument to the call)
  2. Issuing the `ConvertSubnetTx` on the P-Chain
  3. Initializing the subnet validators with the specified weights
  4. Removing the bootstrap nodes as subnet validators

After the subnet is converted, the bootstrap nodes validate only the primary network, and the nodes provided in the subnet conversion only validate the subnet. 

`LocalNetwork` users can choose if and when to convert the subnet. For example, the `teleporter` suite does so in the initial network setup, converting all subnets to be managed by `PoAValidatorManager`. The `governance` suite uses pre-Etna subnet validators as before. The `validator-manager` leaves it to each flow to convert the subnet according to its needs.

Additionally, cleans up the following:
- Adds a getter for the `ERC20TokenStakingManager`'s staking asset
- Uses `signature-aggregator` over the Subnet EVM Warp API to aggregate signatures across all tests
  - There seemed to be some timing inconsistencies with the Warp API that were not present when fetching the signatures directly.
- Generalizes many functions in `validator_manager.go` to use `IValidatorManager` or `IPoSValidatorManager` instances to remove duplicate code.

## How this was tested
CI

## How is this documented
N/A


contracts/validator-manager/ERC20TokenStakingManager.sol
```diff
@@ -119,6 +119,13 @@ contract ERC20TokenStakingManager is
         return _initializeDelegatorRegistration(validationID, _msgSender(), delegationAmount);
     }
 
+    /**
+     * @notice Returns the ERC20 token being staked
+     */
+    function erc20() external view returns (IERC20Mintable) {
```
cam-schultz, line 125: I added this to allow the deployment utility to return an `IPoSValidatorManager` object, and recover the `ERC20` instance after the fact. It also seems reasonable that use cases will emerge that will want to access the staked asset programatically.

scripts/abi_bindings.sh
```diff
@@ -139,21 +139,30 @@ function generate_bindings() {
...
         mkdir -p $gen_path
-
-        $GOPATH/bin/abigen --pkg $(convertToLower $contract_name) \
-                        --combined-json $combined_json \
-                        --type $contract_name \
-                        --out $gen_path/$contract_name.go \
-                        --exc $filtered_contracts
+        echo "Generating Go bindings for $contract_name..."
+        
+        if [ -z "$filtered_contracts" ]; then
```
cam-schultz, line 147: This addresses an edge case in which the only contract in the combined JSON is the contract to compile itself. In that case, there are no contracts to exclude.

tests/network/network.go
```diff
@@ -202,6 +363,37 @@ func (n *LocalNetwork) GetPrimaryNetworkInfo() interfaces.SubnetTestInfo {
 	}
 }
 
+func (n *LocalNetwork) GetSubnetInfo(subnetID ids.ID) interfaces.SubnetTestInfo {
+	for _, subnet := range n.Network.Subnets {
```
cam-schultz, line 367: There are a handful of places where we do a linear search over the subnets and nodes. I didn't think it was worth migrating to maps for random access, given that we're only ever searching over a handful of subnets/nodes.

tests/utils/proxy.go
```diff
@@ -42,8 +41,5 @@ func DeployTransparentUpgradeableProxy[T any](
 	proxyAdmin, err := proxyadmin.NewProxyAdmin(proxyAdminEvent.NewAdmin, subnet.RPCClient)
 	Expect(err).Should(BeNil())
 
-	contract, err := newInstance(proxyAddress, subnet.RPCClient)
-	Expect(err).Should(BeNil())
-
-	return proxyAddress, proxyAdmin, contract
+	return proxyAddress, proxyAdmin
```
cam-schultz, line 44: The target `T` contract instance can be constructed from the `proxyAddress` by the caller.

tests/suites/teleporter/teleporter_suite_test.go
```diff
@@ -57,7 +58,7 @@ var _ = ginkgo.BeforeSuite(func() {
 	Expect(err).Should(BeNil())
 
 	// Create the local network instance
-	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 120*2*time.Second)
```
geoff-vball, line 61: Would this make more sense as `60*4*time.Second`?

tests/flows/teleporter/validator_churn.go
```diff
@@ -57,16 +59,56 @@ func ValidatorChurn(network *localnetwork.LocalNetwork, teleporter utils.Telepor
...
+		network.GetSignatureAggregator(),
+	)
 
 	//
 	// Modify the validator set on Subnet A
 	//
 
 	// Add new nodes to the validator set
-	addValidatorsCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	addValidatorsCtx, cancel := context.WithTimeout(ctx, 90*newNodeCount*time.Second)
```
bernard-avalabs, line 76: nit: test would be more robust if the timeout is based on the sleep period in the newNodeCount loop below (e.g., adding a sleep period constant, and changing 90 to (sleepPeriod + 85)).

bernard-avalabs (commented):
Generally makes sense to me.

tests/suites/teleporter/teleporter_suite_test.go
cam-schultz: 2 corresponds to the number of subnets that are being created and converted.

tests/flows/teleporter/validator_churn.go
cam-schultz: Made this change

tests/network/network.go
```diff
@@ -379,7 +527,21 @@ func (n *LocalNetwork) Dir() string {
 }
 
 func (n *LocalNetwork) GetPChainWallet() pwallet.Wallet {
-	return n.pChainWallet
+	// Create the P-Chain wallet to issue transactions
+	kc := secp256k1fx.NewKeychain(n.globalFundedKey)
+	n.GetSubnetsInfo()
```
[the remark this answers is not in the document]
cam-schultz: Removed

geoff-vball (approved):

tests/network/network.go
```diff
@@ -368,6 +473,12 @@ func (n *LocalNetwork) SetChainConfigs(chainConfigs map[string]string) {
 			log.Error("failed to write subnets", "error", err)
 		}
 	}
+
+	// Restart the network to apply the new chain configs
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(60*len(n.Network.Nodes))*time.Second)
```
iansuvak, line 478: I like moving this here instead of having it in individual tests 👍

iansuvak (approved):

iansuvak (approved):

PR #643, opened 2024-11-13
Centralize reward management in PoSValidatorManager

richardpringle:
## Why this should be merged

If any of the reward logic changes, the change happens in one place. For the delegator, the logic only existed in one place before, but this keeps the pattern consistent and improves the readability of `_completeEndDelegation` function.

## How this works

Same as before, I just moved code

## How this was tested

Existing tests

## How is this documented

inline docs on new functions


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -760,4 +744,41 @@ abstract contract PoSValidatorManager is
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
         return $._posValidatorInfo[validationID].owner != address(0);
     }
+
+    function _withdrawValidationRewards(address rewardRecipient, bytes32 validationID) internal {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
```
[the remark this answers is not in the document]
cam-schultz: We currently retrieve the contract storage in every function that needs it, mainly under the assumption that the Solidity compiler is smart enough to cache subsequent reads of the same value. However, I'm not seeing anything in the [optimizer steps](https://docs.soliditylang.org/en/latest/internals/optimizer.html) that would directly translate. We should do a gas cost analysis of this pattern versus passing the storage object around (not necessary for this PR - let's stick to the existing pattern here)

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -760,4 +744,41 @@ abstract contract PoSValidatorManager is
...
+
+    function _withdrawDelegationRewards(
+        address rewardRecipient,
+        bytes32 delegationID,
+        bytes32 validationID
+    ) internal returns (uint256, uint256) {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        uint256 delegationRewards = 0;
+        uint256 validatorFees = 0;
```
cam-schultz, line 765: OpenZeppelin called out in their recent audit that _not_ initializing variables with their default value is more gas efficient.
```suggestion
        uint256 delegationRewards;
        uint256 validatorFees;
```

cam-schultz (commented):
One small nit, otherwise LGTM

bernard-avalabs (commented):
LGTM other than the small issue that Cam brought up.

contracts/validator-manager/PoSValidatorManager.sol
richardpringle: That's insane that this doesn't compile down to the same bytecode... will fix!

cam-schultz (approved):

bernard-avalabs (approved):
LGTM. Interesting that we are now getting Slither warnings.

iansuvak (approved):

PR #646, opened 2024-11-14
Mock the reward functionality of the native-manager

richardpringle:
## Why this should be merged

We were actually triggering logic in a function named `_expect*`. It was super misleading and really hard to track down why that was happening. Instead, I've properly mocked the functionality, triggering the `vm.deal` logic when the `_reward` function is called instead of afterwards. 

## How this works

I've effectively monkey-patched the `_rewardFunction`.

## How this was tested

It is a test :wink:

## How is this docuomented

There's a code comment that was already there that I moved.

### DIFF
--- a/contracts/validator-manager/tests/NativeTokenStakingManagerTests.t.sol
+++ b/contracts/validator-manager/tests/NativeTokenStakingManagerTests.t.sol
@@ -5,6 +5,7 @@
 
 pragma solidity 0.8.25;
 
+import {Test} from "@forge-std/Test.sol";
 import {PoSValidatorManagerTest} from "./PoSValidatorManagerTests.t.sol";
 import {NativeTokenStakingManager} from "../NativeTokenStakingManager.sol";
 import {PoSValidatorManager} from "../PoSValidatorManager.sol";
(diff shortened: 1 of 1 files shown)


contracts/validator-manager/tests/NativeTokenStakingManagerTests.t.sol
```diff
@@ -249,25 +250,15 @@ contract NativeTokenStakingManagerTest is PoSValidatorManagerTest {
...
-            abi.encodeCall(INativeMinter.mintNativeCoin, (account, amount)),
-            ""
-        );
-        // empty calldata implies the receive function will be called:
-        vm.mockCall({
-            callee: account,
-            msgValue: amount,
-            data: "", // implies receive()
-            returnData: ""
-        });
```
richardpringle, line 263: This wasn't doing anything

contracts/validator-manager/tests/NativeTokenStakingManagerTests.t.sol
```diff
@@ -249,25 +250,15 @@ contract NativeTokenStakingManagerTest is PoSValidatorManagerTest {
...
+        address nativeMinter = address(app.NATIVE_MINTER());
+        bytes memory callData = abi.encodeCall(INativeMinter.mintNativeCoin, (account, amount));
+        vm.mockCall(nativeMinter, callData, "");
+        vm.expectCall(nativeMinter, callData);
     }
 
     function _setUp() internal override returns (IValidatorManager) {
         // Construct the object under test
-        app = new NativeTokenStakingManager(ICMInitializable.Allowed);
+        app = new MockNativeTokenStakingManager(ICMInitializable.Allowed);
```
iansuvak, line 261: I'm not sure that `MockNativeTokenStakingManager` is the best name here since you aren't really mocking here but monkeypatching instead. How about `TestableNativeTokenStakingManager`

iansuvak (commented):
I'm not hard opposed to this and agree with your assessment that the current implementation has problems, but not sure that I love divergence (albeit tiny) from the actual contract we want to test and the contract we are actually testing. `_reward` is simple enough though that this is probably fine here.

contracts/validator-manager/tests/NativeTokenStakingManagerTests.t.sol
richardpringle: done

iansuvak (approved):
LGTM -- if others think that we don't want this divergence we could theoretically abstract it away behind an interface but I think this is fine for now and improvement over status quo.

geoff-vball (approved):
I like this solution. Thanks @richardpringle :)

bernard-avalabs (approved):

PR #647, opened 2024-11-15
staking rewards to any

### LINKED ISSUE #555: Allow stakers to specify a different rewards address
**Context and scope**
Stakers should be able to withdraw rewards to a different address than the `owner` address.

**Open questions**
Should owners also be allowed to change the owner of a validation or delegation?

richardpringle:
## Why this should be merged

Closes #555

## How this works

I use another mapping to set the reward-recipient. 

TODO: I should probably add another method that allows the validator to change the recipient before `completeEndValidation` is called.

## How this was tested

Unit tests
TODO: test delegator reward

## How is this documented

TODO


contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -193,9 +195,7 @@ abstract contract PoSValidatorManager is
             revert UnauthorizedOwner(_msgSender());
         }
 
-        uint256 rewards = $._redeemableValidatorRewards[validationID];
-        delete $._redeemableValidatorRewards[validationID];
-        _reward($._posValidatorInfo[validationID].owner, rewards);
+        _withdrawValidationRewards($._posValidatorInfo[validationID].owner, validationID);
```
richardpringle, line 198: Waiting on #643

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -206,7 +206,9 @@ abstract contract PoSValidatorManager is
         bool includeUptimeProof,
         uint32 messageIndex
     ) external {
-        if (!_initializeEndPoSValidation(validationID, includeUptimeProof, messageIndex)) {
+        if (
+            !_initializeEndPoSValidation(validationID, includeUptimeProof, messageIndex, address(0))
```
richardpringle, line 210: I could overload this function as well, but prefer to keep the API surface a little smaller.

geoff-vball (commented):
This is looking good! When I had originally envisioned this, I was thinking of an external function that just lets the owner of a validation/delegation change their rewards address, which would obviate the need for passing around the address in the `initialize` functions. I think your way works just as well, except if the staker wants to change the reward address in between the `initialize` and `complete` steps. Having the rewards address change in a separate step would also let the caller verify the address is correctly set before calling to disburse the rewards.

We could also set `rewardRecipient = owner` when starting a staking period, which would prevent us from having to check if the address is 0 later.

richardpringle:
> This is looking good! When I had originally envisioned this, I was thinking of an external function that just lets the owner of a validation/delegation change their rewards address, which would obviate the need for passing around the address in the `initialize` functions. I think your way works just as well, except if the staker wants to change the reward address in between the `initialize` and `complete` steps. Having the rewards address change in a separate step would also let the caller verify the address is correctly set before calling to disburse the rewards.
> 
> We could also set `rewardRecipient = owner` when starting a staking period, which would prevent us from having to check if the address is 0 later.

I still need to add the functionality to change the reward-recipient between _initialize_ and _complete_. 

As for always setting the reward-recipient (defaulting to the owner), it's cheaper not to set it. I'm happy to do things either way though. It's one of those things where we save the extra storage where it isn't necessary or we make the cost consistent (no matter where you withdraw the reward)

richardpringle:
I think all that's left is exposing the `changeRewardRecipient` functions in the interface and some documentation.

richardpringle:
> I think all that's left is exposing the `changeRewardRecipient` functions in the interface and some documentation.

Oh actually, I think I need a couple more tests. Will leave comments in the code.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +258,43 @@ abstract contract PoSValidatorManager is
...
+    ) external {
+        // Ignore the return value here to force end validation, regardless of possible missed rewards
+        _initializeEndPoSValidation(validationID, includeUptimeProof, messageIndex, rewardRecipient);
+    }
+
+    function changeRewardRecipient(bytes32 validationID, address rewardRecipient) external {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        if ($._posValidatorInfo[validationID].owner != _msgSender()) {
+            revert UnauthorizedOwner(_msgSender());
```
richardpringle, line 281: Need a test to make sure this is getting hit

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +258,43 @@ abstract contract PoSValidatorManager is
...
+
+    function changeDelegatorRewardRecipient(
+        bytes32 delegationID,
+        address rewardRecipient
+    ) external {
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        if ($._delegatorStakes[delegationID].owner != _msgSender()) {
+            revert UnauthorizedOwner(_msgSender());
+        }
```
richardpringle, line 295: Need a test to make sure this is getting triggered correctly

contracts/validator-manager/interfaces/IPoSValidatorManager.sol
```diff
@@ -151,6 +151,17 @@ interface IPoSValidatorManager is IValidatorManager {
...
+     * @notice See {IPoSValidatorManager-initializeEndValidation} for details of the first three parameters
+     * @param recipientAddress The address to receive the rewards
+     */
+    function initializeEndValidation(
+        bytes32 validationID,
+        bool includeUptimeProof,
+        uint32 messageIndex,
+        address recipientAddress
+    ) external;
+
```
richardpringle, line 164: Need to add the `initializeEndDelegation` and both the `change` functions here

contracts/validator-manager/tests/PoAValidatorManagerTests.t.sol
```diff
@@ -61,10 +61,26 @@ contract PoAValidatorManagerTest is ValidatorManagerTest {
...
     }
 
+    function _forceInitializeEndValidation(
+        bytes32 validationID,
+        bool,
+        address
+    ) internal virtual override {
+        return app.initializeEndValidation(validationID);
+    }
+
```
richardpringle, line 83: 🤔 Why did I do this?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +258,43 @@ abstract contract PoSValidatorManager is
...
+        bytes32 validationID,
+        bool includeUptimeProof,
+        uint32 messageIndex,
+        address rewardRecipient
+    ) external {
+        // Ignore the return value here to force end validation, regardless of possible missed rewards
+        _initializeEndPoSValidation(validationID, includeUptimeProof, messageIndex, rewardRecipient);
+    }
+
+    function changeRewardRecipient(bytes32 validationID, address rewardRecipient) external {
```
geoff-vball, line 277: I think we should name this `changeValidatorRewardRecipient` to avoid confusion with anyone thinking it might also work for delegators.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +258,43 @@ abstract contract PoSValidatorManager is
...
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        if ($._posValidatorInfo[validationID].owner != _msgSender()) {
+            revert UnauthorizedOwner(_msgSender());
+        }
+
+        $._rewardRecipients[validationID] = rewardRecipient;
```
geoff-vball, line 284: With the current implementation, I think we should check that `rewardRecipient != address(0)`, since setting it to `address(0)` has the counterintuitive function of sending the rewards to the owner.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +259,62 @@ abstract contract PoSValidatorManager is
...
+        }
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
+
+        if ($._posValidatorInfo[validationID].owner != _msgSender()) {
+            revert UnauthorizedOwner(_msgSender());
+        }
+
+        if (rewardRecipient == _msgSender()) {
+            delete $._rewardRecipients[validationID];
```
richardpringle, line 293: Need to test this case as well

contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -602,16 +612,113 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
...
+            expectedValidatorFees: expectedValidatorFees,
+            expectedDelegatorReward: expectedDelegatorReward,
+            validatorWeight: DEFAULT_WEIGHT,
+            expectedValidatorWeight: DEFAULT_WEIGHT,
+            expectedNonce: 2,
+            rewardRecipient: rewardRecipient
+        });
+    }
+
+    function testChangeDelegatorRewardRecipient() public {
```
richardpringle, line 677: Need to simplify this test

contracts/validator-manager/tests/PoSValidatorManagerTests.t.sol
```diff
@@ -1054,7 +1177,83 @@ abstract contract PoSValidatorManagerTest is ValidatorManagerTest {
...
+        _completeEndValidationWithChecks({
+            validationID: validationID,
+            validatorOwner: address(this),
+            expectedReward: expectedReward,
+            validatorWeight: DEFAULT_WEIGHT,
+            rewardRecipient: rewardRecipient
+        });
+    }
+
+    function testChangeRewardRecipient() public virtual {
```
richardpringle, line 1221: I left comments here for a couple of different test cases, but this test is overly complicated. The `*WithNon<Validator|Delegator>RewardRecipient` tests already test that when the field is set, so I need to just check if the field is set

contracts/validator-manager/tests/PoAValidatorManagerTests.t.sol
richardpringle: done!

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -600,6 +743,9 @@ abstract contract PoSValidatorManager is
 
             uint256 reward = _calculateDelegationReward(delegator);
             $._redeemableDelegatorRewards[delegationID] = reward;
+            if (rewardRecipient != address(0)) {
```
iansuvak, line 746: Is there a reason both this block and the below one can't be repeated just once outside of this `if/else`?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -296,9 +389,18 @@ abstract contract PoSValidatorManager is
         }
 
         address owner = $._posValidatorInfo[validationID].owner;
+
+        address rewardRecipient = $._rewardRecipients[validationID];
+
+        if (rewardRecipient == address(0)) {
+            rewardRecipient = owner;
+        } else {
+            delete $._rewardRecipients[validationID];
```
iansuvak, line 398: The logic for mapping cleanup looks good to me now but would it make sense to add checks to tests that they are deleted as expected?

Ideally we would also have a data structure on top and be able to verify that they are empty as well but that's definitely beyond the scope of this PR. (although it would enable for looping as well to end all delegations)

iansuvak (commented):
Looks good to me overall! Just a couple of minor questions and a method name typo.

contracts/validator-manager/PoSValidatorManager.sol
iansuvak: Since these are mappings getters won't help right? I was thinking of just asserting that any key that was used during the test is `address(0)` at the end of it but not strictly necessary.

contracts/validator-manager/PoSValidatorManager.sol
richardpringle: I added to the `_calculateDelegationReward` function instead. Otherwise, we would have to do the same check for the status, which is a little more fragile IMO (if the condition changes in the if/else, the author has to remember to update the next block). 

In any case, I reduced the repetition.

contracts/validator-manager/interfaces/IPoSValidatorManager.sol
```diff
@@ -247,4 +290,8 @@ interface IPoSValidatorManager is IValidatorManager {
      * @param validationID The ID of the validation period being ended.
      */
     function claimDelegationFees(bytes32 validationID) external;
+
+    function changeValidatorRewardRecipient(bytes32 validationID, address recipient) external;
+
+    function changeDelegatorRewardRecipient(bytes32 delegationID, address recipient) external;
```
geoff-vball, line 296: Can you add natspec docs here? Should be pretty easy to copy the format from the other functions in this file

tests/utils/validator_manager.go
```diff
@@ -1059,7 +1059,7 @@ func InitializeEndNativeValidation(
 ) *types.Receipt {
 	opts, err := bind.NewKeyedTransactorWithChainID(senderKey, subnet.EVMChainID)
 	Expect(err).Should(BeNil())
-	tx, err := stakingManager.InitializeEndValidation(
+	tx, err := stakingManager.InitializeEndValidation0(
```
geoff-vball, line 1062: Oopsie typo here and line 1164

contracts/validator-manager/interfaces/IPoSValidatorManager.sol
```diff
@@ -201,6 +223,16 @@ interface IPoSValidatorManager is IValidatorManager {
         bool includeUptimeProof,
         uint32 messageIndex
     ) external;
+    /**
```
geoff-vball, line 226: nit: newline

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -645,13 +790,21 @@ abstract contract PoSValidatorManager is
...
             stakingStartTime: delegator.startedAt,
             stakingEndTime: delegationEndTime,
             uptimeSeconds: $._posValidatorInfo[delegator.validationID].uptimeSeconds
         });
+
+        $._redeemableDelegatorRewards[delegationID] = reward;
+
+        if (rewardRecipient != address(0)) {
+            $._delegatorRewardRecipients[delegationID] = rewardRecipient;
+        }
```
geoff-vball, line 805: Would it maybe make sense to set this in `initializeEndDelegation`, so that we don't have to pass the parameter down another level?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +259,62 @@ abstract contract PoSValidatorManager is
...
+
+    function changeValidatorRewardRecipient(
+        bytes32 validationID,
+        address rewardRecipient
+    ) external {
+        if (rewardRecipient == address(0)) {
+            revert InvalidRewardRecipient(rewardRecipient);
+        }
+
+        PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
```
geoff-vball, line 286: Idiomatically we always place this at the top of the function before any checks. If a call reverts, it uses all the gas anyway, so there are no potential savings.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -227,7 +259,62 @@ abstract contract PoSValidatorManager is
...
+
+        if ($._delegatorStakes[delegationID].owner != _msgSender()) {
+            revert UnauthorizedOwner(_msgSender());
+        }
+
+        if (rewardRecipient == _msgSender()) {
+            delete $._delegatorRewardRecipients[delegationID];
+        } else {
+            $._delegatorRewardRecipients[delegationID] = rewardRecipient;
+        }
```
geoff-vball, line 317: It might be worth encapsulating this logic in a private function and having the external func call the private func. Then we could call the private func from `initializeEndDelegation`. Then we wouldn't have the different logical handling that I mentioned in another comment.

contracts/validator-manager/PoSValidatorManager.sol
richardpringle: What if the address is 0, though?

geoff-vball: Either ordering gives exactly the same effects in all cases, since reverted transactions cause all gas to be use, and any side-effects are not saved. Happy to explain more if that doesn't explain it fully

tests/utils/validator_manager.go
geoff-vball: Ohhhh shoot. Yeah my bad! Thanks :)

bernard-avalabs:
> As for always setting the reward-recipient (defaulting to the owner), it's cheaper not to set it. I'm happy to do things either way though. It's one of those things where we save the extra storage where it isn't necessary or we make the cost consistent (no matter where you withdraw the reward)

My personal preference is to set the reward-recipient to the owner by default. It is a bit less error prone since we won't need to special case the 0 address. I imagine the storage cost is pretty small.

bernard-avalabs (commented):
Generally looks good to me. I left a comment about preferring to set the owner as the default reward recipient, but I'm fine either way.

contracts/validator-manager/PoSValidatorManager.sol
richardpringle: I actually just changed this because of [this comment](https://github.com/ava-labs/teleporter/pull/647#discussion_r1852348458). 

Is there a problem passing this down? More expensive? I can revert my change, but this definitely is cleaner (ignoring gas cost)

richardpringle:
> > As for always setting the reward-recipient (defaulting to the owner), it's cheaper not to set it. I'm happy to do things either way though. It's one of those things where we save the extra storage where it isn't necessary or we make the cost consistent (no matter where you withdraw the reward)
> 
> My personal preference is to set the reward-recipient to the owner by default. It is a bit less error prone since we won't need to special case the 0 address. I imagine the storage cost is pretty small.

🤔 I actually agree, I think that's the cleaner solution. However... this works, so I think it would make sense to do in a follow up (locking in the tests and everything now). 

Thoughts @geoff-vball and @bernard-avalabs?

contracts/validator-manager/PoSValidatorManager.sol
richardpringle: See [this response](https://github.com/ava-labs/teleporter/pull/647#issuecomment-2492331512). TL;DR I think I could clean that up in a follow up, but I think this works and solves the problem, can we land this PR before it diverges any further from main?

contracts/validator-manager/PoSValidatorManager.sol
geoff-vball: Yeah no, this is good :)

geoff-vball (approved):

PR #650, opened 2024-11-19
Enforce uint64max weights

cam-schultz:
## Why this should be merged
Enforces that validator weights cannot exceed uint64max, which is the validator weight cap enforced on the P-Chain.

## How this works
Adds an explicit check in `_initializeValidatorRegistration`. Note that the churn tracker also enforces this bound, but an explicit check is more readable and easier to test functionally.

Also updates all variables/fields/params that represent weights to use `uint64`

## How this was tested
New unit test

## How is this documented
N/A


tests/utils/validator_manager.go
```diff
@@ -346,7 +346,7 @@ func InitializeNativeTokenValidatorSet(
 		validationIDs = append(validationIDs, subnetInfo.SubnetID.Append(uint32(i)))
 	}
 
-	Expect(initialValidatorCreatedEvent.Weight).Should(Equal(new(big.Int).SetUint64(nodes[0].Weight)))
+	Expect(initialValidatorCreatedEvent.Weight).Should(Equal(nodes[0].Weight))
```
iansuvak, line 349: it's a nice side effect that these are now primitive types

iansuvak (approved):

geoff-vball (commented):
I love how much this code simplifies by making the go params ints!

Can we add documentation, stating that the reason we cap the total weight at a `uint64` is because the p-chain does the same?

bernard-avalabs (approved):
LGTM

cam-schultz:
> Can we add documentation, stating that the reason we cap the total weight at a uint64 is because the p-chain does the same?

TBH, I think linking to the ACP-77 spec sufficiently documents this. I don't think it'd be helpful to explicitly document this data type here over the data type of any other field. Note that we do document how this limit affects staked value for the PoS case: https://github.com/ava-labs/teleporter/blob/main/contracts/validator-manager/README.md?plain=1#L72

geoff-vball (approved):

PR #654, opened 2024-11-21
P2p peer fetching compatibility

cam-schultz:
## Why this should be merged
Compatibility with https://github.com/ava-labs/awm-relayer/pull/559

**Note:** This PR uses a modified version of Avalanchego v1.12.0-fuji that relaxes the outbound peer request handling. Once this change is formalized and included in an Avalanchego release, we will circle back and update the dependency. This workaround is safe because it only impacts outbound messages from the signature aggregator peer.

## How is this documented


geoff-vball (approved):
This all looks fine to me

tests/utils/chain.go
```diff
@@ -556,28 +553,29 @@ func NewSignatureAggregator(apiUri string, subnets []ids.ID) *aggregator.Signatu
...
+		logging.NoLog{},
+		registry,
+		constants.DefaultNetworkCompressionType,
+		constants.DefaultNetworkMaximumInboundTimeout,
+	)
+	Expect(err).Should(BeNil())
+
 	appRequestNetwork, err := peers.NewNetwork(
-		logging.Info,
+		logging.Error,
```
iansuvak, line 565: (no action necessary) I assume that this is because the network now spews a lot more with P2P changes. Is `Warning` level also too loud?

iansuvak (approved):

tests/utils/chain.go
cam-schultz: I changed this because the signature aggregator is not under test in this repo's e2e test suites, so the logs seemed unnecessary.

bernard-avalabs (approved):

PR #667, opened 2024-12-03
ACP-99: Implement using inheritance

cam-schultz:
## Why this should be merged
Experimental implementation of ACP-99 using inheritance. The concrete contracts are extended to implement the new interfaces `IACP99SecurityModule` and `IACP99ValidatorManager`. 

## How is this documented


contracts/validator-manager/interfaces/IACP99ValidatorManager.sol
```diff
@@ -0,0 +1,71 @@
...
+    uint64 registrationExpiry;
+    PChainOwner remainingBalanceOwner;
+    PChainOwner disableOwner;
+}
+
+interface IACP99ValidatorManager {
+    function initializeValidatorSet(
+        ConversionData calldata conversionData,
+        uint32 messageIndex
+    ) external;
```
richardpringle, line 54: Should this necessarily be part of the interface? In theory, one should be able to deploy a contract with the initial validator set hard-coded, right?

The fact that the security interface doesn't have a matching method to this one, but it does for all others, leads me to believe that this isn't necessarily part of the interface (despite the likelihood of its existence in most instances).

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -25,6 +21,8 @@ import {ContextUpgradeable} from
     "@openzeppelin/contracts-upgradeable@5.0.2/utils/ContextUpgradeable.sol";
 import {Initializable} from
     "@openzeppelin/contracts-upgradeable@5.0.2/proxy/utils/Initializable.sol";
+import {IACP99SecurityModule} from "./interfaces/IACP99SecurityModule.sol";
```
richardpringle, line 24: Is this used anywhere?  And I also notice that this isn't `IACP99ValidatorManager`. Might make sense to just add a comment saying why (I'm guessing because it's not worth the adding all the events that aren't specified in the interface).

contracts/validator-manager/interfaces/IACP99ValidatorManager.sol
```diff
@@ -0,0 +1,71 @@
...
+ */
+struct ValidatorRegistrationInput {
+    bytes nodeID;
+    bytes blsPublicKey;
+    uint64 registrationExpiry;
+    PChainOwner remainingBalanceOwner;
+    PChainOwner disableOwner;
+}
+
+interface IACP99ValidatorManager {
```
richardpringle, line 50: Might be worth adding a comment saying that we should also include the events from `IValidatorManager`?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -174,6 +175,62 @@ abstract contract PoSValidatorManager is
...
     }
 
+    function handleInitializeValidatorRegistration(bytes32 validationID, uint64 weight, bytes calldata args) external {
+        uint256 stakeAmount = weightToValue(weight);
+        (uint16 delegationFeeBips, uint64 minStakeDuration) = abi.decode(args, (uint16, uint64));
+        _initializeValidatorRegistration(validationID, stakeAmount, delegationFeeBips, minStakeDuration);
+    }
+
+    function handleCompleteValidatorRegistration(bytes32 validationID) external {
+        // No-op
```
richardpringle, line 185: Can you add an explanation as to why this is a no-op?

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -174,6 +175,62 @@ abstract contract PoSValidatorManager is
...
+        bool success = _initializeEndPoSValidation(
+            validationID, includeUptimeProof, messageIndex, rewardRecipient
+        );
+        if (force) {
+            return;
+        }
+
+        if (!success) {
+            revert ValidatorIneligibleForRewards(validationID);
+        }
```
richardpringle, line 200: ... I know, I know, this isn't a functional change, but I can't help myself, 😉.

```suggestion
        if (!force && !success) {
            revert ValidatorIneligibleForRewards(validationID);
        }
```

contracts/validator-manager/PoSValidatorManager.sol
richardpringle, line 1: Am I correct in saying that there aren't really any changes here, just the calling convention?

contracts/validator-manager/PoAValidatorManager.sol
```diff
@@ -46,30 +45,34 @@ contract PoAValidatorManager is IPoAValidatorManager, ValidatorManager, OwnableU
...
+    function handleCompleteEndValidation(bytes32 validationID) external {
+        // No-op
+    }
+
+    function handleInitializeValidatorWeightChange(bytes32 validationID, uint64 weight, bytes calldata args) external {
+        // No-op
+    }
+
+    function handleCompleteValidatorWeightChange(bytes32 validationID, bytes calldata args) external {
+        // No-op
```
richardpringle, line 76: Wait... shouldn't POA still be able to cycle validators? Maybe I just don't have the full picture

richardpringle (commented):
Some comments may be irrelevant; don't be afraid to say so!

contracts/validator-manager/PoAValidatorManager.sol
cam-schultz: Yeah, these handlers should likely implement checks against the owner address/key. Keep in mind this PR is meant to demonstrate the architectural changes needed to implement ACP-99, so missing features are expected

PR #668, opened 2024-12-03
ACP-99: Implement using composition

cam-schultz:
## Why this should be merged
Experimental implementation of ACP-99 using composition. Concrete contracts implement either `IACP99SecurityModule` or `IACP99ValidatorManager`. 

Notably, the contract sizes decrease significantly. `ERC20TokenStakingManager` goes from right at the 24kB size limit down to 14.6kB.

**This is not functional as-is, and is meant to demonstrate the necessary refactoring to implement ACP-99 using composition**

## How is this documented


richardpringle:
>Notably, the contract sizes decrease significantly. ERC20TokenStakingManager goes from right at the 24kB size limit down to 14.6kB.

I'm curious to know how much this affects gas costs, that would be one of the major advantages of keeping everything in a single contract, right?

contracts/validator-manager/ACP99ValidatorManager.sol
```diff
@@ -0,0 +1,60 @@
...
+
+pragma solidity 0.8.25;
+
+abstract contract ACP99ValidatorManager is IACP99ValidatorManager, ValidatorManager {
+    IACP99SecurityModule public securityModule;
+
+    // TODO: calling this should be restricted to...who?
+    function setSecurityModule(IACP99SecurityModule _securityModule) external {
+        securityModule = _securityModule;
+    }
```
richardpringle, line 18: Should the security module itself control its own upgrades? If not, should there be a `SecurityUpdateModule`?

contracts/validator-manager/ACP99ValidatorManager.sol
```diff
@@ -0,0 +1,60 @@
...
+        _initializeValidatorSet(conversionData, messageIndex);
+    }
+
+    function initializeValidatorRegistration(
+        ValidatorRegistrationInput calldata input,
+        uint64 weight,
+        bytes calldata args
+    ) external returns (bytes32){
+        bytes32 validationID = _initializeValidatorRegistration(input, weight);
+        securityModule.handleInitializeValidatorRegistration(validationID, weight, args);
```
richardpringle, line 33: I definitely need to go through the ACP again, but I would expect the security module to do some form of validation on the inputs _before_ calling the `_initializeValidatorRegistration`. Or does it not matter since the security module will just revert anyway?

richardpringle (commented):
What's the key difference here, that the security module is a property of the validator-manager instead of the validator-manager implementing the security module interface?

contracts/validator-manager/ACP99ValidatorManager.sol
cam-schultz: It's a little tricky to reverse the order in our current implementation since the security module needs the validationID, which is provided by the manager contract. It will revert if the security module's checks fail, regardless of the ordering. We should definitely keep in mind patterns such as [checks-effects-interactions](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html) if/when we decide to move forward on this implementation.

PR #684, opened 2024-12-12
Relax OpenZeppelin upgrades plugin validation checks

### LINKED ISSUE #648: Make `ValidatorManager` contracts compatible with OpenZeppelin upgrades plugins
**Context and scope**
The current implementations of `ValidatorManager` (`PoAValidatorManager`, `ERC20TokenStakingManager`, `NativeTokenStakingManger`) have multiple incompatibilities with OpenZeppelin upgrades, making the safety validation fail.

This limits the reusability of the contracts outside of the `teleporter` / Avalanche CLI context (ofc developers can still disable the safety validation, but this is far from ideal).

The 2 incompatibilities are:
- The `constructor` defined in all contracts
- The `external` functions present in the `ValidatorMessages` library

as we can see with these errors raised when trying to deploy a `PoAValidatorManager` with `Upgrades.deployTransparentProxy`:
```
[FAIL: setup failed: revert: Upgrade safety validation failed:
✘  lib/teleporter/contracts/validator-manager/PoAValidatorManager.sol:PoAValidatorManager

      lib/teleporter/contracts/validator-manager/PoAValidatorManager.sol:24: Contract `PoAValidatorManager` has a constructor
          Define an initializer instead
          https://zpl.in/upgrades/error-001
      
      lib/teleporter/contracts/validator-manager/ValidatorMessages.sol: Linking external libraries like `ValidatorMessages` is not yet supported
          Use libraries with internal functions only, or skip this check with the `unsafeAllowLinkedLibraries` flag 
          if you have manually checked that the libraries are upgrade safe
          https://zpl.in/upgrades/error-006
```
Solving those doesn't seem too complicated:
1. Remove the `constructor`
2. Make the `ValidatorMessages` functions `internal`

This might be a non-issue if you want ALL deployments of such contracts to be done through the Avalanche CLI (and not via Forge).

A workaround for the external libraries check would be to add `@custom:oz-upgrades-unsafe-allow external-library-linking` to all contracts.

Another alternative is to deploy the contracts more manually (deploy the implementation, then the proxy, etc.) but this is more error-prone.

**Open questions**
- What is the impact of removing the `constructor` that uses `ICMInitializable`?
- Is there any downside to making the `ValidatorMessages` functions `internal`?

cam-schultz:
## Why this should be merged
Fixes #648 by allowing constructors and external library calls.

## How this works
Adds NatSpec comments to the upgradeable `validator-manager` contracts to bypass the `constructor` and `external-library-linking` validation rules.

## How this was tested
Ci

## How is this documented
N/A

### DIFF
--- a/contracts/validator-manager/ERC20TokenStakingManager.sol
+++ b/contracts/validator-manager/ERC20TokenStakingManager.sol
@@ -20,6 +20,7 @@ import {SafeERC20} from "@openzeppelin/contracts@5.0.2/token/ERC20/utils/SafeERC
  * @dev Implementation of the {IERC20TokenStakingManager} interface.
  *
  * @custom:security-contact https://github.com/ava-labs/icm-contracts/blob/main/SECURITY.md
+ * @custom:oz-upgrades-unsafe-allow constructor external-library-linking
  */
 contract ERC20TokenStakingManager is
     Initializable,
--- a/contracts/validator-manager/NativeTokenStakingManager.sol
+++ b/contracts/validator-manager/NativeTokenStakingManager.sol
@@ -20,6 +20,7 @@ import {Initializable} from
  * @dev Implementation of the {INativeTokenStakingManager} interface.
  *
  * @custom:security-contact https://github.com/ava-labs/icm-contracts/blob/main/SECURITY.md
+ * @custom:oz-upgrades-unsafe-allow constructor external-library-linking
  */
 contract NativeTokenStakingManager is
     Initializable,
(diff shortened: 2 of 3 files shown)


iansuvak (approved):
The change LGTM but did we confirm it actually accomplishes the goal?

cam-schultz:
> The change LGTM but did we confirm it actually accomplishes the goal?

I haven't directly. I'm hitting some confusing errors when running a forge script that calls the upgrades plugin. The NatSpec comments should apply to the entire contract , however.

That said, I'd also be fine leaving it to users to pass the relaxation flags to the plugin themselves. In which case, we would just want to document the process and leave the source code alone.

geoff-vball (approved):

bernard-avalabs (approved):

PR #694, opened 2025-01-03
remove icm-services as a go dependency

iansuvak:
## Why this should be merged
Removes circular dependency between this repo and icm-services
## How this works
Replaces direct calls to signature-aggregator with spinning up an external executable.
## How this was tested
E2E tests should pass (this change only affects E2E tests)
## How is this documented
TODO


scripts/versions.sh
```diff
@@ -29,8 +29,9 @@ function extract_commit() {
   echo "$version"
 }
 
-# AWM_RELAYER_VERSION is needed for the docker run setup, but is not a go module dependency.
-AWM_RELAYER_VERSION=${AWM_RELAYER_VERSION:-'v1.0.0'}
+# ICM_SERVICES_VERSION is needed for the E2E tests but is not a direct dependency since that would create a circular dependency.
+# ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'signature-aggregator-v1.0.0-rc.0'}
+ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'7d1e095dc84913d28c9c82806f240595e8cd4c8b'}
```
iansuvak, line 34: Right now this is pointing to a commit on https://github.com/ava-labs/icm-services/pull/616 branch. Should replace when we have the next tag

geoff-vball (approved):

bernard-avalabs (approved):

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+	. "github.com/onsi/gomega"
+)
+
+const (
+	DEFAULT_SIG_AGG_PATH = "~/.teleporter-deps/icm-services/signature-aggregator"
+	DEFAULT_API_PORT     = 9090
+	SIG_AGG_API_PATH     = "/aggregate-signatures"
+)
+
+// This is a wrapper around a signature aggregator binar instead of importing the package directly
```
cam-schultz, line 27:
```suggestion
// This is a wrapper around a signature aggregator binary instead of importing the package directly
```

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	avalancheWarp "github.com/ava-labs/avalanchego/vms/platformvm/warp"
+
+	. "github.com/onsi/gomega"
+)
+
+const (
+	DEFAULT_SIG_AGG_PATH = "~/.teleporter-deps/icm-services/signature-aggregator"
+	DEFAULT_API_PORT     = 9090
```
cam-schultz, line 23: If it's all the same, let's use port `8080` here, which is the default API port used in `icm-services`. `9090` is the default metrics port.

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+	encoder := json.NewEncoder(configFile)
+	err = encoder.Encode(cfg)
+	Expect(err).Should(BeNil())
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	cmd := exec.CommandContext(ctx, sigAggPath, "--config-file", configFile.Name())
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err = cmd.Start()
```
cam-schultz, line 99: We could start a goroutine here that calls `cmd.Wait` and emits and error if we get anything unexpected. We do this in the `icm-services` executable util [here](https://github.com/ava-labs/icm-services/blob/25d5c7f6c877067c998b49bdb8b40f1efa2445de/tests/utils/utils.go#L595).

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+	Expect(err).Should(BeNil())
```
cam-schultz, line 100: We may also want to check the health endpoint before returning, similar to https://github.com/ava-labs/icm-services/blob/25d5c7f6c877067c998b49bdb8b40f1efa2445de/tests/utils/utils.go#L603

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+	req.Header.Set("Content-Type", "application/json")
+	res, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("expected status code 200, got %d", res.StatusCode)
+	}
+
+	defer res.Body.Close()
```
cam-schultz, line 147: Let's move this above the `res.StatusCode` so that it's called on all non-nil codepaths.

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+	SignedMessage string `json:"signed-message"`
+}
+
+func (s *SignatureAggregator) Shutdown() error {
+	s.cancelFunc()
+	return s.cmd.Wait()
+}
+
+// Aggregator utils
+func NewSignatureAggregator(apiUri string, l1IDs []ids.ID) *SignatureAggregator {
```
cam-schultz, line 65: Let's use `subnetID` instead of `l1ID` to align with https://github.com/ava-labs/icm-contracts/pull/695

scripts/install_sig_agg_release.sh
```diff
@@ -0,0 +1,122 @@
...
+ICM_SERVICES_DOWNLOAD_PATH=${BASEDIR}/icm-services-linux-${GOARCH}-${ICM_SERVICES_VERSION}.tar.gz
+
+if [[ ${GOOS} == "darwin" ]]; then
+  ICM_SERVICES_DOWNLOAD_URL=https://github.com/ava-labs/icm-services/releases/download/${ICM_SERVICES_VERSION}/icm-services_${ICM_SERVICES_VERSION#v}_darwin_${GOARCH}.tar.gz
+  ICM_SERVICES_DOWNLOAD_PATH=${BASEDIR}/icm-services-darwin-${GOARCH}-${ICM_SERVICES_VERSION}.tar.gz
+fi
+
+BUILD_DIR=${ICM_SERVICES_BUILD_PATH}-${ICM_SERVICES_VERSION}
+
+extract_archive() {
```
cam-schultz, line 35: [Not a blocker] This function is very similar to the functions of the same name in the Avalanchego and Subnet EVM install scripts. We should combine them. Feel free to defer to a later issue.

tests/utils/signature_aggregator.go
```diff
@@ -0,0 +1,170 @@
...
+	sigAggPath := os.Getenv("SIG_AGG_PATH")
+	if sigAggPath == "" {
+		sigAggPath = DEFAULT_SIG_AGG_PATH
```
cam-schultz, line 68: I have a slight preference to treat this as an error rather than falling back to a default path. If for some reason, the env var `SIG_AGG_PATH` is not set as expected, it seems likely that `DEFAULT_SIG_AGG_PATH` could already contain a binary with an unknown version, which would be difficult to debug.

I'd prefer that the executable path be passed as an argument, and all OS interfacing be done at the test-suite level rather than in utils, but I understand that would require some additional refactoring. Happy to defer, but could you leave a TODO comment to revist?

tests/utils/signature_aggregator.go
iansuvak: Yup I switched it once when debugging it not starting up before realizing there was a zombie process and never switched it back. Thanks for the catch.

tests/utils/signature_aggregator.go
iansuvak: ✅  but had to remove the cmd.Wait from Shutdown

tests/utils/signature_aggregator.go
geoff-vball: Yeah, I think this is a good point. I don't think the default path should live in `utils`. Passing in from the test suite would be a good idea, and the test suite can log whether it's using the default path or one provided by the env.

cam-schultz (approved):

geoff-vball (approved):

PR #695, opened 2025-01-03
Rename l1ID -> subnetID

cam-schultz:
## Why this should be merged

Partially reverts the [recent nomenclature changes](https://github.com/ava-labs/icm-contracts/pull/652) to continue to use `subnetID` instead of `l1ID`.

`subnetID` should be used to refer to both Subnets that have not been converted, and L1s that have. It's not as explicit as it could be, but note that the ACP-77 message specifications use `subnetID` to identify the converted L1. Also note that the `subnetID` is the tx hash of the `CreateSubnetTx`, and therefore does not change when the Subnet is converted to an L1. As such, changing the name of the identifier of a static value depending on the Subnet/L1 status would be confusing in its own right. Eventually, once Subnets are fully deprecated in favor of L1s, it will make more sense to revisit this terminology.

## How this works
Renames all instances of `l1ID` to `subnetID`

## How this was tested
CI

## How is this documented
Adds a note describing this nomenclature, as it jmay be confusing why validator manager contracts that by definition only apply to L1s would use the Subnet terminology.


bernard-avalabs (approved):
LGTM. The comment about the naming scheme is very helpful.

geoff-vball (approved):

iansuvak (approved):

PR #700, opened 2025-01-10
ACP 99 reference implementation

cam-schultz:
## Why this should be merged
Implements ACP-99 as described in https://github.com/avalanche-foundation/ACPs/pull/170

## How this works
Adds `ACP99Manager` to the class hierarchy,. `ValidatorManager` implements `ACP99Manager`.

## How this was tested
CI, fixed breaking tests

## How is this documented
Updates README to point to ACP-99

## Future work

The following items are candidates to be included in this PR. They may also be deferred to later changes.
- Remove `IValidatorManager` altogether, as `ACP99Manager` does the same job of defining the core validator manager functionality
- Rename all `initialize*` methods to `initiate*` to match ACP-99. `initialize` is a misnomer, since nothing is initialized as part of these function calls. Rather, they `initiate` the process of registering/modifying a validator.


contracts/validator-manager/ValidatorManager.sol
richardpringle, line 1: I'm guessing we'll want to fix those reentrancy warnings?

contracts/validator-manager/interfaces/IValidatorManager.sol
richardpringle, line 1: 👍

contracts/validator-manager/ValidatorManager.sol
cam-schultz: This same "reentrancy" (`ValidationPeriodCreated` emitted after the call to the Warp precompile) was already present. This isn't really a concern, since the worst case scenario if reentrancy occurred is that the same event is emitted multiple times. That would require modification to the Warp precompile, which is that is the case, there's much bigger concerns than duplicated events. I don't remember exactly if/when we dismissed this warning in the past, but it's likely retriggered due to source code changes.

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -327,7 +323,8 @@ abstract contract PoSValidatorManager is
     ) internal returns (bool) {
         PoSValidatorManagerStorage storage $ = _getPoSValidatorManagerStorage();
 
-        Validator memory validator = _initializeEndValidation(validationID);
+        _initiateValidatorRemoval(validationID);
+        Validator memory validator = getValidator(validationID);
```
cam-schultz, line 327: The order of these calls matters, since `_initiateValidatorRemoval` modifies the `Validator`. I removed the `Validator` return value from the `ACP99Manager` interface since it was not strictly necessary, and can otherwise be accessed like it is here. If the internal `ACP99Manager` functions do return a `Validator`, it increases the `ERC20TokenStakingManager` contract size by ~900 bytes, surpassing the size limit.

I settled on adding a `@dev` comment to the `ValidatorManager` implementation of these methods to warn developers that this function mutates accessible state.

richardpringle: You should probably add a little blurb inline here. 
```solidity
// order matters for the next two lines, see...
```
Then reference an explanation somewhere? Ideally in a github permalink or something else that is guaranteed not to change. That is unless you can make the comment succinct, haha (I can't).

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -595,25 +608,22 @@ abstract contract PoSValidatorManager is
...
-        // (which implicitly includes the delegation's weight update) to be used to complete delisting
-        // for an earlier delegation. This is necessary because the P-Chain is only willing to sign the latest weight update.
-        if (validator.messageNonce < nonce || delegator.startingNonce > nonce) {
-            revert InvalidNonce(nonce);
+        // If we've already received a weight update with a nonce greater than the delegation's starting nonce,
+        // then there's no requirement to include an ICM message in this function call.
+        if (validator.receivedNonce < delegator.startingNonce) {
+            (bytes32 messageValidationID, uint64 nonce) =
+                completeValidatorWeightUpdate(messageIndex);
+            if (validationID != messageValidationID) {
```
richardpringle, line 616: nit: easier to read if this is broken up
```suggestion
            (bytes32 messageValidationID, uint64 nonce) =
                completeValidatorWeightUpdate(messageIndex);

            if (validationID != messageValidationID) {
```

richardpringle (approved):

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -595,25 +608,22 @@ abstract contract PoSValidatorManager is
...
+                revert InvalidValidationID(delegator.validationID);
```
geoff-vball, line 617: I think we should probably output both the validation ID of the message, as well as the validation ID of the delegator here. (I realize this is the old behaviour, so feel free to punt this)

contracts/validator-manager/PoSValidatorManager.sol
bernard-avalabs: I agree that a blurb should be added. I think it would be difficult for someone to discover the ordering constraint here without one.

bernard-avalabs (commented):
Generally makes sense to me.

contracts/validator-manager/PoSValidatorManager.sol
cam-schultz: I added a comment above the call to `getValidator` here. I also added a `@dev` natspec comment to all of the methods in `ValidatorManager` that touch the internal validator state.

contracts/validator-manager/PoSValidatorManager.sol
cam-schultz: I added a new error type `UnexpectedValidationID` for cases when the provided ID does not match the expected one. The unary `InvalidaValidationID` is still used in the resend methods for cases where the provided validation ID doesn't correspond to a registered validation.

cam-schultz:
See [here](https://github.com/avalanche-foundation/ACPs/pull/170#discussion_r1929347147) for context on the `nodeID` format in events, as implemented in  https://github.com/ava-labs/icm-contracts/pull/700/commits/0f7039f33077091276dce6c2c8deb982af34e95d

iansuvak (approved):
LGTM. Excited to have the reference implementation merged and it's an improvement over the current in comments and function naming as well 👍

PR #704, opened 2025-01-16
Validator params getters

cam-schultz:
## Why this should be merged
Adds a getter to `PoSValidatorManager` to expose a subset of the parameters used to initialize the contract. These settings are useful for potential validators and delegators in order to pass validation checks when registering. At present, doing so requires some amount of trial and error, or prior knowledge of these settings.

I also tried returning the entire `PoSValidatorManagerSettings`, but that caused `ERC20TokenStakingManager` to exceed the contract size limit.

## How this works
Adds `getValidatorParameters` to `PoSValidatorManager`

## How is this documented


geoff-vball (approved):

richardpringle (approved):
Nice to review a smaller PR haha. 
🥇

iansuvak (approved):

contracts/validator-manager/PoSValidatorManager.sol
```diff
@@ -179,6 +179,37 @@ abstract contract PoSValidatorManager is
...
+     */
+    function getValidatorParameters()
+        public
+        view
+        returns (
+            uint256 minimumStakeAmount,
+            uint256 maximumStakeAmount,
+            uint64 minimumStakeDuration,
+            uint16 minimumDelegationFeeBips,
+            uint64 maximumStakeMultiplier
```
michaelkaplan13, line 200: Have we used named returned values in the past? IIUC, at one point we decided to not use them anywhere for consistency, but I could be misremembering. I don't think they're needed here either way though since we don't use the variable names in the implementation.

cam-schultz: I don't recall that exact conversation, but IMO we shouldn't be rigid about named return parameters, since they're useful in some places, but not in others. I prefer them here since nothing else in the function signature makes it obvious what the values are. It also allows for thorough natspec comments on this getter. Without them, a caller would have to read the implementation, and likely have to find the underlying storage variable for full context.

bernard-avalabs (approved):

geoff-vball:
@cam-schultz is this PR relevant anymore after the audit fixes?

cam-schultz:
> @cam-schultz is this PR relevant anymore after the audit fixes?

Nope! Closing this out.

PR #705, opened 2025-01-17
Replace Snyk with Semgrep

chriselbring-avalabs:
## How is this documented


geoff-vball (approved):

iansuvak (approved):
LGTM. I'm guessing the `if: always()` here is a better way of ensuring that the step runs instead of the `continue-on-error: true` in the deleted file?

geoff-vball:
@iansuvak Yeah, `continue-on-error` basically says "If this steps fails, keep going", whereas `if: always()` says "No matter what steps have failed before this, run this step"

PR #714, opened 2025-02-04
Bump avago subnet evm

iansuvak:
## Why this should be merged

Updates avago and subnet-evm to latest versions.

## How this works

Newest changes in avago made tmpnet node ports dynamic on restart. The workaround here explicitly fixes them when restarting. This should be removed when releases are cut to re-enable static local tmpnet node ports. 

## How this was tested
E2E

## How is this documented


scripts/versions.sh
```diff
@@ -31,12 +31,10 @@ function extract_commit() {
 
 # ICM_SERVICES_VERSION is needed for the E2E tests but is not a direct dependency since that would create a circular dependency.
 # ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'signature-aggregator-v1.0.0-rc.0'}
-ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'303b307e816364f62ef2905694ac6d0652323248'}
+ICM_SERVICES_VERSION=${ICM_SERVICES_VERSION:-'1ff2e4f1313e5d0a4961cc6dd680b27d9331fa1f'}
 
 # Don't export them as they're used in the context of other calls
 AVALANCHEGO_VERSION=${AVALANCHEGO_VERSION:-$(extract_commit "$(getDepVersion github.com/ava-labs/avalanchego)")}
-# Temporarily hardcode the Avalanchego version until outbound networking relaxation is available
-AVALANCHEGO_VERSION=v1.12.0
```
iansuvak, line 39: This is not enabled by the update of avago but was possible since the switch to using an external binary for the signature aggregation in tests.

tests/network/network.go
```diff
@@ -495,6 +499,15 @@ func (n *LocalNetwork) SetChainConfigs(chainConfigs map[string]string) {
...
 	}
 
+	// Apply node config flags to explicitly set ports to their existing values
+	// Otherwise nodes might start up with dynamic ports making existing clients and bound contracts
+	// unusable
+	// TODO: remove once tmpnet supports static port option across restarts
+	for _, tmpnetNode := range n.Network.Nodes {
+		port := getTmpnetNodePort(tmpnetNode)
+		tmpnetNode.Flags[config.HTTPPortKey] = port
+	}
```
cam-schultz, line 509: Do we also have to do this ahead of the network restart in `ConvertSubnet`?

iansuvak: Yeah we do the other diff on line 270 is inside `ConvertSubnet`

cam-schultz: D'oh not sure how I missed that

cam-schultz (approved):

tests/network/network.go
```diff
@@ -537,3 +550,12 @@ func (n *LocalNetwork) GetTwoL1s() (
 	Expect(len(l1s)).Should(BeNumerically(">=", 2))
 	return l1s[0], l1s[1]
 }
+
+// TODO: once tmpnet supports static port option across restarts, remove this function
```
michaelkaplan13, line 554: Is this something that is planned to be added on tmpnet? I forget the context on that side, but I thought this support was actually removed since it's no longer needed.

michaelkaplan13 (approved):
Makes sense to me. Just not sure if `tmpnet` is planning on supporting static ports across restarts at all going foward.

tests/network/network.go
iansuvak: Yeah -- it's already added back but requiring explicit configuration:  https://github.com/ava-labs/avalanchego/pull/3697 not including it in this PR since would need to be on branches again and this PR allows us to use properly tagged releases.

PR #729, opened 2025-03-04
Update foundry

geoff-vball:
## Why this should be merged
We had previous created our own fork of foundry because foundry did not release stable versions. They have now released `v1.0.0`, so we can update to that version.

## How is this documented


bernard-avalabs (approved):

scripts/install_foundry.sh
cam-schultz, line 1: [foundryup](https://github.com/foundry-rs/foundry/tree/master/foundryup#usage) supports installing a specific version like so:

```
curl -L https://foundry.paradigm.xyz | bash
foundryup --install ${FOUNDRY_VERSION}
```

However, I noticed that this interface has already changed since `v1`, so pinning the foundryup version like we are here might be the best option after all.

README.md
```diff
@@ -21,14 +21,14 @@ For help getting started with building ICM contracts, refer to [the avalanche-st
 ### Dependencies
 
 - [Ginkgo](https://onsi.github.io/ginkgo/#installing-ginkgo) for running the end-to-end tests.
-- [Foundry](https://book.getfoundry.sh/) Use `./scripts/install_foundry.sh` to install the Ava Labs [fork](https://github.com/ava-labs/foundry) for building contracts.
+- [Foundry](https://book.getfoundry.sh/) Use `./scripts/install_foundry.sh` to install foundry for building contracts.
```
cam-schultz, line 24:
```suggestion
- [Foundry](https://book.getfoundry.sh/) Use `./scripts/install_foundry.sh` to install Foundry for building contracts.
```
nit

contracts/README.md
```diff
@@ -1,14 +1,17 @@
-## Contracts
-This directory contains Solidity contracts that leverage Avalanche Interchain Messaging (ICM) to implement unique cross-chain functions. 
+# Contracts
 
-This repository is set up as a [Foundry](https://github.com/foundry-rs/foundry) project. Use the `scripts/install_foundry.sh` to install the correct version of the ava-labs fork of foundry. 
+This directory contains Solidity contracts that leverage Avalanche Interchain Messaging (ICM) to implement unique cross-chain functions.
+
+This repository is set up as a [Foundry](https://github.com/foundry-rs/foundry) project. Use the `scripts/install_foundry.sh` to install the correct version of foundry.
```
cam-schultz, line 5:
```suggestion
This repository is set up as a [Foundry](https://github.com/foundry-rs/foundry) project. Use the `scripts/install_foundry.sh` to install the correct version of Foundry.
```
nit

scripts/install_foundry.sh
geoff-vball: The problem with the above method is that it pulls `foundryup` from `master`, which isn't pinned. I spent days debugging a breaking change with it before, so I think keeping `foundryup` pinned as well is the best method.

cam-schultz (approved):

cam-schultz (approved):

PR #734, opened 2025-03-06
Support migration from V1 PoA

cam-schultz:
## Why this should be merged
Migrating from the v1 contracts has two main challenges:
1) The storage layout changed due to updated to the `Validator` struct
2) The state previously stored in a single contract must be migrated to a second contract

This PR supports (1) by adding a method that is callable once per active `validationID` that migrates from the old `Validator` struct to the new `Validator` struct.

(2) is **not** supported by this PR. 

As such, this change supports migrating from V1 `PoAValidatorManager` instances to the new `ValidatorManager`, which when deployed on its own, performs the same functionality.

## How this works
Adds the method `migrateFromV1`. This method is only callable once per `validationID`, which is enforced by marking the legacy validator's status as `Unknown` after migrating. Any attempted validator updates will fail before the `validationID` is migrated, since the `_validationPeriods` mapping occupies a new storage slot.

## How this was tested
Tested manually on a local network. E2E test in https://github.com/ava-labs/icm-contracts/pull/736

## How is this documented
Added a migration guide


iansuvak (approved):

iansuvak (approved):

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -138,6 +151,37 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
...
+     */
+    function migrateFromV1(bytes32 validationID, uint32 receivedNonce) external {
+        ValidatorManagerStorage storage $ = _getValidatorManagerStorage();
+        ValidatorLegacy storage legacy = $._validationPeriodsLegacy[validationID];
+        if (legacy.status == ValidatorStatus.Unknown) {
+            revert InvalidValidationID(validationID);
+        }
+        if (receivedNonce > legacy.messageNonce) {
+            revert InvalidNonce(receivedNonce);
+        }
```
geoff-vball, line 167: I'm going back and forth on what I think about this... Since this is a publicly accessible function, users could input any nonce between 0 and `legacy.messageNonce`. Would it be a better UX to just set all `receivedNonce` to 0 since they don't really matter anyway?

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -138,6 +151,37 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
...
+            sentNonce: legacy.messageNonce,
+            receivedNonce: receivedNonce,
+            weight: legacy.weight,
+            startTime: legacy.startedAt,
+            endTime: legacy.endedAt
+        });
+
+        // Set the legacy status to unknown to disallow future migrations.
+        legacy.status = ValidatorStatus.Unknown;
+        $._validationPeriodsLegacy[validationID] = legacy;
```
geoff-vball, line 182: Is it cheaper/easier here to write directly `$._validationPeriodsLegacy[validationID].status = ValidatorStatus.Unknown;`?

geoff-vball (approved):
Just a few small things. Feel free to merge whenever you've addressed/considered them

contracts/validator-manager/MigratingFromV1.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Migrating Proof-of-Authority
+
+In the V2 contracts, Proof-of-Authority validator management is implemented fully in `ValidatorManager`, so there is no need to deploy a separate contract (contrast with the Proof-of-Stake case, which requires a standalone `StakingManager` deployment). Additionally, both the V1 `PoAValidatorManager` and the V2 `ValidatorManager` are `Ownable`, meaning that the V1 `PoAValidatorManager` owner will persist after upgrading to a V2 `ValidatorManager`. Migrating from a V1 to V2 therefore only requires migrating the state variables that track validators. The helper method `migrateFromV1` is provided for this purpose. The general steps are as follows:
+
+1. Upgrade the proxy contract's implementation to a newly deployed V2 `ValidatorManager` using standard methods
+2. For each validator (active or expired), call `migrateFromV1` with its `validationID`. The `receivedNonce` argument can be set to 0, since `PoAValidatorManager`s do not support weight changes.
```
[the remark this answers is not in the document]
cam-schultz: Unless I'm missing something, the only methods the V1 `PoAValidatorManager` exposes are initialize/complet validator registration/removal 
https://github.com/ava-labs/icm-contracts/blob/validator-manager-v1.0.0/contracts/validator-manager/PoAValidatorManager.sol 
https://github.com/ava-labs/icm-contracts/blob/validator-manager-v1.0.0/contracts/validator-manager/ValidatorManager.sol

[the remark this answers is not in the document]
cam-schultz: I changed this to "V1 `PoAValidatorManager`" to clarify this

contracts/validator-manager/ValidatorManager.sol
cam-schultz: Agree that's cleaner.

contracts/validator-manager/ValidatorManager.sol
cam-schultz: I elected to take `receivedNonce` as an argument to make this method as general as possible. Even though migrating from PoA is the only documented workflow, I wanted to leave the door open to using this method for other migration workflows. Since `receivedNonce` is always unused for the PoA case, it doesn't matter if it's set to 0, `legacy.messageNonce`, or anywhere in between.

geoff-vball (approved):

contracts/validator-manager/MigratingFromV1.md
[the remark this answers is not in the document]
geoff-vball: Ooo yeah you're right. I thought we had the ability to change the weight

PR #741, opened 2025-03-12
Emit justification info

cam-schultz:
## Why this should be merged

P-Chain validators attest that a validator is inactive and will never become active by signing a `L1ValidatorRegistrationMessage` with the `registered` field set to false. Since such validators are not held in the P-Chain's active state, P-Chain validators require an [ACP-118](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/118-warp-signature-request#specification) `justification` be provided along with the message to sign. At present, it's difficult to construct these `justification`s on-demand, since that information is not exposes on-chain or emitted by any event. This effectively requires validator owners to independently track this information.

This will require a PR to update the ACP-99 spec as well.

## How this works
Updates `InitiatedValidatorRegistration` to emit the serialized bytes of the `registerL1ValidatorMessage`, and updates `RegisteredInitialValidator` to emit the subnet ID and validator's index in the `ConvertSubnet` message. These events can be queried by the indexed `validationID` to easily recover the information necessary to construct the `justification` for each case.

## How this was tested
CI

## How is this documented
NatSpec


contracts/validator-manager/ACP99Manager.sol
```diff
@@ -74,17 +74,34 @@ struct Validator {
...
+     * @notice The field index is the index of the initial validator in the conversion data.
+     * This is used along with the subnetID as the ACP-118 justification for
+     * signature requests from P-Chain validators over a L1ValidatorRegistrationMessage
+     * when removing the validator
+     */
     event RegisteredInitialValidator(
-        bytes32 indexed validationID, bytes20 indexed nodeID, uint64 weight
+        bytes32 indexed validationID,
+        bytes20 indexed nodeID,
+        bytes32 indexed subnetID,
```
geoff-vball, line 87: Does indexing the subnetID make querying events easier? (not a rhetorical question, I'm actually curious). 

Let's say we have two different ValidatorManager contracts on the C-chain that manage two different subnets. Is it easier to query on an indexed field than it is on the address that emitted the event?

contracts/validator-manager/ACP99Manager.sol
```diff
@@ -74,17 +74,34 @@ struct Validator {
...
+        uint64 weight,
+        uint32 index
```
geoff-vball, line 89: What's the eventual purpose of the index field here?

cam-schultz: It's used along with the `subnetID` to create the preimage of the `validationID` for initial validators. This is the `justification` for P-Chain validators to sign the `L1ValidatorRegistrationMessage` attesting that the validator has been removed. See here: https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/network/warp.go#L153

contracts/validator-manager/ACP99Manager.sol
cam-schultz: I'm not sure. I added this here because there was an unused `indexed` slot, and because searching by `subnetID` seems like a common workflow.

geoff-vball (approved):

bernard-avalabs (approved):

cam-schultz:
Closing in favor of https://github.com/ava-labs/icm-contracts/pull/789

PR #743, opened 2025-03-18
Replay fix for Validator Manager

geoff-vball:
## Why this should be merged
Fixes an issue where:

1. Validator added
2. Validator active
3. Validator removed before `expiry` of original message. 
4. Validator is re-added with original `initiateValidatorRegistration` message.
5. Validator is set to active with original `completeValidatorRegistration` message signed by p-chain

This could cause a validator to be marked as active in the contract, but inactive on the p-chain. Such a validator would not be able to accrue any `uptime`, so in most cases would be ineligible for rewards. We want to avoid potential de-syncs from with the p-chain.

Also aligns `MAXIMUM_REGISTRATION_EXPIRY_LENGTH` to be equal to the [value on the p-chain](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/txs/executor/standard_tx_executor.go#L41).

## How this works
Since we keep validator state indefinitely, we just check if the status is `Unknown` before initiating a registration.

## How is this documented


contracts/validator-manager/tests/ValidatorManagerTests.t.sol
```diff
@@ -482,8 +530,9 @@ abstract contract ValidatorManagerTest is Test {
         uint64 weight,
         uint64 registrationExpiry,
         bytes memory blsPublicKey
-    ) internal returns (bytes32 validationID) {
-        (validationID,) = ValidatorMessages.packRegisterL1ValidatorMessage(
+    ) internal returns (bytes32) {
+        (bytes32 validationID, bytes memory registerL1ValidatorMessage) = ValidatorMessages
+            .packRegisterL1ValidatorMessage(
```
geoff-vball, line 535: Unrelated fix that simplified this code :)

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -94,7 +94,7 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
         0xe92546d698950ddd38910d2e15ed1d923cd0a7b3dde9e2a6a3f380565559cb00;
 
     uint8 public constant MAXIMUM_CHURN_PERCENTAGE_LIMIT = 20;
-    uint64 public constant MAXIMUM_REGISTRATION_EXPIRY_LENGTH = 2 days;
+    uint64 public constant MAXIMUM_REGISTRATION_EXPIRY_LENGTH = 1 days;
```
bernard-avalabs, line 97: Is this change required for this fix?

bernard-avalabs (commented):
Makes sense to me. Just a comment about the expiry length change.

contracts/validator-manager/ValidatorManager.sol
geoff-vball: Not required, no, but aligns with the [value the p-chain uses](https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/txs/executor/standard_tx_executor.go#L41).

PR #759, opened 2025-04-08
N-07 unused named return

geoff-vball:
## Why this should be merged
Named return variables are a way to declare variables that are meant to be used within a function's body for the purpose of being returned as that function's output. They are an alternative to explicit in-line return statements.

In ERC20TokenStakingManager.sol, the [validationID return variable](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ERC20TokenStakingManager.sol#L105) of the initiateValidatorRegistration function is unused.


cam-schultz (commented):
One small comment.

Otherwise, it looks like some changes from the other PRs made its way in. No big deal, but if possible let's merge the other PRs first to minimize the diff against `main` here.

bernard-avalabs (approved):

cam-schultz (approved):

PR #763, opened 2025-04-11
L-06 _validatePChainOwner Does Not Check for Zero Addresses

geoff-vball:
## Why this should be merged
The [_validatePChainOwner function](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L295-L311) performs several checks to ensure that a PChainOwner struct complies with the [ACP-77 specification](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/77-reinventing-subnets/README.md). It validates that the threshold is consistent with the number of addresses and confirms that the address list is sorted. However, it does not check whether any of the addresses in the list are zero addresses.

Allowing zero addresses in the PChainOwner's list could result in unusable ownership configurations. Specifically, if an address is zero, the corresponding owner may be unable to disable their validator on the P-Chain using a DisableL1ValidatorTx, or may be unable to collect their AVAX balance after disabling the validator. This could lead to locked funds or validators that cannot be properly managed, resulting in user-facing issues and operational inefficiencies.

## How this was tested
Unit test

## How is this documented


contracts/validator-manager/ValidatorManager.sol
```diff
@@ -304,6 +305,10 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
         if (pChainOwner.threshold > pChainOwner.addresses.length) {
             revert InvalidPChainOwnerThreshold(pChainOwner.threshold, pChainOwner.addresses.length);
         }
+        // Zero address is invalid. Because we require addresses to be sorted, we only need to check if the first is 0
+        if (pChainOwner.threshold != 0 && pChainOwner.addresses[0] == address(0)) {
```
cam-schultz, line 309: I'd prefer the first condition be `pChainOwner.addresses.length != 0`, since it's a check to prevent out of bounds access on that array.

cam-schultz (approved):

bernard-avalabs (approved):

PR #765, opened 2025-04-11
L-05 Missing Check for Duplicate Addresses in _validatePChainOwner

geoff-vball:
## Why this should be merged
The [_validatePChainOwner function](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L295-L311) ensures that PChainOwner meets the validation criteria outlined in the [ACP-77 specification](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/77-reinventing-subnets#registerl1validatormessage). It correctly checks that if the threshold is 0, the address list is empty, and that the threshold does not exceed the number of addresses. It also verifies that addresses are sorted in ascending order. However, it does not explicitly check for duplicate addresses, which is required by the ACP-77 specification.

If duplicate addresses are allowed, it could lead to unintended behavior when validating ownership thresholds, potentially enabling improper authorization or quorum calculations. This could pose security risks, as the system might incorrectly count duplicate addresses towards the threshold, leading to unintended validator permissions or incorrect staking assignments.

## How this was tested
Unit test

## How is this documented


iansuvak (approved):
Now we have a single error type for two types of user errors but I think that this is fine.

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -311,8 +311,8 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
         // Addresses must be sorted in ascending order
```
cam-schultz, line 311:
```suggestion
        // Addresses must be unique and sorted in ascending order
```
Since we're combining these two cases into a single `InvalidPChainOwnerAddresses` error, let's document the validity check in full

bernard-avalabs (approved):

iansuvak (approved):

cam-schultz (approved):

PR #771, opened 2025-04-14
H-01 ValidatorManager and P-Chain Can Go Out of Sync

geoff-vball:
## Why this should be merged
In the [initiateValidatorRegistration function](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L313) of the ValidatorManager contract, validationId is created using the SHA-256 hash of the [packed registerL1ValidatorMessage](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorMessages.sol#L258) message to be sent to the P-Chain. Since ICM messages are directly linked to validationId, it is expected that validationId is always different. However, a malicious validator in a particular timeframe can generate the same validationId again by reusing the [ValidationPeriod parameters](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L377-L385) which, in turn, allows for reusing the ICM messages later for their own benefit.

An attacker can successfully register through initiateValidatorRegistration with a certain expiry, resulting in a validator with validationId A. Next, the P-chain confirms the registration and sends L1ValidatorRegistrationMessage (M1) which can be [used in completeValidatorRegistration](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L429-L457) to complete the registration on the validation manager contract side. Then, the malicious validator calls initiateValidatorRemoval, once again the P-chain confirms this and sends aL1ValidatorRegistrationMessage (M2), which can be [used in completeValidatorRemoval](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L584-L627). Basically, the validator went through the whole lifecycle of registration and removal.

Now, before the expiry elapses, the attacker calls initiateValidatorRegistration again with the same parameters, resulting in the same validationId A. The P-chain will not confirm this message, as a certain validationId can only be used once from the P-chain's point of view. However, the attacker can reuse the L1ValidatorRegistrationMessage (M1) of the previous validator lifecycle to successfully complete the registration of the validator. Similarly, the attacker can reuse L1ValidatorRegistrationMessage (M2) to complete its removal in line with their own benefits, without requiring confirmation from the P-chain.

This results in the ValidatorManager contract having a different view of the validator set compared to that of the P-chain. Additionally, it introduces complexities such as resetting sentNonce and recreating delegationId using sentNonce.

It is worth noting that this attack is only feasible if the [configured minimum stake duration](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/StakingManager.sol#L47-L48) is less than [the maximum registration expiry window](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/ValidatorManager.sol#L97-L98) as the validator needs to go through the whole validation lifecycle before the registration expiry window is reached.

## How this works
Always uses the max expiry time of 1 day to disable the ability of duplicate registration messages being created by the Validator Manager contract.

## How this was tested
Unit and E2E tests

## How is this documented


contracts/validator-manager/tests/ValidatorManagerTests.t.sol
```diff
@@ -238,19 +229,15 @@ abstract contract ValidatorManagerTest is Test {
...
+            DEFAULT_NODE_ID, DEFAULT_SUBNET_ID, DEFAULT_WEIGHT, DEFAULT_BLS_PUBLIC_KEY
         );
         (, bytes memory registerL1ValidatorMessage) = ValidatorMessages
             .packRegisterL1ValidatorMessage(
             ValidatorMessages.ValidationPeriod({
                 subnetID: DEFAULT_SUBNET_ID,
                 nodeID: DEFAULT_NODE_ID,
                 blsPublicKey: DEFAULT_BLS_PUBLIC_KEY,
-                registrationExpiry: DEFAULT_EXPIRY,
+                registrationExpiry: uint64(block.timestamp) + 1 days,
```
iansuvak, line 240: It looks like `DEFAULT_EXPIRY` is now gone and can be deleted from the constants above but can also make 1 days a new `EXPIRY_LENGTH` constant which would make line 421 more self-explanatory

iansuvak (approved):

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -391,6 +383,12 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
                 weight: weight
             })
         );
+
+        // Ensure that this validation ID has not been used before to prevent replays
+        if ($._validationPeriods[validationID].status != ValidatorStatus.Unknown) {
+            revert InvalidValidatorStatus($._validationPeriods[validationID].status);
+        }
```
cam-schultz, line 390: Is it even possible to hit this case, outside of the obvious case of multiple identical calls to `_initiateValidatorRegistration` in the same block? I'm in favor of keeping this check, but recommend updating the comment to note that it's strictly defensive, and that there's no replay vuln that's exploitable.

contracts/validator-manager/tests/StakingManagerTests.t.sol
```diff
@@ -1391,6 +1386,52 @@ abstract contract StakingManagerTest is ValidatorManagerTest {
...
+        _beforeSend(_weightToValue(DEFAULT_WEIGHT), address(this));
+        bytes memory l1ValidatorRegistrationMessage =
+            ValidatorMessages.packL1ValidatorRegistrationMessage(validationID, true);
+
+        _mockGetPChainWarpMessage(l1ValidatorRegistrationMessage, true);
+
+        vm.expectRevert(
+            abi.encodeWithSelector(ValidatorManager.InvalidValidationID.selector, validationID)
+        );
+        _completeValidatorRegistration(0);
```
cam-schultz, line 1432: This test would hit the same check in `completeValidatorRegistration` if we called that straight away without registering any validators prior.

If we're trying to test the check added in `_initiateValidatorRegistration`, I think we'd want to call `_initiateValidatorRegistration` twice with the same arguments in the same block, such that they result in the same expiry.

cam-schultz: Or, if we're trying to test that the same `l1ValidatorRegistrationMessage` can't be consumed more than once, we could register a validator, initiate removal, and call `_completeValidatorRegistration` twice in succession.

contracts/validator-manager/tests/ValidatorManagerTests.t.sol
```diff
@@ -303,13 +290,45 @@ abstract contract ValidatorManagerTest is Test {
...
+
+        _completeValidatorRemoval(0);
+
+        _beforeSend(_weightToValue(DEFAULT_WEIGHT), address(this));
+        l1ValidatorRegistrationMessage =
+            ValidatorMessages.packL1ValidatorRegistrationMessage(validationID, true);
+
+        _mockGetPChainWarpMessage(l1ValidatorRegistrationMessage, true);
+
+        vm.expectRevert(
```
cam-schultz, line 323: Same comment as above re: target of this test

contracts/validator-manager/ValidatorManager.sol
geoff-vball: So *technically* there's an edgecase of going through the whole register -> complete removal -> register again with same message within the same timestamp, which is practically impossible, but I think having the check in case here is helpful. I can update the comment

contracts/validator-manager/tests/StakingManagerTests.t.sol
geoff-vball: I see, yeah you're right, we need to make a second call to `_initiateValidatorRegistration` at the same timestamp as the first

contracts/validator-manager/tests/StakingManagerTests.t.sol
```diff
@@ -1420,16 +1423,18 @@ abstract contract StakingManagerTest is ValidatorManagerTest {
...
-            ValidatorMessages.packL1ValidatorRegistrationMessage(validationID, true);
 
-        _mockGetPChainWarpMessage(l1ValidatorRegistrationMessage, true);
+        vm.expectRevert(abi.encodeWithSelector(ValidatorManager.InvalidValidatorStatus.selector, 4));
 
-        vm.expectRevert(
-            abi.encodeWithSelector(ValidatorManager.InvalidValidationID.selector, validationID)
-        );
-        _completeValidatorRegistration(0);
+        _initiateValidatorRegistration({
```
cam-schultz, line 1431: Can we add a comment explaining that setting the block timestamp forces this call to resolve to the same `validationID`?

contracts/validator-manager/tests/ValidatorManagerTests.t.sol
```diff
@@ -314,16 +315,18 @@ abstract contract ValidatorManagerTest is Test {
...
-            ValidatorMessages.packL1ValidatorRegistrationMessage(validationID, true);
 
-        _mockGetPChainWarpMessage(l1ValidatorRegistrationMessage, true);
+        vm.expectRevert(abi.encodeWithSelector(ValidatorManager.InvalidValidatorStatus.selector, 4));
 
-        vm.expectRevert(
-            abi.encodeWithSelector(ValidatorManager.InvalidValidationID.selector, validationID)
-        );
-        _completeValidatorRegistration(0);
+        _initiateValidatorRegistration({
```
cam-schultz, line 323: Same comments as above here

cam-schultz (commented):
Some final nits, otherwise LGTM

krud491 (commented):
files

PR #772, opened 2025-04-14
L-04 Use RewardRecipient in claimDelegationFees

iansuvak:
## Why this should be merged

The `claimDelegationFees` function does not take into account the reward recipient that
may have been configured through the `changeValidatorRewardRecipient` function.
Instead, it directly transfers the rewards to the owner of the validator.

This behavior may lead to confusion for validators who expect rewards to be sent to the
designated rewards recipient, as they will instead be transferred to the owner. It could also
introduce unexpected integration challenges for protocols that assume that the rewards
recipient will receive all associated fees.

## How this works
It checks the `$._rewardRecipients[validationID];` before rewarding delegation fees just like `completeValidatorRemoval` does. It also removes the deletion of the reward recipient from the mapping since we don't have a way of knowing that all delegations have been completed yet. 

Alternatives here would be either to implement a way to know this from the `validationID` or to just remove them upon `claimDelegationFees` and `completeValidatorRemoval` but allow the user to reset it again with another call to `changeValidatorRewardRecipient` and then document that option. 

## How this was tested
`testClaimDelegationFeesWithRewardRecipient()`  added

## How is this documented


cam-schultz (approved):
LGTM

PR #777, opened 2025-04-16
N-03 Missing Getter Functions

cam-schultz:
## Why this should be merged
The StakingManager contract contains state variables and mappings, leveraging the [EIP-7201 namespaced storage layout](https://eips.ethereum.org/EIPS/eip-7201). While this data is accessible by off-chain components, accessing this data without getter functions is cumbersome if the contract relies on EIP-7201. Especially, the state variables and mappings related to tracking the [PoS validator information](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/StakingManager.sol#L64) and [delegator information](https://github.com/ava-labs/icm-contracts/blob/8a8f78dcda0b0af60c6fb1762f7927351357d645/contracts/validator-manager/interfaces/IStakingManager.sol#L50-L58) might benefit from getter functions as these will likely be used by the frontend to show the status of certain validators and delegators.

## How this works

Adds getters for all `StakingManager` and `ValidatorManager` state variables that don't purely represent internal contract state.

## How this was tested
CI

## How is this documented
N/A


contracts/validator-manager/ValidatorManager.sol
```diff
@@ -494,6 +494,32 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
...
+        return $._registeredValidators[nodeID];
+    }
+
+    /**
+     * @notice Returns true if the ValidatorManager has been initialized with the initial validator set
+     */
+    function getInitializedValidatorSet() public view returns (bool) {
+        ValidatorManagerStorage storage $ = _getValidatorManagerStorage();
+        return $._initializedValidatorSet;
+    }
```
geoff-vball, line 521: Should we rename this boolean to something like `isValidatorSetInitialized`? This function name makes me think you're returning a set of validators

cam-schultz: Renamed this function, but left the internal variable name alone

contracts/validator-manager/ValidatorManager.sol
```diff
@@ -494,6 +494,32 @@ contract ValidatorManager is Initializable, OwnableUpgradeable, ACP99Manager {
         return _getValidatorManagerStorage()._subnetID;
     }
 
+    /**
+     * @notice Returns the current churn tracker and its configuration
+     */
```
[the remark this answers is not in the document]
cam-schultz: Added return natspec for all new getters that return multiple values

contracts/validator-manager/StakingManager.sol
```diff
@@ -584,6 +584,69 @@ abstract contract StakingManager is
...
+            maximumStakeMultiplier: uint8($._maximumStakeMultiplier),
+            weightToValueFactor: $._weightToValueFactor,
+            rewardCalculator: $._rewardCalculator,
+            uptimeBlockchainID: $._uptimeBlockchainID
+        });
+    }
+
+    /**
+     * @notice Returns the PoS validator information for the given validationID
+     * @notice See {ValidatorManager-getValidator}
```
geoff-vball, line 607:
```suggestion
```

cam-schultz: I included this to emphasize that this getter returns a subset of all the validator information for a given `validationID`. I think updating to something like this would clarify things:

```suggestion
     * @notice See {ValidatorManager-getValidator} to retreive information about the validator not specific to PoS
```

Happy to remove it altogether if you think it doesn't meaningfully clarify things

geoff-vball: Ahhh yes that makes more sense. We also have two `@notice`s here, which I'm not sure is the best way to format the doc?

cam-schultz: Removed

iansuvak (approved):

geoff-vball (approved):

PR #784, opened 2025-05-06
M-01 Limit churn period to 24 hours

cam-schultz:
## Why this should be merged

When registering PoS validators, the 24 hour registration expiry acts as a bonding period to prevent the staked assets from being used to artificially inflate the churn amount. If the churn period is greater than 24 hours, a fixed amount of staking asset can be reflected multiple times in the churn period by waiting for the registration expiry to pass, withdrawing the stake, and using it to register another validation period.

## How this works

Imposes a maximum value for the churn period, set to the registration expiry of 24 hours.

## How this was tested

CI

## How is this documented

Comments in contract


geoff-vball (approved):
Changes LGTM. Looks like a test failed for networking reasons, and the linter will be fixed when https://github.com/ava-labs/icm-contracts/pull/782 is merged

iansuvak (approved):

geoff-vball (approved):

PR #791, opened 2025-05-16
Document delegator stake removal edge case

cam-schultz:
## Why this should be merged

Documents an edge case in which a delegator's stake may be unrecoverable due to repeated issuance of `DisableL1ValidatorTx`s on the P-Chain.

In practice, honest validators should never engage in this behavior, as becoming inactive would impact their claimable validation rewards, and there is no upside to preventing another entity from reactivating them by paying to increase their balance.

## How is this documented


iansuvak (approved):

cam-schultz:
Closing in favor of https://github.com/ava-labs/icm-contracts/pull/793

PR #800, opened 2025-06-13
Bump avago subnet

iansuvak:
## Why this should be merged
Updates dependencies in testing to the current versions.
## How this works

- Updates avago + subnet deps
- Updates existing manually created files to reference `libevm` instead of `go-ethereum`
- Updates the E2E testing code for the new tmpnet interfaces. 

## How this was tested
CI
## How is this documented


scripts/e2e_test.sh
```diff
@@ -73,6 +73,8 @@ cp ${BASEDIR}/subnet-evm/subnet-evm ${BASEDIR}/avalanchego/plugins/srEXiWaHuhNyG
 echo "Copied ${BASEDIR}/subnet-evm/subnet-evm binary to ${BASEDIR}/avalanchego/plugins/"
 
 export AVALANCHEGO_BUILD_PATH=$BASEDIR/avalanchego
+export AVALANCHEGO_PATH=$AVALANCHEGO_BUILD_PATH/avalanchego
```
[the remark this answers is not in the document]
iansuvak: The actual location of the build is the same as it was before. `~/.teleporter-deps/avalanchego/avalanchego` where the final avalanchego is the actual binary but the folder also contains a `plugins` subfolder with the `subnet-evm` binary there.

The only difference here is that instead of passing that path in the code to tmpnet we now set it as a env variable that tmpnet reads.

geoff-vball (approved):

tests/network/network.go
```diff
@@ -129,14 +130,11 @@ func NewLocalNetwork(
 	network := subnetEvmTestUtils.NewTmpnetNetwork(
 		name,
 		bootstrapNodes,
-		utils.WarpEnabledChainConfig,
+		tmpnet.FlagsMap{},
```
cam-schultz, line 133: Do tmpnet chains have the Warp enabled config set by default? This _should_ be the case, but just confirming that removing this is a no-op

tests/utils/chain.go
```diff
@@ -666,3 +666,55 @@ func SetupProposerVM(ctx context.Context, fundedKey *ecdsa.PrivateKey, network *
 	err = subnetEvmUtils.IssueTxsToActivateProposerVMFork(ctx, chainIDInt, fundedKey, client)
 	Expect(err).Should(BeNil())
 }
+
+// Adapted from [subnetEvmUtils.IssueTxsToActivateProposerVMFork]
+// Since those transactions with hardcoded low caps don't get included successfully
+// post Fortuna
```
cam-schultz, line 672: Did this break anything in `subnet-evm`? We should open a PR to fix it there too if so.

cam-schultz (approved):
LGTM, just a couple of quick questions

tests/utils/chain.go
cam-schultz: Makes sense, thanks.

PR #806, opened 2025-07-02
Slot Auction Validator Manager Contract and E2E tests

EmreDincoglu:
## Why this should be merged
Slot Auction Validator Manager is a project that was assigned to me
## How this works
Slot auction validator manager is a new solidity contract that receives bids from users to purchase a validation slot, also has its own E2E testing file with a lot more added utility functions catered towards the SAVM
## How this was tested
Still being tested with in progress E2E testing files that is soon to be completed
## How is this documented
Will be documented with a readme soon


solc-macos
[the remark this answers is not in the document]
EmreDincoglu: Yes I definitely did, I will end up making an interface since it seems useful to make in general, but for this PR I don't need it here

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -0,0 +1,207 @@
...
+        
+        validatorSlots = 0;
+        validationTimeLimit = 0;
+    }
+
+    function initiateValidatorRemoval(
+        bytes32 validationID
+    ) public {
+        VALIDATOR_MANAGER.initiateValidatorRemoval(validationID);
+    }
```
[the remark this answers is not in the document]
EmreDincoglu: Correct

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -0,0 +1,207 @@
...
+        //no re entry
+        //probably a lot more
+       
+        auctionEndTime = block.timestamp + auctionLength; //sets the minimum end time of the auction, can go over
+        validatorSlots = validatorslots;
+        validationTimeLimit = validationLength;
+        _secondPrice = 0;
+
+        delete _bids.tree; //initiates an empty heap array, id rather just create a new one but solidity memory syntax is still confusing to me
+        auctionInProgress = true;
```
geoff-vball, line 85: emit event here

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -0,0 +1,207 @@
...
+        else {
+            return;
+        }
+        bidderInfo[bid] = ValidatorBid(msg.sender, bid, nodeID, blsPublicKey, remainingBalanceOwner, disableOwner);
+        _nodeIsQualified[nodeID] = true;
+    }
+
+    function endAuction() external {
+        //only owner
+        require(auctionInProgress, "Auction not in progress");
```
geoff-vball, line 126: Let's make this check into a modifier

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -0,0 +1,207 @@
...
+        auctionEndTime = block.timestamp + auctionLength; //sets the minimum end time of the auction, can go over
+        validatorSlots = validatorslots;
+        validationTimeLimit = validationLength;
+        _secondPrice = 0;
+
+        delete _bids.tree; //initiates an empty heap array, id rather just create a new one but solidity memory syntax is still confusing to me
+        auctionInProgress = true;
+    }
+
+    function placeBid(
```
geoff-vball, line 88: Emit event on successful bid

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -0,0 +1,207 @@
...
+        while (Heap.length(_bids) > 0) {
+            //grabs next bid
+            uint256 currentBid = Heap.pop(_bids);
+            ValidatorBid memory bidInfo = bidderInfo[currentBid];
+            //sends back extra tokens due to second price
+            uint256 tokenRefund = currentBid - _secondPrice;
+            TOKEN_CONTRACT.transfer(bidInfo.addr, tokenRefund);
+            //registers as validator
+            bytes32 validationID = _initiateValidatorRegistration(bidInfo.nodeID, bidInfo.blsPublicKey, bidInfo.remainingBalanceOwner, bidInfo.disableOwner, validatorWeight);
+            currentValidators[bidInfo.nodeID] = ValidatorInfo(bidInfo.addr, validationTimeLimit + block.timestamp, bidInfo.nodeID, bidInfo.blsPublicKey, validationID, validatorWeight);
```
geoff-vball, line 144: We should emit an event with all of this info in it, aside from pubkey and addr, I think. (probably just look at other validator manager code and see what is emitted, plus add the expiry time).

EmreDincoglu: Emitting an event with an address. Just followed what IStakingManager does

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -0,0 +1,207 @@
...
+        bytes memory nodeID,
+        bytes memory blsPublicKey,
+        PChainOwner memory remainingBalanceOwner,
+        PChainOwner memory disableOwner,
+        uint64 weight
+    ) internal returns (bytes32) {
+        bytes32 validationID = VALIDATOR_MANAGER.initiateValidatorRegistration(
+            nodeID, blsPublicKey, remainingBalanceOwner, disableOwner, weight
+        );
+        return validationID;
```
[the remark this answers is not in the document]
EmreDincoglu: Overall removed the function since the only thing that will call it is the endAuction function which will have non re rentrant and also be only owner so shouldn't need a separate internal function

contracts/validator-manager/interfaces/ISlotAuctionManager.sol
```diff
@@ -0,0 +1,34 @@
...
+
+import {IValidatorManager} from "../interfaces/IValidatorManager.sol";
+import {PChainOwner} from "contracts/validator-manager/interfaces/IACP99Manager.sol"; //not sure why it doesnt like relative path
+
+
+struct ValidatorBid {
+    address addr;
+    uint256 bid;
+    bytes nodeID;
+    bytes blsPublicKey;
```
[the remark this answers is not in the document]
EmreDincoglu: Validator manager takes in values of bytes for both BLS pub key and NodeID, if conversion is feasible then will do otherwise not possible

PR #807, opened 2025-07-09
Update go

geoff-vball:
## How is this documented


geoff-vball:
@iansuvak definitely didn't mean to update `subnet-evm`

PR #820, opened 2025-07-30
Abstract Contract, Unit tests, Upgradability

EmreDincoglu:
## Why this should be merged
Adds a lot of functionality 
## How this was tested
All unit test and e2e tests pass
## How is this documented
In code documentation


contracts/validator-manager/ERC20TokenSlotAuctionManager.sol
```diff
@@ -10,54 +10,92 @@ import {IERC20Mintable} from "./interfaces/IERC20Mintable.sol";
...
-import {AuctionState} from "./interfaces/ISlotAuctionManager.sol";
-
+import {SlotAuctionManagerSettings} from "./interfaces/ISlotAuctionManager.sol";
+import {ICMInitializable} from "@utilities/ICMInitializable.sol";
 
-/**
- * @dev Implementation of the {IERC20TokenStakingManager} interface.
- *
- * @custom:security-contact https://github.com/ava-labs/icm-contracts/blob/main/SECURITY.md
- */
```
geoff-vball, line 25: Did you mean to delete these?

contracts/validator-manager/ERC20TokenSlotAuctionManager.sol
```diff
@@ -10,54 +10,92 @@ import {IERC20Mintable} from "./interfaces/IERC20Mintable.sol";
...
+        IERC20Mintable _token;
+    }
 
-    // struct ERC20TokenStakingManagerStorage {
-    // }
-    
-    // solhint-enable private-vars-leading-underscore
+    // keccak256(abi.encode(uint256(keccak256("avalanche-icm.storage.ERC20TokenSlotAuctionManager")) - 1)) & ~bytes32(uint256(0xff));
+    bytes32 public constant ERC20_TOKEN_SLOT_AUCTION_MANAGER_STORAGE_LOCATION =
+        0x04264d6e045c48d92b64fb3ce155b1f7a2673239fb0c9b60c505be1c17a7e700;
```
geoff-vball, line 29: Putting a note here to confirm this value

contracts/validator-manager/ERC20TokenSlotAuctionManager.sol
```diff
@@ -10,54 +10,92 @@ import {IERC20Mintable} from "./interfaces/IERC20Mintable.sol";
...
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function __ERC20TokenSlotAuctionManager_init(
+        SlotAuctionManagerSettings calldata settings,
+        IERC20Mintable token
+    ) internal onlyInitializing {
+        __SlotAuctionManager_init(settings);
+        __ERC20TokenSlotAuctionManager_init_unchained(token);
+    }
```
geoff-vball, line 70: You should have consistent ordering of the params for these functions

contracts/validator-manager/NativeTokenSlotAuctionManager.sol
```diff
@@ -8,32 +8,43 @@ pragma solidity 0.8.25;
...
-import {IValidatorManager} from "./interfaces/IValidatorManager.sol";
-
+import {ICMInitializable} from "@utilities/ICMInitializable.sol";
+import {SlotAuctionManagerSettings} from "./interfaces/ISlotAuctionManager.sol";
 
-/**
- * @dev Implementation of the {INativeTokenStakingManager} interface.
- *
- * @custom:security-contact https://github.com/ava-labs/icm-contracts/blob/main/SECURITY.md
- */
```
geoff-vball, line 25: Same thing as the other contract

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
+    error ZeroMinBid();
 
-    modifier AuctionOn {
-        if (auctionState != AuctionState.AuctionInProgress) {
+    // solhint-disable ordering
+
+    modifier AuctionOn() {
+        SlotAuctionManagerStorage storage $ = _getSlotAuctionManagerStorage();
+
+        if ($._auctionState != AuctionState.AuctionInProgress) {
```
geoff-vball, line 101: This modifier just checks that the auction is in progress (which is correct), but the other modifier checks for each of the other states and emits a different error for each. I don't have a strong preference on which way you do it, but it should be consistent.

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
+            $._minValidatorDuration,
+            $._auctionEndTime,
+            $._minimumBid
+        );
+    }
+
+    function endAuction() external nonReentrant AuctionOn {
+        SlotAuctionManagerStorage storage $ = _getSlotAuctionManagerStorage();
+
+        $._auctionCooldownEndtime = block.timestamp + 1 days;
```
geoff-vball, line 211: I think the cooldown time should be configurable - unless we've done this specifically to align with the registration expiries, in which case we should make an appropriately named `1 days` constant and should comment on why we're doing this

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
-        auctionState = AuctionState.AuctionFinalizing;
-        if (block.timestamp <= auctionEndTime) {
-            revert AuctionEndTimeNotReached(auctionEndTime);
+        if (address(manager) == address(0)) {
+            revert ZeroAddress();
+        }
+        // Minimum stake duration should be at least one churn period in order to prevent churn tracker abuse.
+        if (auctionSettings.minValidatorDuration < manager.getChurnPeriodSeconds()) {
+            revert InvalidMinValidatorDuration(auctionSettings.minValidatorDuration);
+        }
```
geoff-vball, line 150: I don't think we need this check with the slot auction manager... This constant was introduced into the staking managers so that someone could not stake and unstake rapidly to fill up the churn tracker space. The minimum duration is just how long they have to wait to get their stake back after initially registering, basically

geoff-vball (commented):
Still going, but here are a good chunk of things to work on :)

contracts/validator-manager/SlotAuctionManager.sol
EmreDincoglu: I originally did it because of the registration expiry like you said, due to the potential of slots being won but not used, so there would be a cooldown to wait and see if they end up registering allowing for those slots to be open again, but I think it may not be necessary since the plan is to run this validator manager alongside another meaning there is nothing to gain by paying for a validator slot and not validating so I will make it a configurable value

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
+        if (block.timestamp < $._auctionEndTime) {
+            revert AuctionEndTimeNotReached($._auctionEndTime);
         }
 
-        while (Heap.length(_bids) > 0) {
-            uint256 currentBid = Heap.pop(_bids);
-            ValidatorBid memory bidInfo = bidderInfo[currentBid];
+        // avoids array out of bounds for Heap.peek
+        if ($._secondPrice == 0 && Heap.length($._bids) != 0) {
+            $._secondPrice = Heap.peek($._bids);
```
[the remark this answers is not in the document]
EmreDincoglu: I check in endAuction that if the second price is zero then it makes the lowest winning bidder pay the price they bid. But for clarity I do agree that it would be much simpler to make second price the minimum bid at first so I will change it

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
+            _lock(bid);
+            Heap.insert($._bids, bid);
+        } else if (Heap.peek($._bids) < bid) {
             _lock(bid);
-            // TOKEN_CONTRACT.transferFrom(msg.sender, address(this), bid);
-            uint256 poppedBid = Heap.replace(_bids, bid);
-            _secondPrice = poppedBid;
-            emit BidEvicted(poppedBid, bidderInfo[poppedBid].nodeID);
+
+            $._secondPrice = Heap.replace($._bids, bid);
```
geoff-vball, line 385: imo we should store this in a variable called `evictedBid` and then assign it to `$._secondPrice`. Right now calling `_unlock` on `_secondPrice` is weird and slightly confusing

contracts/validator-manager/interfaces/ISlotAuctionManager.sol
```diff
@@ -44,12 +43,33 @@ struct ValidatorInfo {
...
+    address admin;
+    IValidatorManager manager;
+    AuctionSettings auctionSettings;
 }
+/**
+ * @dev Struct for auction settings, is a part of SlotAuctionManagerSettings but due
+ * to the nature of being able to change auction settings a separate struct is provided
+ */
 
+struct AuctionSettings {
```
geoff-vball, line 62: There's an empty line above this that doesn't need to be there

contracts/validator-manager/interfaces/ISlotAuctionManager.sol
```diff
@@ -44,12 +43,33 @@ struct ValidatorInfo {
...
+    uint16 totalValidatorSlots;
+    uint64 weight;
+    uint256 minValidatorDuration;
+    uint256 minAuctionDuration;
+    uint256 minimumBid;
+}
 /**
  * Proof of Purchase Validator Manager that takes ERC20 tokens.
  */
+
```
geoff-vball, line 72:
```suggestion
```

contracts/validator-manager/interfaces/ISlotAuctionManager.sol
```diff
@@ -44,12 +43,33 @@ struct ValidatorInfo {
...
+ * to the nature of being able to change auction settings a separate struct is provided
+ */
 
+struct AuctionSettings {
+    uint16 totalValidatorSlots;
+    uint64 weight;
+    uint256 minValidatorDuration;
+    uint256 minAuctionDuration;
+    uint256 minimumBid;
+}
```
geoff-vball, line 68:
```suggestion
}

```

geoff-vball (commented):
Left the rest of my comments for the contracts. Will get to the testing when I have a chance

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
-                bidInfo.nodeID, bidInfo.blsPublicKey, bidInfo.remainingBalanceOwner, bidInfo.disableOwner, validatorWeight
+            if (currentBid - $._secondPrice != 0) {
+                _unlock(bidInfo.addr, currentBid - $._secondPrice);
+            }
+            bytes32 validationID = $._manager.initiateValidatorRegistration(
+                bidInfo.nodeID,
+                bidInfo.blsPublicKey,
+                bidInfo.remainingBalanceOwner,
+                bidInfo.disableOwner,
+                $._validatorWeight
```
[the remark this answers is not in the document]
EmreDincoglu: how about `_auctioningValidatorWeight`?

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,440 @@
...
-            revert ValidatorTimeLimitNotPassed(validatorsByValidationID[validationID].endTime);
+        if ($._validatorsByValidationID[validationID].endTime > block.timestamp) {
+            revert ValidatorTimeLimitNotPassed($._validatorsByValidationID[validationID].endTime);
         }
-        delete validatorsByNodeID[validatorsByValidationID[validationID].nodeID];
-        delete validatorsByValidationID[validationID];
-        VALIDATOR_MANAGER.initiateValidatorRemoval(validationID);
+        delete $._validatorsByNodeID[$._validatorsByValidationID[validationID].nodeID];
+        delete $._validatorsByValidationID[validationID];
+        $._openValidatorSlots++;
```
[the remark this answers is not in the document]
EmreDincoglu: I looked and your right if they dont send the message and the time limit expires then they do call `completeValidatorRemoval` instead so ill delete and change the `_occupiedValidatorSlots` there

anviimishra:
again, a general ques to get more context but why do you cap the number of slots to a specific number?

EmreDincoglu:
> again, a general ques to get more context but why do you cap the number of slots to a specific number?

If there wasnt a maximum amount of validator slots then there would be no competition for bidding, meaning everyone would bid the lowest amount possible and then everyone would be a validator and it just wouldnt make any sense in the sense of an auction

tests/utils/validator_manager.go
```diff
@@ -698,75 +794,75 @@ func InitiateAndCompleteNativeValidatorRegistration(
...
-	registrationEvent, err := GetEventFromLogs(
-		receipt.Logs,
-		acp99Manager.ParseCompletedValidatorRegistration,
-	)
-	Expect(err).Should(BeNil())
-	Expect(registrationEvent.ValidationID[:]).Should(Equal(validationID[:]))
-
-	return registrationInitiatedEvent
-}
+// func InitiateAndCompletePoAValidatorRegistration(
```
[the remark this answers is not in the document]
EmreDincoglu: Intern brain made it thinking I needed a function similar to it but it was never used, removed now

contracts/validator-manager/SlotAuctionManager.sol
```diff
@@ -5,242 +5,429 @@
...
+        $._manager.initiateValidatorRemoval(validationID);
     }
 
     function completeRemoveInitialValidator(
         uint32 messageIndex
     ) external {
-        VALIDATOR_MANAGER.completeValidatorRemoval(messageIndex);
+        SlotAuctionManagerStorage storage $ = _getSlotAuctionManagerStorage();
+
+        $._manager.completeValidatorRemoval(messageIndex);
```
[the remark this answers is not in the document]
EmreDincoglu: Now that I have clarity on what i'm doing I realized that I don't need a removeInitialValidator function. What I should do instead is check to see if I have the validationID stored and if not just allow the validator to be removed since that means it was not initiated by the SAVM, this does cause problems in the case where staking manager is run along side SAVM since it would allow users to remove stakingManager validators using the SAVM, but for the sake of all this I don't think I have time or the power to edit both SAVM and StakingManager to allow to coexist

PR #828, opened 2025-08-13
Update go

geoff-vball:
## How is this documented


EmreDincoglu (approved):
Id recommend C++ instead but looks good overall

iansuvak (approved):

PR #852, opened 2025-10-17
Remove subnet-evm submodule

geoff-vball:
## Why this should be merged
Removed the dependency on the subnet-evm submodule. 

## How this works
Copies over `IWarpMessenger`, `INativeMinter`, and `IAllowList`. The contracts were changed from solidity `0.8.24` to `0.8.25` to match our contracts, but this should have no practical effect because we're only using them for interfaces.

## How this was tested
Existing tests

## How is this documented


iansuvak (approved):
🎉

PR #854, opened 2025-10-20
Update to tagged rcs

iansuvak:
## How is this documented


tests/flows/teleporter/relayer_modifies_message.go
```diff
@@ -122,7 +124,9 @@ func createAlteredReceiveCrossChainMessageTransaction(
 	numSigners, err := signedMessage.Signature.NumSigners()
 	Expect(err).Should(BeNil())
 
+	upgradeRules := upgrade.GetConfig(constants.LocalID)
```
geoff-vball: Is there a reason this is different than the value in `teleporter.go`?

michaelkaplan13 (approved):

PR #855, opened 2025-10-23
Epoch duration

geoff-vball:
## Why this should be merged
- Make e2e tests pass with longer epoch duration time.

## How this works
- When signing message, we get the current epoch and use the p-chain height to call CreateSignedMessage if granite activated. 

## How this was tested
- E2E tests

## How is this documented


iansuvak (approved):
We should merge this as it's a functional change of allowing using configurable epoch durations in the tests and revisit documentation and making tests more time efficient separately.

PR #891, opened 2025-11-25
Replaced explicit warp contract in TeleporterMessenger to use an inte…

batconjurer:
This PR replaces the explicit Warp contract in the TeleporterMessenger contracts with an interface. An implimentation of this interface is provided that works identically for internal interop. It is built on top of https://github.com/ava-labs/icm-contracts/pull/888


avalanche/contracts/teleporter/IWarpExt.sol
```diff
@@ -0,0 +1,27 @@
+// (c) 2025, Ava Labs, Inc. All rights reserved.
```
mdelle1, line 1: Why does this code live in the `avalanche/...` directory? As fas as I understand, contracts implementing `IWarpExt` should live on external chains that don't have access lists?

